PR #17, opened 2019-08-27
Array to string processor

HarisAdzemovic:
This PR is mainly a merge of a new processor for SonarQube bug S2116 (https://rules.sonarsource.com/java/type/Bug/RSPEC-2116) of which there is one occurrence in Spoon-Core.

It also disables sniper mode as it is no longer functioning. Seeing as how it was using some imports marked experimental, this isn't too surprising.

It also fixes one of the failing tests (of other processors) but not all. There'll have to be some additional rewriting for those. Namely the NullDereferenceProcessor which doesn't solve its issues with a try-catch block and the NonSerializableSuperClassProcessorTest which doesn't find any issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: java
 
 jdk:
-  - oraclejdk8
+  - oraclejdk9
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: java
 
 jdk:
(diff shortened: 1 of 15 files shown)


monperrus:
Congrats on the first PR.

> It also disables sniper mode as it is no longer functioning.

OK, we may have to fix this :-)

@nharrand do you review and merge?

PR #20, opened 2019-09-05
Boxed equals proocessor

HarisAdzemovic:
Processor for fixing Sonarqube violation S4973, using == instead of .equals() when comparing boxed types.

### DIFF
--- a/src/main/java/BoxedTypesEqualsProcessor.java
+++ b/src/main/java/BoxedTypesEqualsProcessor.java
@@ -1,0 +1,30 @@
+import spoon.processing.AbstractProcessor;
+import spoon.reflect.code.CtCodeSnippetExpression;
+import spoon.reflect.code.CtExpression;
+import spoon.reflect.code.CtInvocation;
+import spoon.reflect.declaration.CtMethod;
+import spoon.reflect.declaration.CtType;
+import spoon.reflect.reference.CtExecutableReference;
+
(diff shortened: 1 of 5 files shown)


src/main/java/BoxedTypesEqualsProcessor.java
```diff
@@ -0,0 +1,54 @@
...
+            CtBinaryOperator op = (CtBinaryOperator)candidate;
+            if(op.getKind() == BinaryOperatorKind.EQ || op.getKind() == BinaryOperatorKind.NE){
+                CtExpression left = op.getLeftHandOperand();
+                CtExpression right = op.getRightHandOperand();
+                /*
+                The reason we don't check for the case where one variable is boxed is because Java implicitly unboxes
+                the boxed type to primitive, making the == check fine. See JLS #5.6.2:
+                https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.6.2
+                */
+                String nullType = getFactory().Code().createCodeSnippetExpression("").getType().NULL_TYPE_NAME;
```
nharrand:
```suggestion
                String nullType = getFactory().Type().NULL_TYPE.getQualifiedName();
```

HarisAdzemovic: Agreed that's a nicer solution. Commit incoming.

PR #22, opened 2019-09-09
WIP Reimplementing sniper mode

HarisAdzemovic:
Sniper mode has been updated and the old implementation in sonarqube-repair
was non-functional. This is the first step to fixing it.

It works to an extent; 2/4 tested files work fine but 2 fail with error message [num substituted for actual numbers]:
spoon.SpoonException: Cannot compare this: [num, num] with other: ["num", "num"].
The two failing files are included in the commit, AbstractSourceFragmentContext.java & CodeFactory.java.
Both are taken directly from the spoon-core repository.

For Codefactory, the offending line is line 200 (CtConstructor<T> constructor =...)
while for AbstractSrc, 4 methods are breaking. They are:
    findIndexOfNextChildTokenByValue
    findIndexOfNextChildTokenByType
    findIndexOfNextChildTokenOfRole
    findIndexOfNextChildTokenOfElement


HarisAdzemovic:
@pvojtechovsky, I know you've previously helped with getting sniper mode to work for this project and was wondering if you could spare some time?

In this PR, there are two files included which cannot be rapaired with Sniper mode (they throw errors). "Normal" repair works fine.

I’ve been able to narrow down the offending code in the files:

For CodeFactory, it’s only the single line:
CtConstructor<T> constructor = ((CtClass) superClass).getConstructor(Arrays.stream(parameters).map(x -> x.getType()).toArray(CtTypeReference[]::new));

While for AbstractSourceFragmentContext, it’s 4 method invocations that fail (all 4 fail individually/all must be disabled for the analysis to work):
findIndexOfNextChildTokenByValue
findIndexOfNextChildTokenByType
findIndexOfNextChildTokenOfRole
findIndexOfNextChildTokenOfElement

I also tried using your PR https://github.com/INRIA/spoon/pull/2683 but got the same issues.

Many thanks,
Haris

pvojtechovsky:
@HarisAdzemovic I will have a look. But I don't know yet when.

HarisAdzemovic:
> @HarisAdzemovic I will have a look. But I don't know yet when.

Much appreciated! Thank you!

pvojtechovsky:
@HarisAdzemovic I checked out branch sniperModeRepair, find there 5 tests, run them
image
and 3 of 5 are failing with different exceptions than you mention in comments above.
Note: I don't know much details about sonarqube-repair, so I need more exact description of which test exactly fails and what is expected, etc... The best would be if you would be able to prepare as small as possible test case, which fails. Then I can analyze that one test and suggest some fix.
But actually your comments above doesn't fit to what I see in my workspace, so I do not know where exactly you need my support.

HarisAdzemovic:
Hi @pvojtechovsky, thank you for helping me out with this. Upon rereading the PR I realise I wasn’t very clear at all so will try to amend that now.

First of all, it’s worrisome that we don’t get the same result when running all tests. Both me and Travis only find the newest test, CodeFactory, failing. What do you think this disparity is caused by? I use the IntelliJ IDE and Travis runs mvn test -B.

Second, I've been able to narrow down the problem to the use of lambdas. @nharrand helped me translate the failing line of CodeFactory into one which doesn't use arrow notation and this made sniper mode work for the file.

I've added a new test case, CodeFactoryTest, which is quite minimal. Running it as-is should fail. 
To see that non-sniper mode repair works you can comment out the failing line, 17, in CodeFactoryTest.java (TestHelp.repair...). [_repair_ = sniper mode and _normalRepair_ is without sniper mode].
To see that it is the lambda causing the issue, you can reactive sniper-mode in the test file and instead comment out line 19 in CodeFactory.java (CtConstructor<T> constructor = ...) while removing the block comment below it which is the equivalent statement (tested by running all spoon-core tests with it substituted) written without arrow notation. The test should now pass.

pvojtechovsky:
I found two problems:
1) CtElementImpl.toString() is now (after https://github.com/INRIA/spoon/pull/3099 ) very heavy operation, which sometime uses sniper mode printer too. So it is not good idea to call it in your processor code. So I replaced toString() calls by getSignature().
2) that lambda parameter position is wrong computed. I made a PR with failing test here https://github.com/INRIA/spoon/pull/3135

monperrus:
Thanks a lot Pavel for the analysis.

We can now also use "toStringDebug" which is meant to be safer.

PR #24, opened 2019-10-04
Spoon 2116 normal repair

HarisAdzemovic:
Applying normal (not sniper) repair to CodeFactory. This was reverted earlier because it needed to go into a PR so that whitespaces could be ignored in the Github GUI (only possible for PRs, not fie history).


HarisAdzemovic:
Original branch was not reverted successfully (still contained the true fix) before applying the repair. Will try again.

PR #25, opened 2019-10-04
Diffs of transformations performed on external projects

HarisAdzemovic:
Applying normal (not sniper) repair to CodeFactory. This was reverted earlier because it needed to go into a PR so that whitespaces could be ignored in the Github GUI (only possible for PRs, not fie history).


HarisAdzemovic:
A lot of interim commits until we settled for the current format of only saving the original and the diff file (calculated using diff -U 99999 original.file repaired.file)

HarisAdzemovic:
@nharrand do you agree with the chosen format? If yes, could you accept the PR? I want to make some more PRs to external libs but it looks like many fixes will need (small) manual changes and those need to be documented.

PR #39, opened 2019-11-13
NoSuchElementException PR to Apache PDFBox

HarisAdzemovic:
Link to PR:
https://github.com/apache/pdfbox/pull/76

Link to Jira ticket:
https://issues.apache.org/jira/browse/PDFBOX-4687

Had to use normal-mode for one of the files due to it throwing some exception in sniper mode.
The file transformed with sniper mode had to have a newline added before the new statement.


fermadeiral:
Just for the record, the link to the PR is wrong. The right one is https://github.com/apache/pdfbox/pull/75

PR #42, opened 2019-12-02
WIP:Failing old processor and test case

HarisAdzemovic:
This is an old implementation of a processor for Sonarqube rule S2055 with a failing test case.
It is recommended that an attempt at fixing it changes the filter to use Spoon rather than the Sonarjava plugin.
For examples of this, look at other processors of type BUG (except S2095).

### DIFF
--- a/src/main/java/NonSerializableSuperClassProcessor.java
+++ b/src/main/java/NonSerializableSuperClassProcessor.java
@@ -2,8 +2,6 @@
  * DISABLED as per commit b965cad6f327d8fd0fb97a3af8f6427de61685c4.
  * The test is unable to find the error in the initial state. See commit message above for more info.
  */
-/*
-/*
 import org.json.JSONArray;
 import org.json.JSONException;
 import spoon.processing.AbstractProcessor;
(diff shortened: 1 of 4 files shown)


monperrus:
Haris, what should we do with this WIP?

HarisAdzemovic:
I reckon one of two things should be done:
1. Have it remain open but change the name to something akin to "Implement S2055" and add a label indicating that it is up for the taking (that it isn't "claimed" by me)
2. Replace this PR with one that deletes the old implementation of S2055 completely. If someone wants to implement it at a later time they can then open a new PR.

What do you think is best?

fermadeiral:
Hi Haris,

> Replace this PR with one that deletes the old implementation of S2055 completely.

But this was already done, by you, in https://github.com/kth-tcs/sonarqube-repair/pull/41

I suggest that we just close this PR. I'm preparing a list of Sonar rules that we might be able to cover, and I will put the rule 2055 on it, to be implemented by any of us.

Fernanda

HarisAdzemovic:
Good catch Fernanda! I agree with closing the PR.

PR #44, opened 2020-03-10
Fix removal of compliance comments for testing purposes.

fermadeiral:
For testing, we need to insert the comment `// Noncompliant` in lines where there are rule violations (this is mandatory according to Sonar).

We have a method that removes this comment from the source code after Sonarqube-repair repairs it. However, this method only works properly when we have the simple comment `// Noncompliant`, and test cases fail when we have a "bigger" comment like `// Noncompliant {{Cast one of the operands of this addition operation to a "long".}}`. This PR proposes a fix for that.

### DIFF
--- a/src/main/java/TestHelp.java
+++ b/src/main/java/TestHelp.java
@@ -87,7 +87,7 @@ public class TestHelp {
             while ((line = file.readLine()) != null) {
                 if(line.contains(complianceComment)){
                     line.trim();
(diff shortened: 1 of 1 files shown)


monperrus:
Thanks a lot Fernanda, + looking at the build of this PR, found #45

fermadeiral:
Thank you for merging, @monperrus. I will investigate the recently discovered issue.

Issue #45, opened 2020-03-10
errors in CI that should be break the build


monperrus:
There are errors in CI https://travis-ci.com/kth-tcs/sonarqube-repair/builds/15260387 starting with

```
org.json.JSONException: JSONObject["line"] not found.
	at org.json.JSONObject.get(JSONObject.java:471)
	at Bug.init(Bug.java:35)
	at Bug.<init>(Bug.java:22)
	at Bug.createSetOfBugs(Bug.java:71)
```

We need to improve the tests such that those errors break the build.

fermadeiral:
The build is passing because the test cases are passing and the expected behavior for those test cases is right. I don't believe that making the build break is the good solution in this case. However, I created a fix so that those errors do not happen anymore (see PR #46).

monperrus:
fixed by #46

monperrus:
Thanks Fernanda.

PR #46, opened 2020-03-10
Fix exceptions being thrown while executing 3 tests.

fermadeiral:
This PR intends to fix the errors reported in #45.


monperrus:
Thanks!

PR #58, opened 2020-03-26
Refactor TestHelp

fermadeiral:
`TestHelp` is a class with major things (such as the core methods) and a minor thing (the method that is actually a test helper). It also contains dead code and it is located in `src/main/java` instead of `src/test/java`. This PR refactors `TestHelp` to make the things much easier to understand. The commit messages in this PR explains nicely what was changed.


monperrus:
Thanks Fernanda.

PR #60, opened 2020-03-27
Organize class files

fermadeiral:
This PR changes the organization of the class files. This is mainly due to the fact that all class files were directly in `src/main/java`, and the association between some classes was not good. Now, we have an organization in packages and subpackages, and the structure of the project is clear on which processors are spoon-based and sonar-based. The commit messages in this PR explains what was changed step-by-step.


monperrus:
Thanks a lot Fernanda

PR #70, opened 2020-04-03
Review test resources

fermadeiral:
It's important that our processors work well at least on the cases presented by Sonar, the ones that are basic, which can be found in https://rules.sonarsource.com/java and https://github.com/SonarSource/sonar-java/tree/master/java-checks-test-sources. This PR reviews our test resources according to those sources.

After it, I found a bug in `BigDecimalDoubleConstructorProcessor`, which was exposed by new test examples added from Sonar by this PR. The thing is: `BigDecimalDoubleConstructorProcessor` works only on the constructor `BigDecimal(double)`, which failed in the following new test cases:

```java
MathContext mc;
BigDecimal bd4 = new BigDecimal(2.0, mc); // Noncompliant {{Use "BigDecimal.valueOf" instead.}}
BigDecimal bd6 = new BigDecimal(2.0f, mc); // Noncompliant {{Use "BigDecimal.valueOf" instead.}}
```

This PR already adds the fix for that. See, now `BigDecimalDoubleConstructorProcessor` produces two types of patches.

Already existing one:
```diff
-       BigDecimal bd2 = new BigDecimal(1.1); // Noncompliant
+       BigDecimal bd2 = BigDecimal.valueOf(1.1); 
```

New one:
```diff
        MathContext mc;
-       BigDecimal bd4 = new BigDecimal(2.0, mc); // Noncompliant {{Use "BigDecimal.valueOf" instead.}}
+       BigDecimal bd4 = new BigDecimal("2.0", mc);
```

### DIFF
--- a/docs/HANDLED_SONARQUBE_RULES.md
+++ b/docs/HANDLED_SONARQUBE_RULES.md
@@ -51,6 +51,17 @@ Example:
 +        BigDecimal bd3 = BigDecimal.valueOf(f); 
 ```
 
+When the constructor of `BigDecimal` being called has two arguments, being the first one of type `float` or `double`, that argument is changed to `String`.
+
+Example:
+```diff
+        MathContext mc;
+-       BigDecimal bd4 = new BigDecimal(2.0, mc); // Noncompliant {{Use "BigDecimal.valueOf" instead.}}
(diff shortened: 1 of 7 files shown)


monperrus:
Thanks Fernanda

PR #79, opened 2020-04-17
Create SQRAbstractProcessor, which extends spoon.processing.AbstractProcessor

fermadeiral:
This PR adds `SQRAbstractProcessor` (SQR = SonarQube-Repair), which extends `spoon.processing.AbstractProcessor`. The motivation for that is that our processors were extending `spoon.processing.AbstractProcessor` directly, without a single, unified super class on sonarqube-repair side for all processors. In this way now, the project is more organized and most important, the same behavior that should be applied to all processors can be implemented only in the super class `SQRAbstractProcessor`, which is the case of the new feature that @henry-lp recently added where the storage of the element being processed is required.


monperrus:
Thanks for the refactoring!

PR #90, opened 2020-04-27
Add SelfAssignementProcessor (SonarRule 1656)

henry-lp:
This PR adds a processor for the rule [https://rules.sonarsource.com/java/type/Bug/RSPEC-1656](Variables should not be self-assigned)

```diff
class SelfAssignement {
   int a,c = 0;
   int[] b = {0};
   int h = 0;
   int[] g = 0;
   SelfAssignementCheckB checkB = new SelfAssignementCheckB();
-  void method(int e,int h) {
-    a = a; // Noncompliant [[sc=7;ec=8]] {{Remove or correct this useless self-assignment.}}
-    this.a = this.a; // Noncompliant
-    b[0] = b[0]; // Noncompliant
-    b[fun()] = b[fun()]; // Noncompliant
-    int d = 0;
-    d = d; // Noncompliant
-    e = e; // Noncompliant
     int[] g = new int[]{ 0 };
-    g[fun()] = g[fun()]; // Noncompliant
-    h = h;
-    checkB.b = checkB.b; // Noncompliant
-    checkB.getSelf().foo = checkB.getSelf().foo; // Noncompliant
-  }
+  void method(int e,int h) {
+    int d = 0;
     int[] g = new int[]{ 0 };
+    this.g[fun()] = g[fun()];
+    this.h = h;
+  }
   int fun() {
     return 0;
   }
}
```


src/main/java/sonarquberepair/processor/SelfAssignementProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+
+		CtExpression<?> leftExpression = candidate.getAssigned();
+		CtExpression<?> rightExpression = candidate.getAssignment();
+		if (rightExpression == null || candidate.getParent(CtAssignment.class) != null) {
+			/* Ignore multiple assigment case*/
+			return false;
+		}
+
+		if (leftExpression.toString().equals(rightExpression.toString())) {
+			if (!candidate.getAssigned().toString().contains("this.") && candidate.getAssigned().toString().contains(".")) {
```
fermadeiral, line 43: Why do you need the two checks in lines 42 and 43?

The checking with sonarjava should be already enough, no?

docs/HANDLED_RULES.md
```diff
@@ -259,6 +260,42 @@ class SynchronizationOnGetClass {
...
+-    b[fun()] = b[fun()]; // Noncompliant
+-    int d = 0;
+-    d = d; // Noncompliant
+-    e = e; // Noncompliant
+-    g[fun()] = g[fun()]; // Noncompliant
+-    h = h;
+-  }
++  void method(int e, int[] g,int h) {
++    this.b[0] = b[0];
++    this.b[fun()] = b[fun()];
```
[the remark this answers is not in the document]
henry-lp: I agree. This bug is fixed in the newest commit :) .

src/main/java/sonarquberepair/processor/SelfAssignementProcessor.java
henry-lp: The problem is sonarjava will also return true for cases such as `a.b = a.b` is not handled by this processor, therefore I only use it as a soft-check then hard check with spoon to filter out cases this processor can handle.

fermadeiral: - Why is there a case that is not handled by this processor? We need to document this.

- The case you just mentioned, `a.b = a.b`, is not related to the check in line 42. The check in line 42 will always be true if it's executed, because sonarjava will not return true if the expressions are not the same.

henry-lp: The motivation is the case `a.b = a.b` is very uncommon (unless you delibrately want to do it) and since it requires to check the class file to look for `CtField` , it might not be in classpath. The most common cases would always occur in the same class file such as `a = a` .

fermadeiral: > the case a.b = a.b is very uncommon

I don't know where you got this information from, but it's your call.

What about the second point I mentioned?

src/main/java/sonarquberepair/processor/SelfAssignementProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+		super(originalFilesPath, new SelfAssignementCheck());
+	}
+
+	@Override
+	public boolean isToBeProcessed(CtAssignment<?,?> candidate) {
+		if (!super.isToBeProcessedAccordingToSonar(candidate)) {
+			return false;
+		}
+
+		CtExpression<?> leftExpression = candidate.getAssigned();
```
fermadeiral, line 31: `leftExpression` is never used.

src/main/java/sonarquberepair/processor/SelfAssignementProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+
+		CtThisAccess access = factory.createThisAccess(type.getReference());
+		CtFieldRead<?> fieldRead = factory.createFieldRead();
+
+		fieldRead.setTarget(access);
+		CtExpression<?> leftExpression = element.getAssigned();
+		CtExpression<?> rightExpression = element.getAssignment();
+		CtExpression<?> leftExpression2Check;
+		CtExpression<?> rightExpression2Check;
+		boolean isArrayAccess = false;
```
fermadeiral, line 60: `isArrayAccess` is never used.

src/main/java/sonarquberepair/processor/SelfAssignementProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+		if (leftExpression instanceof CtArrayAccess && rightExpression instanceof CtArrayAccess) {
+			leftExpression2Check = ((CtArrayAccess)leftExpression).getTarget();
+			rightExpression2Check = ((CtArrayAccess)rightExpression).getTarget();
+			isArrayAccess = true;
+		} else {
+			leftExpression2Check = leftExpression;
+			rightExpression2Check = rightExpression;
+		}
+		System.out.println(element);
+		System.out.println(leftExpression2Check + " " + rightExpression2Check);
```
fermadeiral, line 70: Remove those `println`.

src/main/java/sonarquberepair/processor/SelfAssignementProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+import spoon.reflect.factory.Factory;
+
+import spoon.reflect.code.CtExpression;
+import spoon.reflect.code.CtAssignment;
+import spoon.reflect.code.CtThisAccess;
+import spoon.reflect.code.CtFieldRead;
+import spoon.reflect.code.CtFieldAccess;
+import spoon.reflect.code.CtArrayAccess;
+import spoon.reflect.code.CtVariableAccess;
+import spoon.reflect.code.CtTargetedExpression;
```
fermadeiral, line 14: Unused import.

src/main/java/sonarquberepair/processor/SelfAssignementProcessor.java
henry-lp: Right that is fixed :) .

henry-lp: @fermadeiral I just checked apparently sonarjava also return true even when the expression does not look the same . I need both left and right to look the same to fix in this case :) . therefore line 42 need to kept.

henry-lp: The new updated processor now also handle `a.b = a.b` or `a.getSelf().b = a.getSelf().b` by deleting it once it's seen.

fermadeiral: > I just checked apparently sonarjava also return true even when the expression does not look the same, for instance this.a = a regarded as a false negative in their test file.

That is pretty bad, thank you for letting me know.

> The new updated processor now also handle a.b = a.b or a.getSelf().b = a.getSelf().b by deleting it once it's seen.

Perfect. I will review the entire changes now.

henry-lp: no problem . There were so many details regarding this particular rules. I am glad you asked. >(

docs/HANDLED_RULES.md
```diff
@@ -259,6 +260,44 @@ class SynchronizationOnGetClass {
 ```
 
 -----
+
+#### Variables should not be self-assigned ([Sonar Rule 1656](https://rules.sonarsource.com/java/type/Bug/RSPEC-1656))
+
+Any assignment with identical left and right expressions will be processed. If the identifier being using in the self-assignment exists as both a local variable and a field, then the left expression will be changed by adding this. at the beginning of the expression. In any other case, including cases where there are invocations or access to another class field, such as objectA.b = objectA.b, the assignment will be removed.
```
fermadeiral, line 266: "being using" -> "being used"
"this." -> "`this.`"
"objectA.b = objectA.b" -> "`objectA.b = objectA.b`"

fermadeiral:
@henry-lp, everything seems great, I just left one comment for simple fixes in the documentation. Once you address it, I will merge it, without waiting for Travis since no change in source code will be made.

henry-lp:
@fermadeiral Thanks for the review , Fernanda :). I've fixed it.

fermadeiral:
@henry-lp, thank you very much for this processor, this one was a difficult one!

henry-lp:
@fermadeiral thanks for all the help as usual :) .

Issue #91, opened 2020-04-28
how to pass multiple ruleKeys as argument to the jar ?


kiran-chinnapa:
I want to fix multiple issues example:
java -jar target/sonarqube-repair-1.1-SNAPSHOT-jar-with-dependencies.jar rule1 rule2 rule3 ....

Does the API allow me to do that ?

fermadeiral:
Hi @kiran-chinnapa,

I am happy about you considering using our tool!

> Does the API allow me to do that ?

Not yet, but we are working on it. I will write you news when we solve this.

fermadeiral:
Hi @kiran-chinnapa,

> Does the API allow me to do that ?

Now it does! Please, check out our [README](https://github.com/SpoonLabs/sorald/blob/master/README.md) for instructions.

If you experience any problem, please report it in this repository.

kiran-chinnapa:
Thanks a lot, we will use it and get back if there are issues.

PR #95, opened 2020-04-29
Allow multiple rule keys as input

fermadeiral:
This PR implements the feature required in #91.

See an example of transformation by 3 different processors, all applied to the same file in a single call to sonarqube-repair:

```diff
- BigDecimal bd2 = new BigDecimal(1.1); // Noncompliant by BigDecimalDoubleConstructorCheck
+ BigDecimal bd2 = BigDecimal.valueOf(1.1); 
 
- float twoThirds = 2/3; // Noncompliant by CastArithmeticOperandCheck
+ float twoThirds = (float) 2/3; 
 
  AtomicInteger aInt1 = new AtomicInteger(0);
  AtomicInteger aInt2 = new AtomicInteger(0);
- isEqual = aInt1.equals(aInt2); // Noncompliant by EqualsOnAtomicClassCheck
+ isEqual = aInt1.get() == aInt2.get(); 
```

The general idea is to receive multiple rule keys as args and, for each rule key, to create a `Launcher` with the processor corresponding to the rule key and run it. The support of multiple rules is like that because 1) all files processed by a processor (including changed and non-changed) need to be given as input to the next processor, and 2) we have the option of outputting all files or only the changed ones (see `FileOutputStrategy`). Because of all of that combined, we need an intermediate dir when multiple processors will run, which will contain all the files (changed or not changed) after all processors, except after the last one, because then we generate files in the final output path according to the `FileOutputStrategy` chosen. I added some comments in the source code that might help to understand why some things are happening.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,12 @@ The arguments are the following:
 
   [-h|--help]
 ```
+
+Example of a concrete call to sonarqube-repair, in which multiple rule keys are given as input:
+
+```bash
+$ java -jar target/sonarqube-repair-1.1-SNAPSHOT-jar-with-dependencies.jar --originalFilesPath src/test/resources/MultipleProcessors.java --workspace /tmp/ --ruleKeys 2111,2184,2204
+```
  
 ##### If you want to run it on GitHub projects to propose PRs with fixes
 
--- a/src/main/java/sonarquberepair/DefaultRepair.java
+++ b/src/main/java/sonarquberepair/DefaultRepair.java
@@ -2,6 +2,7 @@ package sonarquberepair;
 
 import java.lang.reflect.Constructor;
 import java.io.File;
+import java.lang.reflect.InvocationTargetException;
 import java.util.List;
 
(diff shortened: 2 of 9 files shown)


fermadeiral:
This PR is ready for review.

fermadeiral:
> What about documenting how to use this feature in the README?

You are totally right. However, our README has not been updated at all regarding the usage of the tool. Since I don't want to create changes in this current PR that are not related to its purpose, I created a new PR for updating the README (see https://github.com/SpoonLabs/sonarqube-repair/pull/102). Once the new PR is merged, I will rebase this current PR and I will update the README with this new feature.

monperrus:
> Once the new PR is merged, I will rebase this current PR and I will update the README with this
new feature.

ack! thanks.

fermadeiral:
@monperrus, the update in the `README.md` is done.

monperrus:
Thanks Fernanda!

PR #99, opened 2020-05-01
[Fix] Dependency bugfixes

henry-lp:
This PR address #94 .

With this we don't need to compile and import local dependency from [https://github.com/fermadeiral/sonar-java](https://github.com/fermadeiral/sonar-java) anymore. 

@fermadeiral counting on you doublechecking this  :) . It passes all the tests but there might be hidden problems .
I doublechecked with Repairnator and it worked fine .

PR #101, opened 2020-05-04
Fix broken jar

### LINKED ISSUE #94: Broken Jar
The `jar-with-dependencies` does not work anymore. 

Bug reproduction steps:
```
git clone git@github.com:SpoonLabs/sonarqube-repair.git
cd sonarqube-repair
./init.sh
mvn package -DskipTests
java -jar target/sonarqube-repair-1.1-SNAPSHOT-jar-with-dependencies.jar --originalFilesPath src/test/resources/ --rulesKey 2116
```

Error output:
```
WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance.
Exception in thread "main" java.lang.ExceptionInInitializerError
	at sonarquberepair.DefaultRepair.repair(DefaultRepair.java:33)
	at sonarquberepair.Main.start(Main.java:84)
	at sonarquberepair.Main.main(Main.java:90)
Caused by: java.lang.UnsupportedOperationException: No class provided, and an appropriate one cannot be found.
	at org.apache.logging.log4j.LogManager.callerClass(LogManager.java:576)
	at org.apache.logging.log4j.LogManager.getLogger(LogManager.java:601)
	at org.apache.logging.log4j.LogManager.getLogger(LogManager.java:588)
	at spoon.Launcher.<clinit>(Launcher.java:589)
	... 3 more
```

On first sight it could very possibly be a dependency collision between `LogManager` of `sonarjava` and `spoon-core`  by checking line `at spoon.Launcher.<clinit>(Launcher.java:589)` . Since `spoon-core` use log4j , its [LogManager](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/LogManager.html) can use the method `LogManager.getLogger()` without any arguments, while `sonarjava` use slf4j , its [LogManager](http://www.slf4j.org/api/org/apache/log4j/LogManager.html) need at least 1 argument causing the error above.

WDYT ? @fermadeiral , @monperrus  :)

fermadeiral:
This PR intends to fix #94.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,7 @@ jdk:
 
 before_install:
   - ./init.sh
+
+script:
+  - java -jar /home/travis/build/SpoonLabs/sonarqube-repair/target/sonarqube-repair-1.1-SNAPSHOT-jar-with-dependencies.jar --originalFilesPath /home/travis/build/SpoonLabs/sonarqube-repair/src/test/resources/ArrayHashCodeAndToString.java --ruleKeys 2184
+  - mvn test
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,7 @@
         <dependency>
             <groupId>fr.inria.gforge.spoon</groupId>
             <artifactId>spoon-core</artifactId>
-            <version>8.2.0-beta-2</version>
+            <version>8.1.0-beta-2</version>
         </dependency>
         <!-- https://mvnrepository.com/artifact/org.sonarsource.java/java-checks-testkit -->
         <dependency>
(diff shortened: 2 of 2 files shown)


fermadeiral:
@henry-lp, now it's your turn to double-check. For now, forget about the integration with Repairnator. Let's first have a working jar, and then we see other issues later.

To sum up here, I found that the broken jar problem started when I upgraded Spoon, which was before starting using sonar-java to help in the detection of violations. So let's just use an older version of Spoon for the moment.

WDYT?

henry-lp:
@fermadeiral can you also try downgrade spoon in #99 and check if it also works ? :)

fermadeiral:
@henry-lp, *please*, read again what I said in the other PR and in this one.

henry-lp:
ok, then in that case, I have no problem merging this . WDYT ? :)

monperrus:
Thanks a lot for the fix and the constructive discussion on this tough problem.

fermadeiral:
@monperrus, thank you for merging.

PR #111, opened 2020-05-31
Feat: skip duplicated types during spoon model building

henry-lp:
A project might have sub projects or files in some resources folder containing classes of the same name as another class in another location. This PR suggests that we ignore loading the same type twice by one-by-one loading in each files into `SpoonLauncher` while skipping the type already occured to improve usability for the CI integration by avoiding `spoon.compiler.ModelBuildingException`. :)

what this PR does:
* Move all buggy files into a new folder in the resources folder
* create a new folder for duplicated files in the resources folder
* Add one-by-one loading logic into `SpoonLauncher`
* Add JSAP option to activate that.

@monperrus @fermadeiral

### DIFF
--- a/src/main/java/sonarquberepair/DefaultRepair.java
+++ b/src/main/java/sonarquberepair/DefaultRepair.java
@@ -101,6 +101,7 @@ public class DefaultRepair {
 		launcher.addInputResource(inputDirPath);
 		launcher.setSourceOutputDirectory(outputDirPath);
 		launcher.getEnvironment().setAutoImports(true);
+		launcher.getEnvironment().setIgnoreDuplicateDeclarations(true);
 		if (this.config.getPrettyPrintingStrategy() == PrettyPrintingStrategy.SNIPER) {
 			launcher.getEnvironment().setPrettyPrinterCreator(() -> {
 						SniperJavaPrettyPrinter sniper = new SniperJavaPrettyPrinter(launcher.getEnvironment());


src/main/java/sonarquberepair/DefaultRepair.java
```diff
@@ -98,7 +104,27 @@ public static boolean deleteDirectory(File directoryToBeDeleted) {
...
+								}
+							});
+				System.out.println("Done loading.");
+			} catch (IOException e) {
+				launcher.addInputResource(inputDirPath);
+			}
+		} else {
+			launcher.addInputResource(inputDirPath);
+		}
+
```
fermadeiral, line 127: What about `launcher.getEnvironment().setIgnoreDuplicateDeclarations(true);` instead of all those things?

src/main/java/sonarquberepair/Constants.java
```diff
@@ -18,6 +19,8 @@
 	public static final String PATCHES = "SonarGitPatches";
 	public static final String PATCH_FILE_PREFIX = "sonarpatch_";
 	public static final String PATH_TO_RESOURCES_FOLDER = "./src/test/resources/";
+	public static final String PATH_TO_BUGGY_FILES = "./src/test/resources/buggyfiles/";
+	public static final String PATH_TO_DUPLICATED_FILES = "./src/test/resources/duplicatedfiles/";
```
fermadeiral, line 23: Why do we need these two different paths?

src/main/java/sonarquberepair/Constants.java
```diff
@@ -9,6 +9,7 @@
 	public static final String ARG_ORIGINAL_FILES_PATH = "originalFilesPath";
 	public static final String ARG_WORKSPACE = "workspace";
 	public static final String ARG_GIT_REPO_PATH = "gitRepoPath";
+	public static final String ARG_SKIP_DUPLICATED_TYPES = "skipDuplicatedTypes";
```
fermadeiral, line 12: Why do we need this? We should just skip the duplicated types every time, no?

src/main/java/sonarquberepair/DefaultRepair.java
henry-lp: Nice . I'll change so that this options will always be true for the launcher.

henry-lp:
> What about `launcher.getEnvironment().setIgnoreDuplicateDeclarations(true);` instead of all those things?

convenience command !  :)

fermadeiral:
Thank you, Henry!

Issue #74, opened 2020-04-15
Use the sniper pretty printing strategy, and if it does not work, try again with the normal strategy.


fermadeiral:
We have two pretty printing strategies, sniper and normal. We currently let the user choose which strategy she wants to use. However, the sniper strategy is very desired, and failure might happen using it, where no repair at all is produced. We should automatically do a second repair trial with the normal strategy in case of failure with the sniper strategy.

fermadeiral:
After discussing with @monperrus and later with @henry-lp, we want now to only use the Sniper printer.

I suggest the following strategy:

1) Currently, Sorald accepts as input an argument for choosing the printer (normal or sniper), where the normal printer is the default. We should keep the normal printer in Sorald, because we might need to switch to the normal printer in some experiment, but the default will now be the sniper printer.

2) All test cases that pass with the sniper printer should call Sorald with no argument about the printer (since the sniper printer will become the default), and all test cases that fail should call Sorald with the argument about the printer, passing the normal printer (this is temporarily).

3) Each test case calling Sorald with the argument about the printer (the ones failing with the sniper printer) should be investigated, and issues should be created in https://github.com/INRIA/spoon/.

4) In Sorald, we should manage to build Spoon, using the latest version, since the Spoon team will be fixing stuff that we need, possibly before they release Spoon with the fixes.

PR #120, opened 2020-06-08
Use sniper printer as the default printer

fermadeiral:
This PR fixes points 1 and 2 in https://github.com/SpoonLabs/sorald/issues/74#issuecomment-639464652.

Note: some test cases do not actually fail with sniper but yet the normal printer is used because the output code is somewhat wrong or with style problems that we might want to report to the Spoon team anyways.


monperrus:
LGTM. Ready to merge?

fermadeiral:
> Ready to merge?

Yes.

monperrus:
Thanks Fernanda!

PR #122, opened 2020-06-09
Fix(config): do not allow that the same processor runs more than once

### LINKED ISSUE #110: Should not repair the same rule twice
Given rules for repair as 2116,2116 . It should only repair once :) .

fermadeiral:
This PR fixes #110.

@henry-lp WDYT?


henry-lp:
It's great :) . LGMT

PR #123, opened 2020-06-12
Build and use spoon snapshot

fermadeiral:
This PR addresses the point 4 described in https://github.com/SpoonLabs/sorald/issues/74#issuecomment-639464652

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ script.sh
 sorald-workspace/
 FileOutputStrategyTest/
 sonar-java/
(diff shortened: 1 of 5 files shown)


monperrus:
Instead of self-compiling, I would be in favor of depending on a beta version (more simple, esp for deployment to Sonatype).

FYI, there is one beta version per week (every Saturday) and we can also do it one in between on demand.

fermadeiral:
@monperrus great.

Issue #125, opened 2020-06-13
Array out of bound exception rule key 2184


henry-lp:
Error when using rule `2184: Math operands should be cast before assignment`  when applying on the following projects
* https://github.com/eclipse/kura
* https://github.com/eclipse/hono

Reproduction:
```
git clone https://github.com/eclipse/hono Repo4Analyze

java -jar target/sorald-1.1-SNAPSHOT-jar-with-dependencies.jar --originalFilesPath Repo4Analyze/ --ruleKeys 2184

```

@fermadeiral WDYT ? :)

fermadeiral:
> @fermadeiral WDYT ? :)

I think that it would be very helpful if you provide me a failing test case, and then I can search for a fix.

henry-lp:
@fermadeiral 
I plan to add a test file and then make a PR . Are you fine with that or would you prefer me to do it in another way ? :)

fermadeiral:
Feel free to proceed, but when it's like that, assign yourself for the issue so I know you are working on it.

Don't forget: do a *minimal* test.

henry-lp:
I just found out that this bug happens when I use the jar file after packaging. Junit test only seems to either pass or report it as a `NullPointerException` . I'll make travis run a script instead .

fermadeiral:
> Junit test only seems to either pass or report it as a NullPointerException 

This doesn't seem precise.

Henry, provide also the test case. We should be able to check the difference between calling the jar or not. **Please**, the resource for the test needs to be minimal, preferred only the file where the error happens.

slarse:
This crash occurs a _lot_, see #337. @fermadeiral do you want me to get on fixing this or do you think you'll get around to it sometime soon?

fermadeiral:
@slarse I will get around to it soon.

> This crash occurs a lot, see #337.

By reading the list that @khaes-kth added in that issue I don't see anything specifically related to this issue. Could you point me exactly where in the issue I can find the crashes related to this issue, please?

slarse:
@fermadeiral Sorry, I had relabeled the errors related to that bug, as they did not in fact cause the total failure of Sorald that @khaes-kth was reporting on (crashes within processors are caught and handled). See this comment: https://github.com/SpoonLabs/sorald/issues/337#issuecomment-764592973

Also, here's an excerpt from an email on the issue (it's in the "Sorald: RQ1 experiments" email thread"):

### EMAIL START

[...]

PR #126, opened 2020-06-13
Fail test case for #125

henry-lp:
Failing test case for issue #125 

@fermadeiral


fermadeiral:
I have some issues with this PR.

First, we should be able to have an actual test case to expose the issue.

Second, why does the script run the same rule in two projects? If there are issues with both projects and they are the same, we need only one project to expose the issue, and if the issues are different, we need one PR for each.

Third and final, please, don't clone the entire project. Put in our resources only the file where the issue is happening and pass this file to Sorald. Cloning entire projects to test one specific issue with one processor is not really good, we need only the file that exposes the issue. Moreover, having a minimal testing resources also helps to debug and to create a fix.

PR #127, opened 2020-06-14
Update spoon version

fermadeiral:
The latest Spoon beta version contains multiple fixes in the sniper printer, thus we should use it.


monperrus:
Thanks!

Issue #129, opened 2020-06-19
Segmented Repair


henry-lp:
Many projects such as (https://github.com/eclipse/eclipse.platform.ui) contains many sonar warnings but we are unable to repair them due to various reasons (below are some of them - issue  #124 , #125) 
* The project is too big (out of heap memory)
* File parsing error causing null pointer exception 

Therefore we need to somehow in a nice way split these projects up in a way such that we can still repair the non-error part as big as possible in such cases. Thus, segment repair is proposed by me , where the ideal goal is to repair as good as the default repair (number of fixes are the same) and allow easy rollback and recovery if the repair attempt failed and savage the situation somehow instead of complete fail.

@monperrus @fermadeiral

henry-lp:
Next I will experiment with the incremental launcher to forge a rollback protocol :) .

fermadeiral:
@henry-lp, should we close this issue?

henry-lp:
Yes, I can close it .

PR #130, opened 2020-06-19
Feat: add single rule SegmentRepair mode

henry-lp:
Issue: #129 
This PR add a new repair mode . How it works:
* Build project file structure as a tree in-memory.
* Segment the tree into parts 
* build model and repair each parts. 

Currently if failure of a segment fails , it will ignore it and proceed with the next segment. This mode only supports single rule repair for now. 

I have test this mode and found that it enable sorald to repair big projects such as (https://github.com/eclipse/eclipse.platform.ui) if the segment size is 100 .


README.md
```diff
@@ -75,6 +76,18 @@ The arguments are the following:
...
+        Max number of files per loaded segment for segmented repair. It should
+        be >= 3000 files per segment. Default: 6500 (256mb Jvm) .  (default:
+        6500)
+
+  [--repairStrategy <repairStrategy>]
+        Type of repair strategy. DEFAULT - load everything without splitting up
+        the folder in segments, SEGMENT - splitting the folder into smaller
+        segments and repair one segment at a time (need to specify
+        --maxFilesPerSegment if not default). Default: DEFAULT (default:
+        DEFAULT)
```
fermadeiral, line 89: I have three points to mention here (that should be fixed first in the code, and then here, with the output from the code):

- You shouldn't write the default values because they are already shown automatically.

- The arg `repairStrategy` should appear before `maxFilesPerSegment` since first one decides the strategy for the repair, and if the segmented one is chosen, one sets the max files per segment.

- Why should the max files per segment be >= 3000?

fermadeiral:
@henry-lp, there is a huge duplication between `DefaultRepair` and the new class `SegmentedRepair`. This should be fixed before I do a deep review.

README.md
henry-lp: Fixed :).

henry-lp:
@fermadeiral I have made some adjustments as you suggested. Ready for some reviews again :)

fermadeiral:
This PR is doing more than one thing. This already happened several times in the past and I already told you to not do it. I will review this PR when there are changes for only one thing.

henry-lp:
What serveral thing do you mean in this PR ? I think I only added `single rule SegmentRepair mode` and then refactor according to your comments.

henry-lp:
I see now that I also pushed my own measurements code parts in this PR. I'll double check everything and ping you again later.

henry-lp:
@fermadeiral I have remove the redundant parts I found in the code and formatted it with Intellij . Can you spare another moment to see if I need to check again to remove some other parts?

henry-lp:
Close for smaller PRs.

Issue #141, opened 2020-10-13
Reduce code duplication in processor test suite with parameterized tests


slarse:
I've been poking around the test suite for the processors (i.e. `src/test/java/sorald/processor`), and I've found that there is a whole lot of code duplication. Most of the test classes are borderline identical, and generally follow this pattern:

```java
public class SomeCheckProcessorTest {
    @Test
    public void test() throws Exception {
        String fileName = "SomeFile.java";
        String pathToBuggyFile = Constants.PATH_TO_RESOURCES_FOLDER + fileName;
        String pathToRepairedFile = Constants.SORALD_WORKSPACE + "/" + Constants.SPOONED +"/" + fileName;

        JavaCheckVerifier.verify(pathToBuggyFile, new SomeCheck());
        Main.main(new String[]{
                Constants.ARG_SYMBOL + Constants.ARG_ORIGINAL_FILES_PATH,pathToBuggyFile,
                Constants.ARG_SYMBOL + Constants.ARG_RULE_KEYS,"<KEY>",
                Constants.ARG_SYMBOL + Constants.ARG_WORKSPACE,Constants.SORALD_WORKSPACE});
        TestHelper.removeComplianceComments(pathToRepairedFile);
        JavaCheckVerifier.verifyNoIssue(pathToRepairedFile, new ArrayHashCodeAndToStringCheck());
    }
}
```

There is some minor variation in the arguments to `Main.main` and variable names, but most classes are otherwise semantically identical. If Sorald is to scale with more rules and more tests, I think this must be revamped.

As a solution, I propose that we write a single [parameterized test](https://www.baeldung.com/parameterized-tests-junit-5), and keep a directory with "broken" Java files (such as the current `src/test/resources` directory). Then, each file can be automatically turned into a separate test case. If we for a moment ignore the problem with the variation in command line arguments, a test case file could be named as `SomethingInformative_<SONAR_CUBE_KEY>.java`, and be used to generate a test case for the rule with key `<SONAR_CUBE_KEY>` (e.g.  `EqualsOnAtomicClass_2204.java` to test rule 2204).

I have yet to figure out the best solution for the variation in arguments to `Main.main`, but I'm sure a reasonable solution can be found through some trial and error.

[...]

PR #144, opened 2020-10-14
test: Parameterize processor tests

slarse:
This is a draft (i.e. not merge-ready yet) of the parameterization of the processor tests as proposed in #141.

Good news: it works! To add a new test case for a rule, you have to do the following.

1. Create a directory with the same name as the sonar checker related to the rule.
    - For example, for the rule related to `MathOnFloatCheck`, the directory should be called `MathOnFloat`
    - Of course, if said directory already exists, then skip this step.
2. Put any number of Java source files in the checker-named directory. Each Java file results in one test case.

See [the directories here](https://github.com/slarse/sorald/tree/cca28511da210378ffa1cfbdb7a6862730cafee3/src/test/resources/processor_test_files) for examples.

The test logic is all contained in [ProcessorTest.java](https://github.com/slarse/sorald/blob/issue/141-parameterize-processor-tests/src/test/java/sorald/processor/ProcessorTest.java). The actual test case should look very familiar. Then there's a bunch of supporting functionality in [ProcessorTestHelper.java](https://github.com/slarse/sorald/blob/issue/141-parameterize-processor-tests/src/test/java/sorald/processor/ProcessorTestHelper.java) that assists in extracting test cases from the aforementioned "checker-named" directories.

There are two problems I need to resolve.

First, Maven's surefire plugin doesn't appear to respect the display names of the tests. When running with IntelliJ's test runner, the parameterized test names are displayed on the form `[1] ruleKey=2164 ruleName=MathOnFloat source=path/to/sourc.java`. However, Maven just displays something along the lines of `testProcessSingleFile[1]`, which is next to useless on a test failure.

Second, the [sonar check class is loaded very insecurely](https://github.com/slarse/sorald/blob/cca28511da210378ffa1cfbdb7a6862730cafee3/src/test/java/sorald/processor/ProcessorTestHelper.java#L71-L72), which is a bit ironic considering the current goals of this project. I'm thinking I can extract the TEMP_SONAR_CHECKS into a lookup table and use that instead: https://github.com/SpoonLabs/sorald/blob/7a150bf970eddd5306a40805c2c9ae2f892b7afa/src/main/java/sorald/miner/MineSonarWarnings.java#L50

As a final note, as I'm redoing a large part of the test suite, I think we should set up code coverage before this is merged. It would be a shame to lose a bunch of testing because I mess up. I'll do a separate PR for that.

Input is welcome :)


slarse:
Oh, and this PR looks huge, but that's because I've copied a bunch of the Java test source files. I will eventually remove the duplicates as the parameterization is meant to replace the old tests. It's really only `ProcessorTest` and `ProcessorTestHelper` that are new additions.

slarse:
**Note to self** on coverage before conversion

* Statement coverage: 90%
* Branch coverage: 74%
* Mean coverage: 83%

slarse:
@fermadeiral Alright, this is finally ready for review :). It's basically what I stated in the initial comment on this PR, but there are some slight differences in the details.

* Test logic in [ProcessorTest.java](https://github.com/slarse/sorald/blob/issue/141-parameterize-processor-tests/src/test/java/sorald/processor/ProcessorTest.java)
    - The processors/checks for which the sniper printer causes problems are hard-coded into the test, it was the simplest solution (and they should not exist, really, will open issues about them later).
* Helper methods in [ProcessorTestHelper.java](https://github.com/slarse/sorald/blob/issue/141-parameterize-processor-tests/src/test/java/sorald/processor/ProcessorTestHelper.java)
* Resources for the parameterized processor test are in [`src/test/resources/processor_test_files`](https://github.com/slarse/sorald/blob/issue/141-parameterize-processor-tests/src/test/resources/processor_test_files)
    - Each file in the subdirectories of `processor_test_files` results in a single test case with a single rule. The names of the files don't matter.
    - Each subdirectory starts with `<RULE_KEY>_`, e.g. `2142_InterruptedException`. The test helper extracts the key from the directory name and uses the corresponding rule for each Java file contained in that directory. Anything after `<RULE_KEY>_` in the directory name doesn't matter, but I just put the processor's name.
* All previous processor tests have been removed, except for:
    - NoSonarTest: This for some reason tests both the rule, and that sorald respects the "max amount of fixes" thing.
    - MultipleProcessorsTest: This tests multiple rules at the same time, and so doesn't fit the pattern of the other tests. I think it makes sense to have this on its own. We could generalize the parameterized test to test multiple rules (e.g. have the directory prefix be (`<RULE_KEY_1>_<RULE_KEY_2>...`), but that I feel is for a future PR.

To add a new test for a rule with existing tests, simply plop a new Java file into any of the existing subdirectories. To add tests for a rule without existing tests, create a new directory according to the naming scheme above, and then plop a new Java file into that.

**There's one thing I haven't done yet**, and that is to update `docs/CONTRIBUTING.md` with the new process for adding test cases. I figured that could be a separate PR, as this one is getting so large and has lived for so long. WDYT (on that and this PR as a whole)?

slarse:
Ping @fermadeiral

fermadeiral:
@slarse, thank you very much for this big improvement!

> There's one thing I haven't done yet, and that is to update docs/CONTRIBUTING.md with the new process for adding test cases. I figured that could be a separate PR

I will wait for that!

PR #145, opened 2020-10-14
Test: Add code coverage

slarse:
This PR adds code coverage with Jacoco and Codecov in preparation for #141. I think it's a good idea to have that seeing as many tests will be migrated to a new format, and it'd be easy to drop one or two by accident.

For everything to work, codecov probably needs to be authorized to the repo. If you'd rather use something else (e.g. coveralls), let me know, I just picked these things because they're what I usually use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,5 @@ jdk:
 script:
   - java -jar /home/travis/build/SpoonLabs/sorald/target/sorald-1.1-SNAPSHOT-jar-with-dependencies.jar --originalFilesPath /home/travis/build/SpoonLabs/sorald/src/test/resources/ArrayHashCodeAndToString.java --ruleKeys 2184
   - mvn test
+
+after_success: bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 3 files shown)


fermadeiral:
@slarse, nice!

@monperrus, could you authorize codecov to this repository? I don't think I can do that because I don't have rights on SpoonLabs and I think it requires that.

Let me know if I am missing something here.

slarse:
Yes, probably @monperrus needs to authorize that. To be clear, the authorization that's needed is for codecov to post coverage to PRs as a check. As the repo is public, we can already view the coverage here: https://codecov.io/gh/spoonlabs/sorald

And it also appears as if I did not mess up with the README badge, as it now shows 83% :)

monperrus:
OK, so we can merge this, and we'll do the codecov config together.

fermadeiral:
> OK, so we can merge this, and we'll do the codecov config together.

Great!

fermadeiral:
Thanks, @slarse!

slarse:
@fermadeiral Great! Now that we've got coverage, I'll turn my attention back to #144

PR #147, opened 2020-10-15
refactor: Extract list of sonar check classes to Constants

slarse:
#141 

As mentioned in #144, I need a more secure way to load the check classes in the parameterized test. As there already is a large list of checks in `MineSonarWarnings`, I figured I can just extract that and build a lookup table from it. It's not an exhaustive list, but adding a new check to it when we need to is trivial. **This PR just extracts the list of checks to `Constants.java`**, and changes it to contain the classes themselves rather than instances.

I created this separate PR as #144 is becoming so large.

On a side note, the `WarningMinerTest` is dependent on the order of the checks in the list, which tripped me up a bit.

### DIFF
--- a/src/main/java/sorald/Constants.java
+++ b/src/main/java/sorald/Constants.java
@@ -194,9 +194,9 @@ public class Constants {
 		sonarChecks.add(NonNullSetToNullCheck.class);
 		sonarChecks.add(CustomUnclosedResourcesCheck.class);
 
-//        TEMP_SONAR_CHECKS.add(new DefaultMessageListenerContainerCheck());
-//        TEMP_SONAR_CHECKS.add(new SingleConnectionFactoryCheck());
-//        TEMP_SONAR_CHECKS.add(new DependencyWithSystemScopeCheck());
+//        sonarsChecks.add(DefaultMessageListenerContainerCheck.class);
+//        sonarsChecks.add(SingleConnectionFactoryCheck.class);
+//        sonarsChecks.add(DependencyWithSystemScopeCheck.class);
 		SONAR_CHECK_CLASSES = Collections.unmodifiableList(sonarChecks);
 	}
 
--- a/src/main/java/sorald/Constants.java
+++ b/src/main/java/sorald/Constants.java
@@ -68,7 +68,7 @@ public class Constants {
 
 	public static final String PATH_TO_STATS_OUTPUT = "experimentation/stats/warnings";
 
-	public static final List<Class<? extends JavaFileScanner>> SONAR_CHECKS;
+	public static final List<Class<? extends JavaFileScanner>> SONAR_CHECK_CLASSES;
 
 	static {
 		List<Class<? extends JavaFileScanner>> sonarChecks = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


slarse:
Hold off on this one for a moment, I might have messed up with generics.

fermadeiral:
@slarse, I have created the labels WiP and ready-for-review to facilitate our communication. Every time I ask for changes I will remove the label ready-for-review and add WiP, and then I will wait for you to do the other way around so I can know that I can review your PR.

If I understood properly, this PR is ready for review, right?

slarse:
@fermadeiral Yes, ready for review.

I don't have access to attach labels, so it's probably easier to use the prefix in the title as either "wip: " for WIP, or anything else for ready-for-review.

monperrus:
> I don't have access to attach labels


added you, should be solved

fermadeiral:
> added you, should be solved

Thank you for that.

src/main/java/sorald/Constants.java
```diff
@@ -45,4 +68,136 @@
...
+		sonarChecks.add(OptionalGetBeforeIsPresentCheck.class);
+		sonarChecks.add(MinMaxRangeCheck.class);
+		sonarChecks.add(ConditionalUnreachableCodeCheck.class);
+		sonarChecks.add(NullDereferenceCheck.class);
+		sonarChecks.add(NonNullSetToNullCheck.class);
+		sonarChecks.add(CustomUnclosedResourcesCheck.class);
+
+//        TEMP_SONAR_CHECKS.add(new DefaultMessageListenerContainerCheck());
+//        TEMP_SONAR_CHECKS.add(new SingleConnectionFactoryCheck());
+//        TEMP_SONAR_CHECKS.add(new DependencyWithSystemScopeCheck());
```
fermadeiral, line 199: Could you please change these commented lines to match the current implementation?

```
//        sonarChecks.add(DefaultMessageListenerContainerCheck.class);
//        sonarChecks.add(SingleConnectionFactoryCheck.class);
//        sonarChecks.add(DependencyWithSystemScopeCheck.class);
```

I don't recall why we have those lines, but I prefer to keep them for now.

slarse: Right, missed those lines. That would have been confusing later, good spot!

slarse:
@fermadeiral I have discovered one potentially unwanted side effect of this refactoring. When a new check is added to the list (for example because there's a processor for which the corresponding check is currently not there), it obviously adds another check to the warnings miner. That causes the output of the warnings miner to change with the addition of a new check, which in turn causes the corresponding test case to fail as it checks for absolute equivalency between a reference output in `src/test/resources/warning_miner/test_results.txt`. Here's an excerpt of that file: https://github.com/SpoonLabs/sorald/blob/ab99cf0fc08f2862717409d4914f425b01f2664d/src/test/resources/warning_miner/test_results.txt#L2-L5

The strict comparison here is also why the order of the checks in the list matters.

I think that if we add a new check, then we do want the warning miner to use that check, and so I think it's the test case that needs to be adjusted. I suggest that (in a separate PR, it's not needed here) we weaken the test case such that it doesn't expect the exact output of `test_results.txt`, but just checks equivalency of non-zero checks (e.g. `CompareStringsBoxedTypesWithEqualsCheck=240` and `NullDereferenceCheck=1` in the excerpt above). Thoughts on that?

Finally, there's a decrease in coverage in this PR, because of this catch: https://codecov.io/gh/SpoonLabs/sorald/compare/ab99cf0fc08f2862717409d4914f425b01f2664d...2974c0b0a03ddee79377cb0e49e2571553a5b65c/diff#D1-36

It's needed because the fetching and calling of the constructor of the class may throw a few checked exceptions, which isn't allowed in a lambda, and I must deal with it somehow in that static context anyway. I could add a contrived test case to trigger the behavior, but it seems more trouble than it's worth.

fermadeiral:
> Edit: See #149 for the proposed weakening

Already merged.

> I could add a contrived test case to trigger the behavior, but it seems more trouble than it's worth.

Don't bother about that!

Thanks for this PR!

PR #149, opened 2020-10-16
test: Weaken the WarningMinerTest to only compare non-zero checks

slarse:
This PR adds the proposed solution to the problem introduced in #147 that adding new checks to the `SONAR_CHECK_CLASSES` list alters the output of the warnings miner in a non interesting way. It also solves the problem that the test is currently dependent on the order of the checks by sorting the output.

This PR is itself independent of #147 and can be merged before or after.


slarse:
I was a bit hasty, there's a li'l bug here.

slarse:
There we go, _now_ it's ready for review! I had forgotten that `String.matches` implicitly adds `^` and `$` around the regex. I've put them in there explicitly to show intent.

I've also removed the zero-results from the output file (see bf749c3) to demonstrate that the revised test actually does not care about zero-results.

fermadeiral:
Nice solution, @slarse!

Issue #82, opened 2020-04-18
work and papers related to sonarqube-repair 


monperrus:

monperrus:
Are SonarQube Rules Inducing Bugs?
https://www.researchgate.net/profile/Valentina_Lenarduzzi/publication/340400027_Are_SonarQube_Rules_Inducing_Bugs/links/5e8c1e8aa6fdcca789fbf3ef/Are-SonarQube-Rules-Inducing-Bugs.pdf

monperrus:
On the fault proneness of sonarqube technical debt violations: A comparison of eight machine learning techniques
https://arxiv.org/pdf/1907.00376

monperrus:
Some SonarQube Issues have a Significant but Small Effect on Faults and Changes. A large-scale empirical study
https://arxiv.org/pdf/1908.11590

monperrus:
How developers engage with static analysis tools in different contexts
https://link.springer.com/article/10.1007/s10664-019-09750-5

slarse:
The master's thesis by Haris http://urn.kb.se/resolve?urn=urn:nbn:se:kth:diva-273924

@HarisAdzemovic I got a lot of value from reading your thesis in my getting started with sorald, and in getting a lay of then land in automatic code repair. Thanks!

monperrus:
FYI

Automatically Customizing Static Analysis Tools to Coding Rules Really Followed by Developers.
https://dblp.org/rec/conf/wcre/UedaIM21

PR #152, opened 2020-10-16
Add academic bibliographic references in the README

fermadeiral:
This PR adds academic bibliographic references in the README. I realized that this information was missing in our repository because of @slarse's comment [here](https://github.com/SpoonLabs/sorald/issues/82#issuecomment-709903413).

Currently, there is only @HarisAdzemovic's thesis, but soon we will have @henry-lp's thesis and other related scientific works.

@monperrus, WDYT?


slarse:
I was actually about to suggest this at the same time as adding that comment, but then I didn't for some reason... As a visitor to the repo, I'd definitely be more impressed with some academic references in the README (and let's face it, the top level README is to roughly 95% an advert for the project).

monperrus:
Great idea thanks @fermadeiral !

we can replace the full bibtex by a link [bibtex] pointing to the bibtex export of Diva.

PR #156, opened 2020-10-19
chore: Update to sonar-java 6.9.0

### LINKED ISSUE #89: update to sonar-java >=6.3
We depend on sonar-java 5.14, we should depend on a recent version, see https://search.maven.org/search?q=org.sonarsource.java

However, as noted by @fermadeiral , there is a dependency conflict to analyze and fix, see https://github.com/SpoonLabs/sonarqube-repair/pull/88#issuecomment-619750105

slarse:
Fix #89 

This PR depends on #144 (that's why it looks huge), I will rebase on master once that's merged. Only b6b3383 and onward is actually new.

The transition to sonar-java 6.9.0 mostly went smoothly, but there were/are a few problems.

First, the interface of the `JavaCheckVerifier`/`MultiFilesJavaCheckVerifier` have been entirely deprecated, and so everything needed to be revamped. I've added an adapter, `Verifier`, that serves as the interface with the currently unstable API of the verifier functionality of sonar-java.

Second, some checks have slightly altered behavior. In particular, `ArrayHashCodeAndToStringCheck` doesn't flag the code in `CodeFactory.java` as non-compliant anymore: https://github.com/SpoonLabs/sorald/blob/ea7d6264d5ad02eb52189fed7cda65ebac516a6d/src/test/resources/CodeFactory.java#L19-L23

I'm not sure if this is a bug, I'll keep investigating.

Third, the miner test fails miserably. I haven't looked at that yet.


slarse:
The miner test fails because of some very strange use of paths in the internals of sonar-java. Essentially, it assumes that the current working directory is the root of the project being analyzed, and then tries to resolve input files relative to it. I'm trying to work around it, but it may require us to patch sonar-java. It's not really designed to be used the way we're using it.

slarse:
Actually, the relative paths thing doesn't appear to be the whole issue. Will need to keep debugging.

slarse:
Now I've figured out the actual problem. We're essentially using the testing tooling for `sonar-java`, and a newly introduced feature is that the `verifyIssues` function looks for two things: lines with non-compliant code, and lines `// Noncompliant` comments. If it finds non-compliant code on a line that lacks such a comment, it will raise an assertion error.

**Key code parts**

* [Here it adds the parser that finds the comments](https://github.com/SonarSource/sonar-java/blob/2e3b7d25cebac51d5040d30dc3048c2f575d6e92/java-frontend/src/main/java/org/sonar/java/testing/InternalCheckVerifier.java#L206)
    - [And here's where the comment extraction happens](https://github.com/SonarSource/sonar-java/blob/2e3b7d25cebac51d5040d30dc3048c2f575d6e92/java-frontend/src/main/java/org/sonar/java/testing/Expectations.java#L298-L301)
* [Here is where the matching between expected issues and actual issues occurs](https://github.com/SonarSource/sonar-java/blob/2e3b7d25cebac51d5040d30dc3048c2f575d6e92/java-frontend/src/main/java/org/sonar/java/testing/InternalCheckVerifier.java#L290-L311)

This entirely defeats our use case, as the code we want to analyze will obviously not have comments marking non-compliant code. I will keep digging a bit to see if there's any way around this behavior, but I don't think we have any other choice than to fork the project. It'd be a pretty low-maintenance fork, though.

fermadeiral:
> I'm trying to work around it, but it may require us to patch sonar-java. It's not really designed to be used the way we're using it.

I agree with you, sonar-java is not designed to be used through the java API (it's designed to be used through the SonarSource products, i.e., SonarQube, SonarCloud, and SonarLint), which is why I believe they might not accept patches from us that change their API, but we can try!

> a newly introduced feature is that the verifyIssues function looks for two things: lines with non-compliant code, and lines // Noncompliant comments. If it finds non-compliant code on a line that lacks such a comment, it will raise an assertion error.

Arghhhh.

> I will keep digging a bit to see if there's any way around this behavior

Ok.

I think it would be nice to discuss this issue by voice (in person or remotely).

slarse:
> I think it would be nice to discuss this issue by voice (in person or remotely).

I agree. We may have time to touch upon it during the sync meeting today at 14. At the very least, we should have time to schedule such a discussion for some other time :)

slarse:
> a newly introduced feature is that the verifyIssues function looks for two things: lines with non-compliant code, and lines // Noncompliant comments. If it finds non-compliant code on a line that lacks such a comment, it will raise an assertion error.

Okay, so there is a very simple fix for this in the testing utilities. Simply change the [method here](https://github.com/SonarSource/sonar-java/blob/2e3b7d25cebac51d5040d30dc3048c2f575d6e92/java-frontend/src/main/java/org/sonar/java/testing/InternalCheckVerifier.java#L226-L237) that currently looks like this:

```java
  private void checkIssues(Set<AnalyzerMessage> issues) {
    if (expectations.expectNoIssues()) {
      assertNoIssues(issues);
    } else if (expectations.expectIssueAtFileLevel() || expectations.expectIssueAtProjectLevel()) {
      assertComponentIssue(issues);
    } else {
      assertMultipleIssues(issues);
    }
    if (customIssueVerifier != null) {
      customIssueVerifier.accept(issues);
    }
  }
```

to this

```java
  private void checkIssues(Set<AnalyzerMessage> issues) {
    if (customIssueVerifier != null) {
      customIssueVerifier.accept(issues);
    }
    if (expectations.expectNoIssues()) {
      assertNoIssues(issues);
    } else if (expectations.expectIssueAtFileLevel() || expectations.expectIssueAtProjectLevel()) {
      assertComponentIssue(issues);
    } else {
      assertMultipleIssues(issues);
    }
  }
```

In other words, invoke the `customIssueVerifier.accept` method before checking. I've not been able to get the tests for `sonar-java` to run because of signature errors on the jdt jarfiles, so I can't tell if it breaks anything.

slarse:
I think a better solution here is to _not_ use the testing methods (e.g. `verifyIssues`) in production, but rather build our own wrappers around the core components. The testing methods do a whole lot more than we want to do. Essentially, all we want `sonar-java` to do for us outside of our unit tests is to scan ASTs for issues, and then we take it from there.

slarse:
I've been messing about with this all day. Essentially, doing this the "proper" way does not seem possible, because there is currently no way to include both [sonarlint-core](https://github.com/SonarSource/sonarlint-core) and [sonar-java](https://github.com/SonarSource/sonar-java) in the same project. The problem is that the former is signed, and the latter isn't, and they have some overlapping packages which causes security exceptions. I've been trying for hours and can't get around it, short of manually removing the signatures from the jarfiles.

There's another way, though, but that still involves using some amount of testing infrastructure. It's the easiest for now, though, so I'll go with that. With a proper facade around the sonarqube stuff, future refactoring should be simple.

src/main/java/sorald/sonar/RuleVerifier.java
```diff
@@ -54,6 +83,78 @@ public static void verifyHasIssue(List<String> filesToScan, JavaFileScanner chec
...
+
+    @SuppressWarnings("UnstableApiUsage")
+    private static JavaAstScanner createAstScanner(
+            SonarComponents sonarComponents, List<JavaFileScanner> checks) {
+        JavaAstScanner scanner = new JavaAstScanner(sonarComponents);
+        VisitorsBridge visitorsBridge =
+                new VisitorsBridge(
+                        checks,
+                        // TODO set the classpath to something reasonable
+                        Collections.emptyList(),
```
slarse, line 112: This argument sets the classpath for the VisitorsBridge. It's later supplied to the parser, so I'm assuming that it's used to resolve types etc (but I haven't checked inside the parser). We will probably want to set it to something reasonable in the context. For example, when analyzing a project that uses maven, the full classpath can be generated with `mvn dependency:build-classpath`.

src/main/java/sorald/sonar/RuleVerifier.java
```diff
@@ -54,6 +83,78 @@ public static void verifyHasIssue(List<String> filesToScan, JavaFileScanner chec
...
+        //      Java 14)
+        //      visitorsBridge.setJavaVersion(new JavaVersionImpl(8));
+        scanner.setVisitorBridge(visitorsBridge);
+        return scanner;
+    }
+
+    private static SoraldSonarComponents createSonarComponents(File baseDir) {
+        // FIXME The SensorContextTester is an internal and unstable component in sonar,
+        //       we should implement our own SensorContext
+        SensorContextTester context = SensorContextTester.create(baseDir);
```
slarse, line 126: Like the comment in the code says, we shouldn't be using the internal `SensorContextTester`, but it will serve for the time being.

src/main/java/sorald/sonar/RuleVerifier.java
```diff
@@ -54,6 +83,78 @@ public static void verifyHasIssue(List<String> filesToScan, JavaFileScanner chec
...
+     * way".
+     */
+    private static class SoraldSonarComponents extends SonarComponents {
+        private final List<AnalyzerMessage> messages;
+
+        public SoraldSonarComponents(DefaultFileSystem fs) {
+            // FIXME I'm very unsure about supplying null for all of these constructor values.
+            //       They should not be null, but at this time I don't know that to put there, and
+            //       with our current usage this hack appears to work.
+            super(null, fs, null, null, null, null);
```
slarse, line 146: Like the comment says, this is a pretty nasty hack. It appears to work, but it's definitely not pretty nor safe as it initializes fields that should not be `null` to `null`.

slarse (commented):
@fermadeiral I'm not sure I'd call this ready for merge, but I think it's time you look it over.

Essentially, I've managed to upgrade to sonar-java 6.9.0. I ran into quite a few issues with inconsistent use of relative/absolute and non-normalized/normalized paths, which prompted me to open #166 . 

There's also one point of breakage: The previously noncompliant code in `CodeFactory.java` is now not detected as noncompliant. I've ignored it for now, I have not been able to solve it, nor figure out why it happens.

I've also found our use case is a little bit special. sonar is built on a client/server model, and our application is both the server (that generates the analyzer messages) and the client (that consumes them). It doesn't quite mesh with the design of the libraries. These parts are also largely undocumented, and so this is something of trial and error on my part. I've put in a few hacks here (see review comments) to make it work for now, but I expect to be able to remove them as my understanding of how sonar is built improves.

To be clear, though, the way sorald used sonar before was _also_ a hack, so this is not a step backwards. Perhaps a bit sideways :)

Let me know what you think of all of this, and if you have any ideas for the particular hacks that I've employed. As you've been working on this longer, you may know things I don't.

fermadeiral:
Conflict.

slarse:
Conflicts fixed, and also applied a trivial optimization in 163e4c2 that makes execution of checks that don't require symbolic execution several times faster.

Let's postpone a review of this until next week, I'll probably need the rest of this week to conjure up the rule prioritization scripts anyway :)

fermadeiral:
> @fermadeiral I'm not sure I'd call this ready for merge, but I think it's time you look it over.

I've done it.

> There's also one point of breakage: The previously noncompliant code in CodeFactory.java is now not detected as noncompliant. I've ignored it for now, I have not been able to solve it, nor figure out why it happens.

In the current stage of the things, this is minor in my opinion, I'm happy you just ignored it.

> To be clear, though, the way sorald used sonar before was also a hack, so this is not a step backwards. Perhaps a bit sideways :)

Maybe we will always need to hack, because the SonarSource team didn't design its products to be used like we want. I see this PR as a step forwards because now we use a newer version of sonar java at least.

For now, I think this PR is pretty good. Ready to merge?

slarse:
@fermadeiral Yes, ready to merge!

fermadeiral:
@slarse thank you for this great PR!

pom.xml
```diff
@@ -49,22 +49,29 @@
...
             <artifactId>spoon-core</artifactId>
-            <version>8.2.0-beta-12</version>
+            <version>8.3.0</version>
+            <exclusions>
+                <exclusion>
+                    <!-- must be excluded as it is signed, which causes problems with unsigned version from sonar -->
+                    <groupId>org.eclipse.jdt</groupId>
+                    <artifactId>org.eclipse.jdt.core</artifactId>
+                </exclusion>
+            </exclusions>
```
algomaster99, line 59: @slarse Do you remember what problems were arising in the absence of its exclusion? I am using the local build of spoon with this [change](https://github.com/INRIA/spoon/pull/4320#issuecomment-988121027) in the comment, but sorald throws an error that it cannot find the `.get` method for `HashtableOfPackage`. I am assuming it is because of this line.

slarse: As I recall, Sonar-java 6.9 pulled in (or bundled, maybe) an unsigned `jdt` package, while the one pulled in from Spoon is signed. Signed as in there's a GPG signature on it.

That caused errors on building it.

As both Spoon and Sonar-Java use jdt, you need to make sure that the version they depend on is the same. Otherwise you'll get strange errors.

PR #160, opened 2020-10-20
chore: Remove the maven.inria.fr-snapshot repository

slarse:
This PR removes the `maven.inria.fr-snapshot` snapshot repository. I'm guessing it's there because sorald used Spoon snapshots way back when.

As Spoon now uses OW2 Nexus as a snapshot repo (inria/spoon#3650), we should either update the snapshot repository to OW2 Nexus, or simply remove it until such a time where we need a Spoon snapshot. I suggest simply removing it with this PR.


fermadeiral:
> I'm guessing it's there because sorald used Spoon snapshots way back when.

You are right.

Thanks for the cleaning!

PR #161, opened 2020-10-20
chore: Adopt Google's formatting style

### LINKED ISSUE #157: Adopt a house style?
We currently don't have any style enforcement, and it shows in some places. For example [CompareToReturnValueProcessor](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/processor/CompareToReturnValueProcessor.java) (along with most other processors) uses tabs for indentation, while the entire [`sorald.segment`](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/segment/) subpackage uses spaces.

I don't think we need anything rigorous, but at the very least we should decide upon and enforce a style of indentation.

slarse:
Fix #157 

@fermadeiral I know we said no review on this one, but this is a tiny bit different than what we discussed, so I want to run it by you before I merge. Instead of using Checkstyle, I opted for [Spotless](https://github.com/diffplug/spotless). It's a neat tool that can plug into Maven to do the formatting for you, and can also run checks. `mvn spotless:apply` to apply formatting fixes, and `mvn spotless:check` to check for violations.

Now, what I found pragmatic about Spotless is that it will only report a check violation if it can be automatically fixed. So, for example, Google's formatting style is very picky about having a Javadoc comments on every public method and class, that lines are less than 100 characters, etc. However, the autoformatter often _can't fix those things_, and so fixing them is a manual affair. Spotless will not complain about such violations, because `google-java-format` can't fix them.

This seems like a nice tradeoff. We get a consistent code style with absolutely no manual effort, and it doesn't hinder contributions by others as all violations can be fixed by a simple `mvn spotless:apply`. What do you think about this?

And one more thing: 2 spaces or 4 spaces for indentation, any preference? The tool allows us to choose.

Ps. Only `.travis.yml` and `pom.xml` contain changes other than formatting. Note that I've updated Travis to use OpenJDK11 as `google-java-format` requires that as of 1.8 (and they are now on 1.9). This requires developers of sorald to use JDK11+, but it can still be built with JDK8+ as Spotless is not hooked into any phase and won't run unless explicitly asked to. Alternatively, we can downgrade to `google-java-format` 1.7, but I really do think it's time for the world to move on from Java 1.8 :)


fermadeiral:
> This seems like a nice tradeoff. We get a consistent code style with absolutely no manual effort, and it doesn't hinder contributions by others as all violations can be fixed by a simple mvn spotless:apply. What do you think about this?

Seems good to me!

> And one more thing: 2 spaces or 4 spaces for indentation, any preference? The tool allows us to choose.

I prefer 4 spaces.

slarse:
I also prefer 4 spaces, so let's go with that. Will fix that and merge tomorrow morning!

fermadeiral:
@slarse ok!

src/main/java/sorald/DefaultRepair.java
```diff
@@ -14,101 +14,107 @@
...
+            } else { // more than one processor, thus we need an intermediate dir, which will always
+                // contain all files (the changed and non-changed ones), because other processors
+                // will run on them
+                if (i == 0) { // the first processor will run, thus we use the given input file
+                    // dir and output
+                    // *all* files in the intermediate dir
+                    inputDirPath = this.config.getOriginalFilesPath();
+                    outputDirPath = intermediateSpoonedPath;
+                } else if ((i + 1)
+                        == ruleKeys.size()) { // the last processor will run, thus we use as input
```
slarse, line 51: The formatter does this to limit line lengths when there's an inline comment that makes the line too long. It looks pretty funky. We should probably try to avoid inline comments like this one :)

slarse:
@fermadeiral I'm merging this as I need it right now to be able to keep working. Feel free to retroactively change anything. Here's a summary of the non-formatting changes.

* Spotless added to `pom.xml`.
    - Not bound to any phase, must be run manually with `mvn spotless:check` to check compliance or `mvn spotless:apply` to apply the formatter.
    - AOSP style used, i.e. 4 spaces
* `CONTRIBUTING.md` now contains a step about ensuring code compliance with `mvn spotless:apply`
* `.travis.yml` updated to run `spotless:check`, and use JDK11 (required by newer versions of `google-java-formatter`)

fermadeiral:
Thanks, @slarse.

PR #162, opened 2020-10-21
refactor: Encapsulate sonar-java verification facilities in adapter

slarse:
As mentioned in #156, we're using unstable parts of the sonar-java internals that need to be isolated for maintainability. This PR adds the `sorald.sonar.Verifier`, which encapsulates the calls to the testing internals. It's pretty likely that we'll want to slightly change the interface of this class later down the line, but that's not too much of a bother.

The idea is that we'll put all the sonar adapters/facades into the `sorald.sonar` subpackage, to keep it all nice and tidy.

PR #163, opened 2020-10-21
refactor: Add facade around AnalyzerMessage

slarse:
This is a follow-up on #162, and depends on it. ~**Don't review this until #162 i smerged,** I will fix the commit history.~

This PR extracts the previously internal `SoraldAbstractProcessor.Bug` class into a separate `sorald.sonar.RuleViolation` class, which is just a thin facade around `org.java.sonar.AnalyzerMessage`. Depending on how we proceed with #156, we may not have access to the actual `AnalyzerMessage` (but some higher-level abstraction instead), and so this PR is a precaution.

Since the facade was essentially already there in form of the Bug class, this change was very little effort, and also coincidentally cleans up some of the code base. With this change, the only sonar internals we touch in sorald are the check classes, but it doesn't seem reasonable at the moment to put adapters around them. I think we'll leave them be.

### DIFF
--- a/src/main/java/sorald/miner/MineSonarWarnings.java
+++ b/src/main/java/sorald/miner/MineSonarWarnings.java
@@ -19,8 +19,8 @@ import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.sonar.plugins.java.api.JavaFileScanner;
 import sorald.Constants;
-import sorald.sonar.Bug;
 import sorald.sonar.RuleVerifier;
+import sorald.sonar.RuleViolation;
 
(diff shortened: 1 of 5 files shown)


slarse:
@fermadeiral Ready for review! Code coverage drop is due to removal of code.

fermadeiral:
@slarse, thank you for this refactoring.

I noticed you have added documentation in the source code since previous PRs. That is great!

slarse:
@fermadeiral Glad you like it! I tend to add documentation to most of my code. I enjoy being tidy :)

Sometimes I add too much documentation as I like writing, so you may want to be on the lookout for overly verbose docs where something less verbose would do just fine.

Issue #153, opened 2020-10-16
Reusing the same workspace for each test causes dependencies between tests


slarse:
The `Constants.SORALD_WORKSPACE` path is used for every test, but it's not cleaned up. This means that if a source file is used for two tests, it's output path may very well be the same between the two tests. Therefore, the second test may pass without doing anything if the first test produced the correct output.

I see two reasonable solutions.

1. Use a temporary directory for the workspace
2. Clean up the workspace after or before each test

I prefer solution 1, but because of the current setup it may be less work to implement solution 2.

fermadeiral:
Thank you for pointing that out.

We already have a solution but not exactly for the same problem, and I think this solution is only applied in one test class: see `FileOutputStrategyTest`, the `tearDown`method.

> I prefer solution 1, but because of the current setup it may be less work to implement solution 2.

I don't have strong feelings about any of the proposed solution, so you can go with the one you prefer.

slarse:
@fermadeiral Right, thanks. That approach can be generalized to a JUnit test listener that applies to the entire test suite, which is essentially solution 2. I'm not entirely sure how to do it for our test suite as it's currently a mix of JUnit4 and 5, and their respective test listeners are slightly different. But it should be manageable.

The reason I prefer solution 1. is that it JUnit5 has great support for [automatic handling of temporary directories](https://junit.org/junit5/docs/5.4.0/api/org/junit/jupiter/api/io/TempDir.html). It also becomes explicit that the directory used is indeed intended to be used temporarily. Again, not sure how to apply that across JUnit4+5, it may be preferable to convert the whole test suite to JUnit5 before proceeding here.

PR #164, opened 2020-10-21
test: Convert all JUnit4 tests to JUnit5

### LINKED ISSUE #155: Fully adopt JUnit5
To simplify further maintenance of the test suite, it'd be preferable to fully adopt JUnit5. This will in the short term make #153 much easier to solve, and in the long term lead to less headache.

It's really not difficult, the current test suite uses very little JUnit4-specific functionality.

This must be done after #144 is merged, as that touches so many parts of the test suite.

slarse:
Fix #155 

This PR fully adopts JUnit5 by converting all remaining JUnit4 tests. We no longer need the vintage engine, so that's dropped from the pom. This will make more elaborate test fixtures (e.g. the stuff discussed in #153) easier to manage.

The entire conversion was automatic, with the exception of editing the pom and a single test case that had to be slightly rewritten (see review comment).

### DIFF
--- a/src/test/java/sorald/SegmentStrategyTest.java
+++ b/src/test/java/sorald/SegmentStrategyTest.java
@@ -1,6 +1,8 @@
 package sorald.processor;
 
-import org.junit.Test;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
 import org.sonar.java.checks.ArrayHashCodeAndToStringCheck;
 import sorald.Constants;
 import sorald.Main;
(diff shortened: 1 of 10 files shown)


src/test/java/sorald/SegmentStrategyTest.java
```diff
@@ -38,15 +40,13 @@ public void arrayToStringProcessor_success_Test() throws Exception {
         RuleVerifier.verifyNoIssue(pathToRepairedFile, new ArrayHashCodeAndToStringCheck());
     }
 
-    @Test(expected = RuntimeException.class)
+    @Test
```
slarse, line 43: This test is the only one that I had to convert manually.

slarse:
@fermadeiral Ready for review.

By the way, do you get a notification when I assign you as a review? If so I can stop pinging in a comment.

fermadeiral:
> @fermadeiral Ready for review.

Thanks for this PR! I'm merging.

> By the way, do you get a notification when I assign you as a review? If so I can stop pinging in a comment.

Yes, I get notification on everything: when you push, when you assign me, when you ping me, etc. It's a bit annoying in fact, but once you come to the office we can discuss on how to communicate nicely.

slarse:
A simple solution for the time being would be for you to reduce your watch status on the repository from "Watching" to "Releases only" or lower. I'm watching the repo, and so can ping you if there's anything you actually need to look at.

PR #165, opened 2020-10-21
test: Automatically clean default workspace before each test

### LINKED ISSUE #153: Reusing the same workspace for each test causes dependencies between tests

slarse:
Fix #153 

This PR introduces a test listener that automatically cleans up the default workspace _before_ each test. It is therefore now safe to use the default workspace for all tests, and all tests that previously interfered with each other now don't.

The listener in question is located in `src/test/sorald/WorkspaceCleaner.java`, and is registered as a test service in `src/test/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,12 @@
             <version>5.7.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.platform</groupId>
+            <artifactId>junit-platform-launcher</artifactId>
+            <version>1.7.0</version>
+            <scope>test</scope>
+        </dependency>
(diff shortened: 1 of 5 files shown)


fermadeiral:
Thanks @slarse!

Issue #142, opened 2020-10-13
Processor tests are weak


slarse:
Another issue related to #141 is that I've found the test cases for processors to be fairly weak. They only assert that the repaired file contains no issues according to sonar-java. That's not a strong assertion, as I found when I very incorrectly re-implemented the  [CompareStringsBoxedTypesWithEqualsProcessor.java](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/processor/CompareStringsBoxedTypesWithEqualsProcessor.java). In fact, one can go as far as replace the processor implementation with this:

```java
	@Override
	public void process(CtElement element) {
		super.process(element);
		element.delete();
	}
```
it passes the test suite just fine. I found you can do something clearly incorrect like that with several other of the processors as well. I'm a little bit hesitant to get to work with refactorings as my first attempt at one was massively incorrect, yet the test suite said nothing about it.

I'm note sure about the best way to proceed here, but one solution would be to manually add an expected output file to each test case. This should typically not be too hard, given that sorald is template-based. I've found that many of the SonarCube rules contain a non-compliant example, and a directly related compliant example, which would give the oracle for free.

A straight equality check with Spoon between the expected output and the actual output may be enough, but it could be too strict (e.g. formatting differences could come into play). I worked extensively with equality checks during my thesis, and so have quite a few backup plans if the equality check with Spoon is too strict. For example, gumtree-spoon could be used, and a zero-diff would indicate equality.

Thoughts on this?

slarse:
Now that we have the capability for stronger tests, it's just a matter of adding more expected outcomes.

PR #167, opened 2020-10-26
test: Add stronger processor tests

slarse:
#142 

This PR adds the capacity for adding _expected output files_ for any given processor test file. See for example 656def8, which adds an expected output file for `BigDecimalDoubleCounstructor.java`. There's a new parameterized test case in `ProcessorTest.java` that picks up test files that have a corresponding expected file, runs sorald on the test file, and then compares the processed output to the expected output file.

These tests are _much_ stronger than the existing processor tests, which just verify that there are no issues in the processed files. With the existing tests, I can implement most processors as simply deleting their input from the AST, and pass the tests. That's not possible with these tests, as then the expected output does not match the actual output.

This PR is a bit big, as it required some refactoring in `ProcessorTest` to not have too much code duplication. I can easily split it into two PR's: one for the refactoring, and one for the addition of this new test. Let me know if you want me to do that.


fermadeiral:
Conflict.

fermadeiral:
@slarse, thank you!

PR #169, opened 2020-10-27
feat: Make Sonar checks accessible by type and key

### LINKED ISSUE #158: Separate sonar checks in Constants.java according to their types.
The sonar checks of different types are mixed in Constants.java, i.e., the checks related to bugs are mixed with the ones related to code smells (DeadStoreCheck and SerializableFieldInSerializableClassCheck). Thinking about the miner, it's better for us to organize the checks in a separate way so that data collection will be organized and easier to be analyzed by us.

slarse:
Fix #158 

The new `Checks.java` class allows for easy access to Sonar checks, either by type or by key.


slarse:
Coverage drop is due to each check class previously being added in a statement each in `Constants.java`, but now they're all inlined in method calls in `Checks.java`. I'm not sure how `Constants.java` has 75% coverage, though, as it only has static code and no branches. Nothing I can do about that.

fermadeiral:
@slarse, thank you!

Issue #168, opened 2020-10-27
Optimize warnings miner by running all checks in one pass


slarse:
Currently, the warnings miner loops through the checks and applies them one-by-one, which essentially means re-parsing the file and traversing the AST for each check. Running time can be drastically reduced by applying all checks in one pass, which will be easy to do once #156 is merged.

Issue #51, opened 2020-03-19
security / vulnerability warnings to implement in Sorald


monperrus:
at some point, we'll need to prioritize the warnings to repair. on this topic, see

"Developer-Driven Code Smell Prioritization" https://fabiano-pecorelli.github.io/publications/conferences/C7.pdf

slarse:
@monperrus We're about to get to the point where it's appropriate to start implementing new processors, and I've been looking specifically at vulnerabilities. 

Before I dive deeper into that, I just want your opinion on an issue. SonarQube appears to "cross-classify" many minor vulnerabilities as code smells. For example, if I look for vulnerabilities in a project, [this rule often pops up](https://sonarcloud.io/project/issues?id=apache_struts&open=AW_X5CrW1FYS5oU8LTdB&resolved=false&rules=java%3AS1444&types=VULNERABILITY). Note how it's classified as a vulnerability in the list to the left.

However, when I later go to look up the details of the rule, [it's classified as a code smell](https://rules.sonarsource.com/java/RSPEC-1444). Should we count such rules as vulnerabilities?

slarse:
I've composed a list of "pure" vulnerability rules that I think are good candidates for sorald to fix. There are many encryption-scheme rules that I think would be relatively simple to fix with templating. The list is in no particular order, and I haven't made any attempt yet to find out how common any of these are in practice.

* [5542: Encryption algorithms should be used with secure mode and padding scheme](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-5542)
* [5547: Cipher algorithms should be robust](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-5547)
    - I'm not quite sure what differentiates this one from 5542
* [4432: AES encryption algorithm should be used with secured mode](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-4432)
    - Again, not quite sure what differentiates this one from 5547 and 5542.
* [2277: Cryptographic RSA algorithms should always incorporate OAEP (Optimal Asymmetric Encryption Padding)](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2277)
* [2070: SHA-1 and Message-Digest hash algorithms should not be used in secure contexts](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2070)
* [4434: LDAP deserialization should be disabled](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-4434)
[...]

PR #174, opened 2020-10-29
experiment: Add script for extracting sonar warnings per commit

slarse:
Related to #51 

This is heavily a WIP, and I'm not sure if we want this script in this repository, or if I should just create another one. In any case, the script processes a list of repo urls, clones them to disc, and extracts sonar warnings for each commit using the warnings miner. Results are written to CSV files (one for each repo url).

The current implementation of the warnings miner is too slow, processing a single commit in Spoon takes several minutes, and so #168 must be implemented first. With the optimizations in place, processing a commit in Spoon (some ~400 source files in spoon-core) takes around 11-12s, using 4 CPU cores. There's still 3000 commits in Spoon, so on my personal machine just processing Spoon's history will take around 10 hours.

If we really want to scale this up, we could use MPI. But first I need to figure out the best way to use the data. The amount of hits for a given rule can both increase and decrease as code is added and removed, but this is not necessarily related to actually fixing the errors.


monperrus:
Cool! We want this in the repository. On parallelization, MPI is old and maybe suboptimal, there are several newer alternatives to consider (Kubernetes, etc).

slarse:
@monperrus MPI might be old, but I know how to use it, and I'm not aware of a 1:1 alternative. Kubernetes is for container orchestration, so I don't see how it's a replacement for MPI? In fact, I just looked around, and there are solutions to use MPI on a Kubernetes cluster.

But that was just a wild thought, I don't think it will necessary for us to scale beyond a single node. There's already single-node parallelization in the script.

To reduce workload we could process every nth (e.g. 10th) commit or something like that, which should still give a pretty good idea of warning-change over time.

monperrus:
Plus we have powerful servers in clusters should you need one.

slarse:
The script now computes both the absolute values, and the deltas between commits. Question is now: what to do precisely with this data?

Here are two sample files based on the Tic-Tac-Toe repo already used in some tests in Sorald:

* [Absolute results per commit](https://github.com/SpoonLabs/sorald/blob/c56969d43daab58788e53f3556d11d9b2c0e286f/scripts/commit_warnings_extractor/https___github.com_mayurkadampro_Tic-Tac-Toe.csv)
* [Deltas](https://github.com/SpoonLabs/sorald/blob/c56969d43daab58788e53f3556d11d9b2c0e286f/scripts/commit_warnings_extractor/https___github.com_mayurkadampro_Tic-Tac-Toe.deltas.csv)
    - For example, `240.0` in a column means that 240 of that warning were introduced with that commit, and `-240.0` means that so many warnings were removed.
    - So, positive delta is bad, negative delta is good!
    - The reason for the massive deltas in this example is that the entire code base was added and removed a few times

slarse:
@monperrus I've now got something relatively usable for extracting per-commit warnings in arbitrary Maven repositories. It supports multi-module projects as well. Processing _all_ commits from any given projects seems unnecessarily granular, and so I've added a limit and a "step size", such that one can sample a fixed amount of the latest commits with a fixed step size between each sample point. Using the warnings miner from #188 to run only the security checks we're currently interested in, running experiments is relatively quick.

I don't really know what to do with the data yet, though, so I'll leave this as-is until we can discuss it during the meeting on Wednesday.

monperrus:
Perfect. You can push the data to https://github.com/SpoonLabs/sorald-experiments/

(per our practice of our [experimental data repos](https://github.com/SpoonLabs/?q=experim))

we discuss on Wednesday!

monperrus:
@khaes-kth will look at this

slarse:
@khaes-kth I will put some more useful data up here tomorrow. Would you like me to slap a CLI onto the script so you can run it yourself on projects of your choice?

FYI, the reason this is a Python script on top of the miner is that it's so much faster to build this out in Python than in Java. I'm also not convinced we'll even want to have this in the core tool, it seems like too specific a use case.

khaes-kth:
Hi @slarse 



> I'm also not convinced we'll even want to have this in the core tool, it seems like too specific a use case.

I think it's quite useful to have it in the repository.  So, if you're OK with that, let's keep it.

Is the current version finalized? Did you decide about the parallelization?



> Would you like me to slap a CLI onto the script so you can run it yourself on projects of your choice?

That would be great! (+ a sample execution command)

slarse:
@khaes-kth 

> I think it's quite useful to have it in the repository. So, if you're OK with that, let's keep it.

I think you misunderstood what I meant. I meant I'm not sure we want this functionality in the actual Sorald Java application, as it's a highly specific use case. The script we can of course keep, it does not add any particular maintenance overhead for us.

> Is the current version finalized? Did you decide about the parallelization?

Finalized, not really. It's thrown together at best :). The parallelization is simply node-local, it will scale with the amount of CPU cores on the machine it runs on. If we want to run on a cluster, I think the easiest (and sufficient) thing is to just do one job per node. 

> That would be great! (+ a sample execution command)

Alright, cool, I'll put a CLI on it right away and get back to you when it's there!

khaes-kth:
> I think you misunderstood what I meant. I meant I'm not sure we want this functionality in the actual Sorald Java application, as it's a highly specific use case. The script we can of course keep, it does not add any particular maintenance overhead for us.

Aha! You're right I did not understand it at first. I agree, I think we can keep it as it is. Did anyone in the meeting have other thoughts about it?



> The parallelization is simply node-local

Good! I just noticed you are already using multiprocessing.



> I'll put a CLI on it right away and get back to you when it's there!

Great! Looking forward to it.

slarse:
@khaes-kth CLI is up! See the README file in the same directory as the script for usage examples.

slarse:
Oh, one thing I forgot: the `--ruleTypes` argument to sorald is currently hard-coded in the script to `vulnerability`, see line 205. Forgot to add it to the CLI, but you can just change that in the script for now.

khaes-kth:
Hi Simon,

Thanks a lot for the preparing the CLI.

I tested it on two repositories and it worked well.

Two points: first, the function "multiprocessing.cpu_count()" returns 0 on a virtual box VM. I think it is better to have "num_cpus = max(multiprocessing.cpu_count(), 1)" at line 159. Second, on windows I faced this error and could not fix it:

`PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\kheso\\AppData\\Local\\Temp\\tmppegj4a65\\pkgextractor'`

Do you know how I can fix it?

slarse:
@khaes-kth

> "multiprocessing.cpu_count()" returns 0 on a virtual box VM

I see, that I had no idea about. Will fix!

> PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\kheso\\AppData\\Local\\Temp\\tmppegj4a65\\pkgextractor'

I'm not sure what's causing that, but I'm guessing it's this part where I add/remove worktrees to the repositories: https://github.com/slarse/sorald/blob/a16cb4a127abc53433a746972be63b51b1700578/scripts/commit_warnings_extractor/main.py#L254-L266

That happens in parallel without synchronization, which afaik is safe as it's just a matter of reading from the local repository and adding a directory in `.git/worktrees`. But perhaps Windows doesn't like something about simultaneous reads from the same file? I don't know too much about Windows, really. Want me to try and fix it, or is it sufficient for you to run in a VM?

slarse:
@khaes-kth Fixed the `num_cpu` thing, you can now specify it on the CLI whit `--num-cpu`. Will default to the expression you suggested.

khaes-kth:
@slarse 

> @khaes-kth Fixed the `num_cpu` thing, you can now specify it on the CLI whit `--num-cpu`. Will default to the expression you suggested.

Great! Thanks!



> Want me to try and fix it, or is it sufficient for you to run in a VM?

No, I tested it on a VM and it was fine. We should just know it might face some issues on Windows.

Please let me know when it is ready to be merged.

slarse:
@khaes-kth 

> Please let me know when it is ready to be merged.

I just need to add the ability to pass options to the miner (e.g. `--ruleTypes`), will do that ASAP in the morning.

> We should just know it might face some issues on Windows.

Okay! To be honest, I wasn't even considering Windows, as I don't use it.

Now, there's one last thing: how robust should I make this? We could merge it pretty much like this, but it's essentially written as a one-off script without testing or anything like that. I could make it more robust and add tests for it, but I don't know how useful it will be to us yet as I don't know what to do with the data it extracts. I see two options:

1. Merge without automated tests and add tests down the road if we actually find use for the scripts more than once.
2. Add tests before merging.

What do you think?

slarse:
@khaes-kth I'm done in terms of functionality, so now it's just a matter of if we want it tested as well, or if we're ok with it being untested.

khaes-kth:
I do not think it is required to be tested. So, I'll merge it.

PR #175, opened 2020-10-29
refactor: Run all checks in one pass in warnings miner

### LINKED ISSUE #168: Optimize warnings miner by running all checks in one pass

slarse:
Fix #168 

This PR will conflict with #156 , that one must be merged first.

Optimizes the warnings miner by running all checks in one pass, instead of running one check at a time. Makes the miner many times faster than it was previously.

### DIFF
--- a/src/main/java/sorald/miner/MineSonarWarnings.java
+++ b/src/main/java/sorald/miner/MineSonarWarnings.java
@@ -202,11 +202,14 @@ public class MineSonarWarnings {
                     e.printStackTrace();
                 }
             }
-            for (JavaFileScanner javaFileScanner : SONAR_CHECK_INSTANCES) {
-                Set<RuleViolation> ruleViolations =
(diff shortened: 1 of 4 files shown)


slarse:
This is now ready for review. Essentially does three things:

1. Adds a new method `RuleVerifier.analyze` that accepts multiple checks, and then applies all of those checks in one pass.
2. Adapts the warnings miner to use the now method.
3. Adds a new test to check that the warnings miner outputs results even for rules that are not violated (i.e. zero-results)

I realized the new test was necessary as when I first implemented the optimization in the warnings miner it would only output results for violated rules. That caused issues in my warnings-extractor script in #174 .

fermadeiral:
Thanks, @slarse!

PR #176, opened 2020-10-30
refactor: Refactor CompareStringsBoxedTypeWithEqualsProcessor

slarse:
I was feeling very unproductive, so I decided to do something easy. This PR is a pure refactor, and also adds the expected output of the processor as computed _before_ the refactor. Having the expected output is hugely beneficial when refactoring; I completely broke the processor on my first attempt, which was caught by the computed output not matching the expected output.


fermadeiral:
Even though it is simple, I love this kind of refactoring! Thank you!

PR #180, opened 2020-10-30
doc: Update contributing guidelines

### LINKED ISSUE #178: Contributing guidelines for adding a processor need some updating
When implementing my first processor, I noticed some things in the docs that can be improved/fixed.

- [x] There is a reference to the old `SQRAbstractProcessor`, which is now called `SoraldAbstractProcessor`
- [x] There is no mention of the possibility of adding a ".expected" file when creating a test input file. This should probably be a requirement for new processors.
- [x] There is no mention that the check class the processor is implemented for must be added to its corresponding category in `Checks.java`

slarse:
Fix #178 

Feel free to disagree on requiring an expected output file for each test case, I just thought that as the PR should include the diff, they must essentially have the expected out put anyway.


fermadeiral:
> I just thought that as the PR should include the diff, they must essentially have the expected out put anyway.

Totally agree!

PR #185, opened 2020-11-02
fix: Fix handling of crashing checks

### LINKED ISSUE #183: No handling of failure to analyze a file
If Sonar-Java fails to analyze a file, the current implementation just keeps going like nothing happened. This is a problem especially in testing, as the tests currently don't check that there are no errors.

### LINKED ISSUE #184: WeakSSLContextCheck crashes due to unspecified version number
The `WeakSSLContextCheck` crashes as the java version isn't set on the visitors bridge. As such, we must set it to _something_. Ideally, we should set it to something that is appropriate for the project/file being analyzed, but as a quick-fix we can simply set it to 8.

slarse:
Fix #183 
Fix #184 

This PR does two things.

1. Expose check crashes by not catching all exceptions in the warnings miner
    - If there is a crash, we want to know about it, especially when the crash is our fault
2. Fix such a crash in WeakSSLContextCheck by setting the Java version number in the VisitorsBridge to 14

Setting the Java version to 14 is an imperfect fix. It could theoretically cause false positives if a project uses a lesser version of Java by suggesting fixes that aren't applicable to it. Most of these seem related to `< java 8`, though. But setting it too low might cause parsing issues as the parser expects a certain compliance level. As such, I think setting the version a bit too high is preferable to a bit too low at the moment.

### DIFF
--- a/src/main/java/sorald/miner/MineSonarWarnings.java
+++ b/src/main/java/sorald/miner/MineSonarWarnings.java
@@ -185,37 +185,42 @@ public class MineSonarWarnings {
         return res;
     }
 
-    private static Map<String, Integer> extractWarnings(String projectPath) {
+    /**
+     * @param projectPath The root path to a Java project
+     * @param checks Checks to run on the Java files in the project
+     * @return A mapping (checkClassName -> numViolations)
+     */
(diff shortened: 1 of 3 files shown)


fermadeiral:
Nice!

PR #186, opened 2020-11-02
refactor: Replace SoraldAbstractProcessor.getSonarCheck with Checks.getCheckInstance

### LINKED ISSUE #177: SoraldAbstractProcessor.getSonarCheck can be removed
With the new `Checks` class, which has a static method for getting checks by key, there is no longer a reason for a processor to provide its corresponding check with `getSonarCheck`.

https://github.com/SpoonLabs/sorald/blob/196103354028d9495e39d20117880e5518347fc1/src/main/java/sorald/sonar/Checks.java#L58-L71

https://github.com/SpoonLabs/sorald/blob/196103354028d9495e39d20117880e5518347fc1/src/main/java/sorald/processor/SoraldAbstractProcessor.java#L31

Instead, the key of the check can be fetched from the `ProcessorAnnotation`, and used to find the correct check. This reduces boilerplate, and also ensures that new checks are actually added to the map in the `Checks` class when a corresponding processor is added.

slarse:
Fix #177 

This replaces the abstract method `SoraldAbstractProcessor.getSonarCheck` with the `Checks.getCheckInstance` method. As we have all checks in key-to-check mappings, and each processor has an annotation with the key for the rule, this should be a fully sufficient replacement.

The benefit is that new processors don't have to implement the `getSonarCheck` method and, in fact, don't need to know about sonar at all anymore as the abstract base class handles that.

### DIFF
--- a/src/main/java/sorald/processor/ArrayHashCodeAndToStringProcessor.java
+++ b/src/main/java/sorald/processor/ArrayHashCodeAndToStringProcessor.java
@@ -1,8 +1,6 @@
 package sorald.processor;
 
 import java.util.Arrays;
-import org.sonar.java.checks.ArrayHashCodeAndToStringCheck;
-import org.sonar.plugins.java.api.JavaFileScanner;
 import sorald.Constants;
 import sorald.ProcessorAnnotation;
 import spoon.reflect.code.CtExpression;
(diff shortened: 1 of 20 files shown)


slarse:
Coverage drop due to large amount of removed files, these two new lines that are kind of pointless to test: https://codecov.io/gh/SpoonLabs/sorald/compare/4a77be94a7ba961f22eabc2328d970486f068bc2...ac9aac34418c69c1886162f054c661d16e866060/src/src/main/java/sorald/processor/SoraldAbstractProcessor.java#L137

fermadeiral:
Love this refactoring!

PR #187, opened 2020-11-02
refactor: Move delete directory method to file utils

slarse:
Pure refactor. Saw the comment in the source code, and as it was a 2-min fix I did it.

### DIFF
--- a/src/main/java/sorald/SoraldAbstractRepair.java
+++ b/src/main/java/sorald/SoraldAbstractRepair.java
@@ -21,19 +21,6 @@ public abstract class SoraldAbstractRepair {
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/sorald/SoraldAbstractRepair.java
```diff
@@ -21,19 +21,6 @@ public SoraldAbstractRepair(SoraldConfig config) {
         }
     }
 
-    // FIXME: this method was copied from TestHelper.java. We should extract it to a FileHelper to
-    // be
-    // visible for both main code and test code.
```
slarse, line 26: This is the comment I referred to in the PR body

fermadeiral:
Cleaning...this feels great!

Issue #148, opened 2020-10-15
Add documentation for the Sonar warnings miner.


fermadeiral:

monperrus:
that will be done after the CLI refactoring of #192

slarse:
See also discussion in #188

PR #188, opened 2020-11-02
feat: Add --ruleTypes option to warnings miner

### LINKED ISSUE #182: Make it possible to run mining with only rules from a specific category
For the experiments, I need to be able to mine warnings from a specific category (vulnerabilities). So it should be possible to specify something like "--ruleCategory vulnerability" on the command line.

~It seems most reasonable to me that this should be additive to the individual rules, e.g. that I can specify "--ruleCategory vulnerability --ruleKeys 1234". However, it should also be possible to specify only the rule category, or only the rule keys.~

Oh, wait, the warnings miner is actually a different CLI entirely from the repair application. It seems to me like these should be joined together, with the miner and the repairer as subcommands, but that's for a different time.

slarse:
Fix #182 

~Not ready for review and depends on #185~

This PR makes it possible to specify one or more rule types on the command line with the `--ruleTypes` option, followed by a `,`-delimited list of rule types (one of "vulnerability", "code_smell", "bug" or "security_hotspot"). Omitting `--ruleTypes` causes the miner to behave as per usual, using all checks. Specifying e.g. `--ruleTypes bug,vulnerability` causes the miner to run only the checks classified as `bug` and `vulnerability` in `Checks.java`.

The reason this is useful is that the more checks one scans with, the longer it takes. So, for my experiments where I specifically want to know about vulnerabilities, I can cut the running time by more than 50% simply by only running with vulnerability checks.

There's essentially no error handling in the parsing of rule types from the command line. If one specifies for example `--ruleTypes bla`, then there will be a crash like so:

```
Exception in thread "main" java.lang.IllegalArgumentException: No enum constant sorald.sonar.Checks.CheckType.BLA
```

I can add error handling, but there's essentially no error handling on any of the other inputs, which is why I haven't added anything. I think that we should probably redefine the entire CLI, both for the miner and the repairer, with a more modern library (JSAP 2.1 was released in 2006).


khaes-kth:
Hello @slarse,

Thanks for the PR. A useful feature + a clean and clear implementation.

Could you please a couple of lines to the readme file to explain the new parameter?

Thanks,

slarse:
@khaes-kth Which readme do you mean? The primary README.md in the root of the project does not mention the miner, and the `docs/MINE_SONAR_WARNINGS.md` file does not describe any of the current arguments.

I think adding a description of this argument would probably fall under #148

khaes-kth:
@slarse You are right. The miner README does not describe arguments because the arguments are already described in the root READMe.
I think we have two options: 1- '--ruleType' should also work for repairing (not only mining). In this case, the description could be added to the primary README.md. 2- we can change --ruleType to --minerRuleType, and add a description for this argument to miner README.md file.
WDYT?

slarse:
@khaes-kth 

> The miner README does not describe arguments because the arguments are already described in the root READMe.

There are several miner-specific arguments that aren't described anywhere, at least `--gitReposList` and `--tempDir` from just a glance. So I still think that in any way documenting an argument exclusive to the miner outside of using `--help` in the CLI, should be deferred to #148 

> 1- '--ruleType' should also work for repairing (not only mining).

I strongly disagree with doing this (and definitely would not do it in this PR). It's a feature we don't know if we need, and so should not add it. At this point, we don't even know if it's useful for developers to be able to repair multiple rules in one invocation of Sorald.

> 2- we can change --ruleType to --minerRuleType, and add a description for this argument to miner README.md file

I don't see a reason to change the argument name to `--minerRuleTypes`. Anyone invoking the miner will know that the arguments they specify will be related to the miner. Similarly, `--gitReposList` isn't called `--minerGitReposList`, even though it's exclusive to the miner.

On a related note, the listing in the root README is just the output from `--help`, which in my opinion isn't useful to even have in the README as it's liable to go out-of-date. Which, in fact, it has already as I forgot to update it when I added the latest processor. I think it's better to either a) automatically generate that documentation in CI, or b) simply refer users to run `--help` for detailed usage information, and have some usage _examples_ in the repo docs instead.

PR #191, opened 2020-11-04
feat: Add XxeProcessingProcessor (SonarSource rule 2755)

slarse:
This PR adds a processor for the rule [XML parsers should not be vulnerable to XXE attacks](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2755).

Example transformation:

```diff
+import javax.xml.XMLConstants;
 import org.w3c.dom.Document;
 import org.xml.sax.InputSource;

 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;

 public class DocumentBuilderNoSecurity {
     public static Document parse(String xmlFile) throws Exception {
         DocumentBuilderFactory df = DocumentBuilderFactory.newInstance();
+        df.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+        df.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
         DocumentBuilder builder = df.newDocumentBuilder();
         return builder.parse(new InputSource(xmlFile));
     }
 }
```

See #51 for rationale in choosing this processor.

This is heavily WIP at the moment. The rule is complex and can take on many different shapes. For this initial PR, I focus solely on the `DocumentBuilderFactory` variations. The current variations that I know of are:

1. A local variable of type `DocumentBuilderFactory` is declared and initialized. See example transformation above.
    - It's possible that the declaration and initialization are separated, but I've _never_ seen that happen, and so it doesn't seem worth it to account for. As such, the current implementation deals with this case well enough.
2. A local variable of type `DocumentBuilder` is created by chaining `DocumentBuilderFactory.newInstance().newDocumentBuilder()`.
3. Like 1, but instead a field is used.
    - Here, initialization may very well occur in a constructor, and so to be prudent one should attempt to locate any initialization of the field.
4. Like 2, but instead a field is used.

Currently, I've solved case 1. It's rather simple and requires only the addition of a few statements. Cases 2-4 require rewriting the existing code. I think the simplest way to do this is to replace the statements with a method invocation. For example, for case 2, one could substitute `DocumentBuilderFactory.newInstance().newDocumentBuilder()` for `createDocumentBuilder()`, which is defined like so.

```java
private static DocumentBuilder createDocumentBuilder() {
    DocumentBuilderFactory df = DocumentBuilderFactory.newInstance();
    df.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
    df.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
    return df.newDocumentBuilder();
}
```
The tricky part with cases 3 and 4 is that the initialization of the field may very well be separate from the declaration. As such, all initializations must be located and replaced with an invocation to the replacement method.

Finally, there is one more concern: in cases 1 and 3, it's possible that just one of the required attributes is set. We ideally don't want to duplicate them, and so should look for already existing invocations to `df.setAttribute`. Sometimes, the `XMLConstants.FEATURE_SECURE_PROCESSING` is set to `true`. We want to remove that if we find it, as it either does nothing, or is an obscure way to set the two attributes we actually want to set.

### DIFF
--- a/src/main/java/sorald/processor/XxeProcessingProcessor.java
+++ b/src/main/java/sorald/processor/XxeProcessingProcessor.java
@@ -22,34 +22,55 @@ import spoon.reflect.reference.CtTypeReference;
         key = 2755,
         description = "XML parsers should not be vulnerable to XXE attacks")
 public class XxeProcessingProcessor extends SoraldAbstractProcessor<CtInvocation<?>> {
+    private static final String DOCUMENT_BUILDER_FACTORY = "DocumentBuilderFactory";
+    private static final String DOCUMENT_BUILDER = "DocumentBuilder";
+
     @Override
     public boolean isToBeProcessed(CtInvocation<?> candidate) {
         return super.isToBeProcessedAccordingToStandards(candidate)
-                && candidate.getType().getSimpleName().equals("DocumentBuilderFactory")
+                && candidate.getType().getSimpleName().equals(DOCUMENT_BUILDER_FACTORY)
                 && candidate.getParent() instanceof CtLocalVariable<?>;
     }
 
     @Override
     public void process(CtInvocation<?> element) {
-        CtLocalVariable<?> builderFactoryVar = (CtLocalVariable<?>) element.getParent();
-        CtBlock<?> block = (CtBlock<?>) builderFactoryVar.getParent();
+        super.process(element);
+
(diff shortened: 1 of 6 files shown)


slarse:
Replaced by #198

PR #197, opened 2020-11-06
feat: Add IncompleteProcessor annotation

### LINKED ISSUE #189: Add annotation to signify that a processor is not a full fix for its associated rule
When implementing fixes for vulnerabilities, I suspect that in a lot of cases, we can't make a "complete" fix that always works. For example, [Mutable members should not be stored or returned directly](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2384) is probably impossible to create a one-size-fits-all template solution for. It may also be the case that some rule is possible but complex to create a full fix for, and as such we may want to implement it partially at first.

I think that it may be a good idea to somehow do some bookkeeping on which processors are supposedly complete, and which are known to be only partial solutions for their rule. One idea would be to keep that information in the `HANDLED_RULES.md` file, but that may cause it to go out of sync. Another idea would be to add it as a value to the `@ProcessorAnnotation`. That would force one to take a stance on the issue at least for the initial implementation.

slarse:
Fix #189 

As explained in #189, many processors will be incomplete fixes for their associated rules. We need to make this known both to ourselves (for further development), and to users (to avoid misunderstandings). This PR adds the `IncompleteProcessor` annotation, which can be applied like so.

```java
@IncompleteProcessor(description = "does not fix variable naming")
@ProcessorAnnotation(key = 1444, description = "\"public static\" fields should be constant")
public class PublicStaticFieldShouldBeFinalProcessor extends SoraldAbstractProcessor<CtField<?>> {
```

The description is then added to the `--help` message in the CLI.

```bash
Arguments: 

  [--ruleKeys ruleKeys1,ruleKeys2,...,ruleKeysN ]
        Choose one or more of the following rule keys (use ',' to separate
        multiple keys):
        2272: "Iterator.next()" methods should throw "NoSuchElementException"
        1860: Synchronization should not be based on Strings or boxed primitives
        2116: "hashCode" and "toString" should not be called on array instances
        1444: "public static" fields should be constant
                (incomplete: does not fix variable naming)
        2184: Math operands should be cast before assignment
```

The README has also been updated to reflect his new notion of incomplete processors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ The arguments are the following:
         2272: "Iterator.next()" methods should throw "NoSuchElementException"
         2116: "hashCode" and "toString" should not be called on array instances
         1860: Synchronization should not be based on Strings or boxed primitives
+        1444: "public static" fields should be constant
+                (incomplete: does not fix variable naming)
         2184: Math operands should be cast before assignment
         4973: Strings and Boxed types should be compared using "equals()"
         2095: Resources should be closed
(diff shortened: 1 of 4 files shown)


slarse:
@khaes-kth Don't forget about reviewing this one :)

khaes-kth:
@slarse Seems fine to me. Can I merge it?

slarse:
@khaes-kth yes, ready to merge :)

PR #200, opened 2020-11-09
refactor: Generate the Processors class

### LINKED ISSUE #195: Generate the "Processors" class through annotation processing
Currently, the `Processors` class uses runtime introspection to inspect annotations. To clean this up, I'm implementing a processor annotator, that can also be used for processing the future processor proposed by #189 .

slarse:
Fix #195 

This PR splits the project into two submodules, `sorald-annotations` and `sorald`, such that annotation processors can be defined in `sorald-annotations` and used in `sorald`. It also includes the annotation processor `GenerateProcessorsClass`, which generates the `sorald.Processors` class based on the `ProcessorAnnotation` annotations and makes the rule descriptions available as a literal constant `Processors.RULE_DESCRIPTIONS`. The generate class looks like so:

```java
package sorald;
class Processors {
    private static final java.util.Map<java.lang.Integer, java.lang.Class<? extends sorald.processor.SoraldAbstractProcessor>> RULE_KEY_TO_PROCESSOR = new java.util.HashMap() {{
    put(2184,sorald.processor.CastArithmeticOperandProcessor.class);
    put(1444,sorald.processor.PublicStaticFieldShouldBeFinalProcessor.class);
    put(1948,sorald.processor.SerializableFieldInSerializableClassProcessor.class);
    put(3984,sorald.processor.UnusedThrowableProcessor.class);
    put(2116,sorald.processor.ArrayHashCodeAndToStringProcessor.class);
    put(4973,sorald.processor.CompareStringsBoxedTypesWithEqualsProcessor.class);
    put(2142,sorald.processor.InterruptedExceptionProcessor.class);
    put(2111,sorald.processor.BigDecimalDoubleConstructorProcessor.class);
    put(2164,sorald.processor.MathOnFloatProcessor.class);
    put(2272,sorald.processor.IteratorNextExceptionProcessor.class);
    put(1854,sorald.processor.DeadStoreProcessor.class);
    put(1860,sorald.processor.SynchronizationOnStringOrBoxedProcessor.class);
    put(2095,sorald.processor.UnclosedResourcesProcessor.class);
    put(2204,sorald.processor.EqualsOnAtomicClassProcessor.class);
    put(1656,sorald.processor.SelfAssignementProcessor.class);
    put(3067,sorald.processor.SynchronizationOnGetClassProcessor.class);
    put(3032,sorald.processor.GetClassLoaderProcessor.class);
    put(2167,sorald.processor.CompareToReturnValueProcessor.class);
    }}
    ;

    public static final java.lang.Class getProcessor(int key) {
        return RULE_KEY_TO_PROCESSOR.get(key);
    }

    public static final java.lang.String RULE_DESCRIPTIONS = "2184: Math operands should be cast before assignment\n1444: \"public static\" fields should be constant\n1948: Fields in a \"Serializable\" class should either be transient or serializable\n3984: Exception should not be created without being thrown\n2116: \"hashCode\" and \"toString\" should not be called on array instances\n4973: Strings and Boxed types should be compared using \"equals()\"\n2142: \"InterruptedException\" should not be ignored\n2111: \"BigDecimal(double)\" should not be used\n2164: Math should not be performed on floats\n2272: \"Iterator.next()\" methods should throw \"NoSuchElementException\"\n1854: Unused assignments should be removed\n1860: Synchronization should not be based on Strings or boxed primitives\n2095: Resources should be closed\n2204: \".equals()\" should not be used to test the values of \"Atomic\" classes\n1656: Variables should not be self-assigned\n3067: \"getClass\" should not be used for synchronization\n3032: JEE applications should not \"getClassLoader\"\n2167: \"compareTo\" should not return \"Integer.MIN_VALUE\"";
}
```

The reason this class needs to be generated is that when using `picocli`, the descriptions of options must be constants as they are defined as annotations. As such, they cannot be generated at runtime as is currently done in `Processors.getRuleDescriptions()`. I haven't been able to find a way around this, and so I think that if we want to use picocli, we must generate the rule descriptions at compile time.

I'm a bit conflicted about this refactor, as it adds significant complexity. `GenerateProcessorsClass` is much more complex than the class it generates (`Processors`). It just seems silly. On the other hand, there are upsides.

1. We can use picocli!
2. Much safer; using reflection to find classes is inherently insecure.
3. Faster as it's done at compile time.
4. The infrastructure for more annotation processing in the future is in place.

@fermadeiral @khaes-kth @monperrus any thoughts on this?


monperrus:
Have a look at what we do in Spoon in CtGenerationTest

slarse:
@monperrus Thanks! I did however figure out how to do "proper" annotation processing without splitting the package into multiple modules by precompiling the annotations subpackage before the normal compile phase. The missing piece of the puzzle was that I had to explicitly disable annotation processing for the precompilation. See #215

monperrus:
nice solution

PR #199, opened 2020-11-06
feat: Add support chained for instantiation of DocumentBuilder (SonarRule 2755)

slarse:
Working toward #194 

This is an expansion upon #198 (**wich must be merged before this PR!**) XxeProcessingProcessor which repairs [rule 2755](https://rules.sonarsource.com/java/RSPEC-2755). This is the fix for case 2 as described in #194 

> 2. A local variable of type DocumentBuilder is created by chaining DocumentBuilderFactory.newInstance().newDocumentBuilder().

The transformation looks like so:

```diff
         // somewhere in a method body
-        DocumentBuilder builder = DocumentBuilderFactory.newInstance().createDocumentBuilder();
+        DocumentBuilder builder = createDocumentBuilder();
         [...]
     }

+    private static javax.xml.parsers.DocumentBuilder createDocumentBuilder() {
+         DocumentBuilderFactory df = DocumentBuilderFactory.newInstance();
+         df.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+         df.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+         return df.newDocumentBuilder();
+     }
```

This also lays the groundwork for fixing fields, as initialization must be substituted with a method call.

PR #198, opened 2020-11-06
feat: Add XxeProcessingProcessor (SonarSource rule 2755)

slarse:
Working toward #194 

This PR adds a processor for the rule [XML parsers should not be vulnerable to XXE attacks](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2755).

This is a **partial** fix for the rule. Very partial, in fact. See #194 for more details. Here is the transformation implemented with this PR:

```diff
+import javax.xml.XMLConstants;
 import org.w3c.dom.Document;
 import org.xml.sax.InputSource;

 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;

 public class DocumentBuilderNoSecurity {
     public static Document parse(String xmlFile) throws Exception {
         DocumentBuilderFactory df = DocumentBuilderFactory.newInstance();
+        df.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+        df.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
         DocumentBuilder builder = df.newDocumentBuilder();
         return builder.parse(new InputSource(xmlFile));
     }
 }
```

This processor is a good testing ground for highly complex processors. The case fixed in this PR is the simplest one, yet it is relatively complex compared to our other processors.

Part 2 is available in #199, but this PR must be merged first and then I need to rebase as there will be conflicts. The `XxeProcessingProcessor` will also be annotated with the new `IncompleteProcessor` annotation (#197) once that is merged.

Issue #194, opened 2020-11-05
Gradually implement processor for Sonar rule 2755 (XML parsers should not be vulnerable to XXE attacks)


slarse:
This is related to introducing a processor for  [XML parsers should not be vulnerable to XXE attacks](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2755). It's a rather complicated processor, so I'm introducing gradually. As can be noted, it covers a variety of libraries, but there are only a couple of major variations in usage.

I'm initially targeting only the `DocumentBuilderFactory` use cases to see if this is even feasible to repair adequately. I've found 6 major variations to what it can look like. Here's an example repair of the simplest case:

```diff
+import javax.xml.XMLConstants;
 import org.w3c.dom.Document;
 import org.xml.sax.InputSource;

 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;

 public class DocumentBuilderNoSecurity {
     public static Document parse(String xmlFile) throws Exception {
         DocumentBuilderFactory df = DocumentBuilderFactory.newInstance();
+        df.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+        df.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
         DocumentBuilder builder = df.newDocumentBuilder();
         return builder.parse(new InputSource(xmlFile));
     }
 }
```
The cases I've found are:

1. A local variable of type `DocumentBuilderFactory` is declared and initialized. See example transformation above.
    - It's possible that the declaration and initialization are separated, but I've _never_ seen that happen, and so it doesn't seem worth it to account for. As such, the current implementation deals with this case well enough.
2. A local variable of type `DocumentBuilder` is created by chaining `DocumentBuilderFactory.newInstance().newDocumentBuilder()`.
3. Chaining everything together in one expression, like so: `DocumentBuilderFactory.newInstance().newDocumentBuilder().parse("xmlfile.xml")`
4. Like 1, but instead a field is used.
    - Here, initialization may very well occur in a constructor, and so to be prudent one should attempt to locate any initialization of the field.
5. Like 2, but instead a field is used.
6. Like 3, but instead a field is used.

[...]

PR #201, opened 2020-11-09
 feat: Add XxeProcessingProcessor (SonarSource rule 2755)

slarse:
Working toward #194

This PR adds a processor for the rule [XML parsers should not be vulnerable to XXE attacks](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2755), and replaces #198 and #199 . It's been long in the making as I made some initial mistakes in #198 with regard to the level of abstraction in the transformation.

This is a **partial** fix for rule 2755, targeting only the DocumentBuilderFactory variations of violations. The transformation is fairly simple: for any flagged invocation of `DocumentBuilderFactory.newInstance()`, replace it with an invocation to another method `createDocumentBuilderFactory()` that looks like this.

```java
    private static DocumentBuilderFactory createDocumentBuilderFactory() {
        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
        factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
        factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
        return factory;
    }
```
So, it's a simple substitution of one method invocation for another, both returning the same value. In theory, it cannot go wrong, unless there is already a method with the signature `createDocumentBuilderFactory()` (we should probably add that as a check in the future).

Here's an example transformation.

```diff
         // somewhere in a method body
-        DocumentBuilder builder = DocumentBuilderFactory.newInstance().createDocumentBuilder();
+        DocumentBuilder builder = createDocumentBuilderFactory().createDocumentBuilder();
         [...]

         // somewhere in a method body
-        DocumentBuilderFactory df = DocumentBuilderFactory.newInstance();
+        DocumentBuilderFactory df = createDocumentBuilderFactory();
         [...]

+    private static DocumentBuilderFactory createDocumentBuilderFactory() {
+         DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+         factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+         factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+         return factory;
+     }
```

See the added test files for full transformations.

### Limitations
* As stated, only handles `DocumentBuilderFactory.newInstance()`
* Does not look for "partially secure" instances, such as if just one of the attributes are set, or if the `FEATURE_SECURE_PROCESSING` (which is not secure in general) is used
    - Therefore, there may be duplication in setting of attributes
* Fields don't work, as I can't get Sonar to flag fields as noncompliant. I think that the transformation would work if only Sonar would actually flag the fields as noncompliant, as all it does is replace the invocation to `DocumentBuilderFactory.newInstance()` I'll look more into this later.


slarse:
@fermadeiral FYI I noticed that the documentation in handled rules was incorrect (it reflected the old transformation in #198), and fixed it. Code is unchanged since I set this as ready-for-review.

fermadeiral:
@slarse nice, even that it's a partial fix! Also, thank you for documenting the limitations in the PR.

monperrus:
Excellent PR, thanks!

PR #202, opened 2020-11-10
feat: Add support for TransformerFactory in XxeProcessingProcessor

slarse:
Working toward #194 

Depends on #201 to be merged first.

This PR adds support for the `TransformerFactory` in the `XxeProcessingProcessor` which handles the rule [XML parsers should not be vulnerable to XXE attacks](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2755).

The transformation is identical to the one in #201, with the one exception that `XMLConstants.ACCESS_EXTERNAL_SCHEMA` is replaced with `XMLConstants.ACCESS_EXTERNAL_STYLESHEET`, as per the recommendations in [2755](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2755).


slarse:
Just hardening a test case to avoid coverage drop

slarse:
Null-check was seemingly redundant after re-ordering the call to `super.isToBeProcessedAccordingToStandards`, so I've removed it until I can find a test case that causes a null-exception again.

fermadeiral:
That is a fast evolution of the new processor. Thank you!

PR #203, opened 2020-11-10
feat: Add support for XMLInputFactory in XxeProcessingProcessor

slarse:
Working toward #194 

Depends on #201 and #202 to be merged first.

This PR adds support for the `XMLInputFactory` in the `XxeProcessingProcessor` which handles the rule [XML parsers should not be vulnerable to XXE attacks](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2755).

The transformation is identical to the one in #201, with the one exception that the setter method is called `setProperty` instead of `setAttribute`.


slarse:
Opening new PR to get better coverage data.

PR #206, opened 2020-11-10
fix: Avoid processing directories with .java extension as files

### LINKED ISSUE #204: Directories ending in ".java" cause Sorald to crash
It thinks it's a viable java file.

slarse:
Fix #204 

This fixes a crash that occurs when you apply Sorald to a project that has a directory ending in `.java`.

This also applies to the miner, but I'll fix that separately.

### DIFF
--- a/src/main/java/sorald/FileUtils.java
+++ b/src/main/java/sorald/FileUtils.java
@@ -1,8 +1,12 @@
 package sorald;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.List;
+import java.util.stream.Collectors;
 
 public class FileUtils {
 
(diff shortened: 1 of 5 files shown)


src/main/java/sorald/processor/SoraldAbstractProcessor.java
```diff
@@ -39,10 +39,10 @@ public SoraldAbstractProcessor initResource(String originalFilesPath, File baseD
             if (file.isFile()) {
                 filesToScan.add(file.getAbsolutePath());
             } else {
-                try (Stream<Path> walk = Files.walk(Paths.get(file.getAbsolutePath()))) {
+                try {
                     filesToScan =
-                            walk.map(x -> x.toFile().getAbsolutePath())
-                                    .filter(f -> f.endsWith(Constants.JAVA_EXT))
```
slarse, line 45: This is what caused the crash; it doesn't check if the file is actually a regular file (so it could be a directory ending with `.java`).

src/main/java/sorald/FileUtils.java
```diff
@@ -52,4 +57,34 @@ public static boolean deleteDirectory(File directoryToBeDeleted) {
...
+     * @param ext A file extension including the leading dot
+     * @return All files in the given directory or any subdirectory with a matching extension
+     */
+    public static List<File> findFilesByExtension(File directory, String ext) throws IOException {
+        if (!directory.isDirectory()) {
+            throw new IllegalArgumentException(directory.toString() + " is not a directory");
+        }
+        try (Stream<Path> files = Files.walk(directory.toPath())) {
+            return files.map(Path::toFile)
+                    .filter(File::isFile)
```
slarse, line 75: This is the key :)

PR #213, opened 2020-11-11
feat: Add support for XMLInputFactory in XxeProcessingProcessor

slarse:
Working toward #194 

This PR adds support for the `XMLInputFactory` in the `XxeProcessingProcessor` which handles the rule [XML parsers should not be vulnerable to XXE attacks](https://rules.sonarsource.com/java/type/Vulnerability/RSPEC-2755).

The transformation is identical to the one in #201, with the one exception that the setter method is called `setProperty` instead of `setAttribute`.

### DIFF
--- a/docs/HANDLED_RULES.md
+++ b/docs/HANDLED_RULES.md
@@ -410,6 +410,7 @@ Currently, we target the following types:
 
 * `DocumentBuilderFactory`
 * `TransformerFactory`
+* `XMLInputFactory`
 
 The transformation is highly similar regardless of type, and consists of replacing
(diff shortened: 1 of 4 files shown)


slarse:
Coverage drop is due to this switch statement: https://codecov.io/gh/SpoonLabs/sorald/pull/213/diff?src=pr&el=tree#D2-230

The default case is never hit (the +1 partial), and there are two lines in it to throw an exception (+2 misses). It seems a bit pointless to test that case, it's just a guard in case we add support for a new factory type but forget to add it in any of the cases down there.

fermadeiral:
Nice!

PR #212, opened 2020-11-11
chore: Switch to Java 11

slarse:
Fix #211 

We're dropping support for Java 8 and now only support Java 11+. This PR updates the pom and related documentation.

PR #214, opened 2020-11-11
fix: Avoid mining directory with Java extension as file

### LINKED ISSUE #207: Directories ending with ".java" cause the miner to crash
For the same reason as #204 . Can be fixed by using the new method introduced in #206

slarse:
Fix #207 

~Depends on #212~

This fixes a crash where the sonar miner would try to read directories that end in `.java` as files. This is essentially the same fix as in #206


khaes-kth:
@slarse Thanks for the PR.

Issue #216, opened 2020-11-12
Redesign SoraldAbstractRepair to be more testable


slarse:
There are a number of things that need to be improved here.

* I need a method that returns a Spoon model in order to test some of the finer details
    - It's very hard to test some of the finer details by first writing to a file, as bugs can then be located in the processor that performed the repair, the printer, or the parser
    - This is also a good step toward defining a nice API for Sorald
* The `repair` method does too many things. It reads files, does repairs, write files, and also does user I/O. All of those things need to be pieced out into methods of their own.
* I don't think having `SoraldAbstractRepair` as an abstract base class is meaningful. It would be better of as an interface. The state currently kept in the base class could simply be passed as parameters to the methods. It would go some way to untangling the structure.

PR #217, opened 2020-11-13
feat: Auto-import types for added references and use unqualified names

### LINKED ISSUE #208: Automatically import types for type references added by Sorald
Right now, we set auto-import to true, and then in the typical case also define a custom pretty-printer creator:

https://github.com/SpoonLabs/sorald/blob/a39cef1b8c63f2e7bf6c2395492aaa481f6452cf/src/main/java/sorald/SoraldAbstractRepair.java#L48-L62

The custom pretty-printer creator overrides auto-import, as the auto-import processors are only added if a custom pretty-printer creator is not defined in the environment. Therefore, Sorald always prints fully qualified names for type references added in the transformation unless the package reference is explicitly set to be implicit (which is done e.g. here: https://github.com/SpoonLabs/sorald/blob/a39cef1b8c63f2e7bf6c2395492aaa481f6452cf/src/main/java/sorald/processor/XxeProcessingProcessor.java#L197-L201

Of course, that's really annoying to remember to do in all transformations, and so we should try to automate this task for all processors.

However, we don't actually want to use the full auto-import functionality, as that may automatically import types for references that were defined with qualified names before running Sorald. We probably want to do something very similar to [what's done in Spoon with auto-imports](https://github.com/INRIA/spoon/blob/8a2cce96f45d74d6cc6b3d678947cee28cfdf991/src/main/java/spoon/support/StandardEnvironment.java#L643-L659), but create our own variation of the `ForceImportProcessor` that only operates on elements introduced by ourselves.

slarse:
Fix #208 

This PR introduces a feature where Sorald automatically imports types and uses unqualified names for any type reference that Sorald adds. This is really important for "merge-ready patches", as developers typically don't write the fully qualified names of types.

It fixes two things:

* `Environment.setAutoImports(true)` has no effect when using a custom pretty-printer creator (see inria/spoon#3690)
* We don't actually want the full auto-import feature, as that may cause Sorald to import types that were used with qualified references in the original file, and the change those qualified references to unqualified. That's unrelated to any repair, and so we don't want to do it.

I don't have good testing for this yet, because I need a more granular API for the repair process in order to test the finer details (see #216). This is currently implicitly tested in that some test cases have types that are added, and thus need to be imported (imports are checked against the expected file in `ProcessorTest`). There's also a new pair of test files with fully qualified names, that ensures that these aren't imported (again verified in `ProcessorTest`).


src/test/resources/processor_test_files/2116_ArrayHashCodeAndToString/ArrayHashCodeAndToString.java.expected
```diff
@@ -1,12 +1,14 @@
...
+import java.util.Arrays;
+
 public class ArrayHashCodeAndToString {
 
 	// Tests from https://rules.sonarsource.com/java/type/Bug/RSPEC-2116
 	public static void main( String[] args ) {
-		String argStr = java.util.Arrays.toString(args); // Noncompliant
-		int argHash = java.util.Arrays.hashCode(args); // Noncompliant
+		String argStr = Arrays.toString(args); // Noncompliant
+		int argHash = Arrays.hashCode(args); // Noncompliant
```
slarse, line 11: Example of how this update changes the output of a processor (note also that `java.util.Arrays` is imported at the top of the file)

src/main/java/sorald/processor/XxeProcessingProcessor.java
```diff
@@ -157,32 +152,14 @@ public void process(CtInvocation<?> element) {
...
-        CtImport requiredImport = getFactory().createImport(mustBeImported.getReference());
-
-        for (CtImport imp : cu.getImports()) {
-            if (imp.toString().equals(requiredImport.toString())) {
-                return;
-            }
-        }
-
-        cu.getImports().add(requiredImport);
-    }
```
slarse, line 176: Special handling of imports in single processor that can now be removed

src/test/resources/processor_test_files/1444_PublicStaticFieldShouldBeFinal/ExistingQualifiedNamesRemainQualified.java.expected
```diff
@@ -0,0 +1,14 @@
...
+public class ExistingQualifiedNamesRemainQualified {
+    public static final String s = "some value";
+
+    public static void main(String[] args) {
+        // now this is what we're actually testing, if Sorald is too aggresive in processing types
+        // for auto-importing, java.util.ArrayList and javax.xml.XMLConstants will be imported
+        java.util.ArrayList<Integer, String> list = new java.util.ArrayList<>();
+        String constant = javax.xml.XMLConstants.ACCESS_EXTERNAL_DTD;
+    }
+}
```
slarse, line 14: This test case file checks that we don't accidentally import types for already existing qualified names.

slarse:
Sorry, wait, forgot some docs!

slarse:
Okay, all good, ready for review!

fermadeiral:
> This is really important for "merge-ready patches", as developers typically don't write the fully qualified names of types.

I totally agree. Thanks for handling this!

Issue #192, opened 2020-11-04
Replace the CLI library


slarse:
If we want Sorald to be user-friendly, I think we should replace JSAP as the CLI library. Not only is it somewhat clunky to use, the latest release is also 14 years old.

There are much better alternatives out there. One that I like is [picocli](https://picocli.info/), which is very easy to use and actively maintained. It also has very nice support for subcommands, such that we could integrate the miner and repairer into the same CLI. For example:

```bash
# run miner
$ java -jar sorald.jar mine-warnings [arguments ...]
# run repair
$ java -jar sorald.jar repair [arguments ...]
```

This would be much more user-friendly than the current state of affairs, where the two CLIs are entirely disconnected from each other.

```bash
# run miner
java -cp sorald.jar sorald.miner.MineSonar [arugments ...]
# run repair
java -jar sorald.jar [arguments ...]
```

Rewriting the entire CLI wouldn't be too much work, as it's fairly small.

monperrus:
Yes, we proceed with picocli!

slarse:
Doing this all in one PR would make for a very big PR, so I'm going to build out the new CLI alongside the existing one in multiple PRs, and then do a swap when we're satisfied with it.

slarse:
Pankti also uses picocli, so there are more of us in the group using it :)

https://github.com/castor-software/pankti/blob/0fddf273a8a6acbaba51a318fcc7fc6e5b84058f/pankti-extract/pom.xml#L54-L58

PR #218, opened 2020-11-13
feat: Use picocli for the repair application

slarse:
#192 

A drop-in replacement for the old CLI in the primary repair application (i.e. not the warnings miner). After this, I will also convert the warnings miner, and then finally I will compose the two CLIs into one, in two subcommands (e.g. `repair` for this one and `mine-warnings` for the other one).


src/main/java/sorald/cli/Cli.java
```diff
@@ -0,0 +1,163 @@
...
+            if (gitRepoPath != null) {
+                config.setGitRepoPath(gitRepoPath.getAbsolutePath());
+            }
+            config.setPrettyPrintingStrategy(prettyPrintingStrategy);
+            config.setFileOutputStrategy(fileOutputStrategy);
+            config.setMaxFixesPerRule(maxFixesPerRule);
+            config.setMaxFilesPerSegment(maxFilesPerSegment);
+            config.setRepairStrategy(repairStrategy);
+            return config;
+        }
```
slarse, line 143: This is one place where we really lower complexity: picocli does a whole lot for us in parsing. See `Main.initConfig` above for what this used to look like.

src/main/java/sorald/Main.java
```diff
@@ -1,210 +1,14 @@
...
-        this.getConfig().setMaxFixesPerRule(arguments.getInt(Constants.ARG_MAX_FIXES_PER_RULE));
-        int segmentSize = arguments.getInt(Constants.ARG_MAX_FILES_PER_SEGMENT);
-        if (segmentSize <= 0) {
-            throw new RuntimeException("Segment size need to be bigger than 0");
-        }
-        this.getConfig().setMaxFilesPerSegment(segmentSize);
-        this.getConfig()
-                .setRepairStrategy(
-                        RepairStrategy.valueOf(arguments.getString(Constants.ARG_REPAIR_STRATEGY)));
-    }
```
slarse, line 172: This is the old method corresponding to the new `createConfig` below.

fermadeiral:
Thanks @slarse!

PR #219, opened 2020-11-16
feat(rule 2755): Move calls setting FEATURE_SECURE_PROCESSING into factory creator method

slarse:
#194 

This is what [rule 2755](https://rules.sonarsource.com/java/RSPEC-2755) says about `FEATURE_SECURE_PROCESSING`

>Avoid FEATURE_SECURE_PROCESSING feature to protect from XXE attacks because depending on the implementation:
>    * it has no effect to protect the parser from XXE attacks but helps guard against excessive memory consumption from XML processing.
>    * or it's just an obscur shortcut (it could set ACCESS_EXTERNAL_DTD and ACCESS_EXTERNAL_SCHEMA to "" but without guarantee).

Calling `setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true)` is an unreliable way of setting secure processing, but as it does have an effect (e.g. memory consumption) I don't think it's appropriate to remove it. As such, if present, I move it into the factory creator method. Below is a patch generated for RedHat-RSP. Note that the `createTransformerFactory` method sets the `FEATURE_SECURE_PROCESSING` attribute, as it was set for the original factory, but `createDocumentBuilderFactory` does not.

```diff
                Document document = null;
                try {   
-                       DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+                       DocumentBuilderFactory factory = createDocumentBuilderFactory();
                        DocumentBuilder parser = factory.newDocumentBuilder();
                        document = parser.parse(new InputSource(in));
                        Node node = document.getFirstChild();
@@ -126,7 +126,7 @@ public final class XMLMemento implements IMemento {
        public static XMLMemento createWriteRoot(String type) {
                Document document;
                try {
-                       document = DocumentBuilderFactory.newInstance().newDocumentBuilder().newDocument();
+                       document = createDocumentBuilderFactory().newDocumentBuilder().newDocument();
                        Element element = document.createElement(type);
                        document.appendChild(element);
                        return new XMLMemento(document, element);
@@ -348,8 +348,7 @@ public final class XMLMemento implements IMemento {
                Result result = new StreamResult(os);
                Source source = new DOMSource(factory);
                try {
-                       TransformerFactory factory = TransformerFactory.newInstance();
-                       factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+                       TransformerFactory factory = createTransformerFactory();
                        Transformer transformer = factory.newTransformer();
                        transformer.setOutputProperty(OutputKeys.INDENT, "yes"); //$NON-NLS-1$
                        transformer.setOutputProperty(OutputKeys.METHOD, "xml"); //$NON-NLS-1$
@@ -436,4 +435,18 @@ public final class XMLMemento implements IMemento {
                }
                return ""; //$NON-NLS-1$
        }
+
+       private static DocumentBuilderFactory createDocumentBuilderFactory() {
+               DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+               factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+               factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+               return factory;
+       }
+
+       private static TransformerFactory createTransformerFactory() {
+               TransformerFactory factory = TransformerFactory.newInstance();
+               factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+               factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+               factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_STYLESHEET, "");
+               return factory;
+       }
```


slarse:
Hmm, instead of removing the `FEATURE_SECURE_PROCESSING`, perhaps moving it into the builder method would be an option.

slarse:
With 2ef2c9e I move the setting of `FEATURE_SECURE_PROCESSING` into the factory creator method, instead of removing it. On redhat-rsp, it gives the exact same diff as in the opening comment, but with a `createTransformerFactor` method that looks like so:

```diff
+       private static TransformerFactory createTransformerFactory() {
+               TransformerFactory factory = TransformerFactory.newInstance();
+               factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+               factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_STYLESHEET, "");
+               factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+               return factory;
+       }
```

slarse:
Ready for review! What do you think, is this a reasonable solution?

Issue #193, opened 2020-11-04
in addition to the diff, generate a JSON file


monperrus:
in addition to the diff, generate a JSON file containing the metadata about the fix, that could be used later for the explanation in the PR message.

slarse:
As per our discussion today, we want at least the following:

* For each repair
    - The rule that was repaired
    - Location of the repair (relative from repo root filepath + line nr?)
* Failures (when a processor crashes, see #210 )
* Various statistics
    - Time to parse
    - Time to repair

Ping @khaes-kth @fermadeiral @nharrand just so you know this issue exists!

fermadeiral:
I think it is also important to record the number of warnings per rule before and after the transformations.

fermadeiral:
Maybe we should also record the args given as input so we can know how to reproduce something given a json file.

slarse:
> I think it is also important to record the number of warnings per rule before and after the transformations.

That would incur heavy overhead in Sorald. Perhaps we should have an extra argument, `--gatherStats`, or something like that?

> Maybe we should also record the args given as input so we can know how to reproduce something given a json file.

That is an excellent idea! For reproducibility, we should probably include the version number as well.

slarse:
As #220 is now merged, I'll start working on this. There may be some overlap in JSON stuff with @khaes-kth's work on #225, but we can fix that after the fact.

slarse:
To make this modular, I'm thinking that we'll want to pass some sort of event handler to the repair process. And then, for example, when Sorald starts parsing the input, it registers an event `PARSE_START` with the event handler, and the handler can do whatever it wants with that. When parsing ends, it registers `PARSE_END`, and that way we can time the parse time.

Similarly, we could pass `VIOLATION_DETECTED` when a violation is detected, and `VIOLATION_REPAIRED` (along with more info) whenever a violation is repaired. We can also squeeze in stuff like `PROCESSOR_CRASH` to cover #210. 

I haven't quite thought this through all the way. For example, different events will require different metainfo (perhaps a simple hashmap will do for that). But I think this could be a decent idea. I plan to work on this tomorrow, so hopefully I'll have a PR with a suggestion by the end of the day.

slarse:
To give an extremely rough idea of what I'm thinking, create an event handler like so and pass it to the `repair` method:

```java
            SoraldEventHandler eventHandler =
                    new SoraldEventHandler() {
                        private long parseStart = -1;
                        private long parseEnd = -1;

[...]

Issue #170, opened 2020-10-28
Segment strategy only repairs one rule


slarse:
If multiple rules are specified along with the segment strategy, only the first rule is processed, while the rest are ignored. The issue is right here: https://github.com/SpoonLabs/sorald/blob/c76d525603345ad82149951ce3f2c7339ce6236d/src/main/java/sorald/SegmentRepair.java#L29-L30

There's no loop to iterate through the rules, as is the case for the default strategy.

slarse:
To be clear, the PR that introduced segment mode (#134) clearly specifies that it's for _single rule repair_, so this is not an accident. There's just no check to verify that only a single rule is specified.

slarse:
This will probably be fixed as a side effect of #220, but a test case should still be added to verify this.

khaes-kth:
Hi Simon,

I was looking at #233. Before I merge that one, I want to know what exactly segment mode is supposed to do. I read the introducing PR but I still have doubts about it. Can you please define it quickly here?

slarse:
@khaes-kth It splits a project into multiple chunks, and builds the Spoon model for one chunk at a time. This is to avoid out-of-memory errors for very large projects. See the comments in #154 (where I was figuring out what it does) for more details.

khaes-kth:
Thanks a lot @slarse for the explanation.

PR #220, opened 2020-11-17
refactor: Redesign the repair strategies

### LINKED ISSUE #216: Redesign SoraldAbstractRepair to be more testable

slarse:
Fix #216 

This PR is not technically a refactor, as it coincidentally fixes a bug (#170), but it's close enough as to make that the most reasonable label :)

In any case, this PR does away with the strategy pattern that was applied to `SoraldAbstractRepair`, and the `DefaultRepair` and `SegmentRepair` implementations. The abstraction was very leaky, as the two concrete classes functioned fundamentally differently, requiring different options in the config, different handling of processors, etc etc. Code reuse was also poor, and `SegmentRepair` did not handle multiple rules (#170). This redesign fixes the problems in #170 simply by restructuring the code, the segment strategy can now repair multiple rules in the same way that the default does.

The new design is less object-oriented, and in my opinion significantly easier to understand. No inheritance required. This is in large part in preparation of gathering statistics (#193), as well as providing an actual API for Sorald's library use case. Note however that the `Repair` class is _not_ what I consider the public API of Sorald, as I want the public API to not have to deal with the `SoraldConfig` object.

The `repair`, `defaultRepair` and `segmentRepair` methods are however good indications of what the API will eventually look like, and are package-private for testing purposes.


src/main/java/sorald/Repair.java
```diff
@@ -0,0 +1,306 @@
...
+                .map(
+                        segment -> {
+                            processor.initResource(segment, inputBaseDir);
+                            Launcher launcher = createSegmentLauncher(segment);
+                            CtModel model = launcher.getModel();
+                            repairModelWithInitializedProcessor(model, processor);
+                            return model;
+                        })
+                .takeWhile(model -> processor.getNbFixes() < config.getMaxFixesPerRule());
+    }
```
slarse, line 118: This is essentially what became of the previous inheritance structure. This is much easier for me to wrap my head around. I envision the public API of Sorald to look something like this.

slarse:
@fermadeiral This is ready for review, but it will conflict with #218. So pick whichever one you feel like reviewing first!

slarse:
@fermadeiral Conflict fixed!

fermadeiral:
That's really great! Thanks @slarse!

PR #221, opened 2020-11-18
feat: Use picocli for miner application

### LINKED ISSUE #192: Replace the CLI library

slarse:
Fix #192 

This PR is _almost_ a drop-in replacement for the warnings miner CLI. There are two differences:

* `--statsOnGitRepos` is now actually a flag (i.e. you specify the option without argument), whereas before you had to write `--statsOnGitRepos true`.
* `--ruleTypes` was previously case-insensitive (e.g. you could specify `--ruleTypes BUG` or `--ruleTypes bug`. It's now case-sensitive and requires upper-case, but the CLI is very clear about that.

The next and final step is to incorporate the two CLIs as subcommands of a single CLI, and update the documentation accordingly. I initially did that in this PR, but it turned into such a big PR that I decided to split it up.

Note that some untested functionality is lost, such that checking that certain files/directories exist. I will test and re-implement this as part of a separate issue (see #222)


khaes-kth:
Thanks @slarse for the PR :)

PR #232, opened 2020-11-20
feat: Add statistics collection

slarse:
#193 

This PR represents my general idea for the architecture required for elaborate statistics collection. It's not complete by far in regards to the statistics outlined in #193, but I think this contains all of the parts that we need. It would be a good stepping stone to start with; it's big enough already :)

Put briefly, this PR introduces the `sorald.event.SoraldEventHandler` interface, which allows for arbitrary events to be registered based on the `sorald.event.EventType` enum. The interface is very simple, it only has a single method `registerEvent(SoraldEvent)`, where `SoraldEvent` is another functional interface with a single method `EventType type()`, denoting its type.

There's currently only a single event handler, `sorald.event.StatisticsCollector`, which simply collects some basic statistics. We'll want to expand upon this to collect the additional statistics mentioned in #193, but I don't think we need much more logic in there.

Here's a look at what a stats file currently contains:

```json
{
    "repairs": [
        {
            "ruleViolationPosition": "(/home/slarse/Documents/github/work/sorald/src/test/resources/processor_test_files/2755_XxeProcessing/DocumentBuilderChainedLocalVariable.java:9)",
            "ruleKey": "2755"
        },
        {
            "ruleViolationPosition": "(/home/slarse/Documents/github/work/sorald/src/test/resources/processor_test_files/2755_XxeProcessing/DocumentBuilderFullyChainedParse.java:9)",
            "ruleKey": "2755"
        },
        {
            "ruleViolationPosition": "(/home/slarse/Documents/github/work/sorald/src/test/resources/processor_test_files/2755_XxeProcessing/DocumentBuilderLocalVariable.java:9)",
            "ruleKey": "2755"
        },
        {
            "ruleViolationPosition": "(/home/slarse/Documents/github/work/sorald/src/test/resources/processor_test_files/2755_XxeProcessing/InsecureTransformerFactoryUsage.java:10)",
            "ruleKey": "2755"
        },
        {
            "ruleViolationPosition": "(/home/slarse/Documents/github/work/sorald/src/test/resources/processor_test_files/2755_XxeProcessing/InsecureXMLInputFactoryUsage.java:9)",
            "ruleKey": "2755"
        },
        {
            "ruleViolationPosition": "(/home/slarse/Documents/github/work/sorald/src/test/resources/processor_test_files/2755_XxeProcessing/InsecureXMLInputFactoryUsage.java:15)",
            "ruleKey": "2755"
        },
        {
            "ruleViolationPosition": "(/home/slarse/Documents/github/work/sorald/src/test/resources/processor_test_files/2755_XxeProcessing/SetFeatureSecureProcessing.java:11)",
            "ruleKey": "2755"
        }
    ],
    "repairTimeNs": 392604045,
    "originalArgs": [
        "--originalFilesPath",
        "./src/test/resources/processor_test_files",
        "--ruleKeys",
        "2755",
        "--statsOutputFile",
        "/home/slarse/Documents/github/work/sorald/sorald-workspace/stats.json"
    ],
    "parseTimeNs": 131803165
}
```
This content is to be considered very much non-final, what I'm pitching with this PR is the approach to gathering statistics rather than the content of the stats file.

In terms of other use of the event handling system, we could for example do logging, nice CLI user feedback, and the like. If statistics collection becomes complicated, we may even want to have multiple event handlers just for that. Thoughts and feedback on this approach very much welcome!


slarse:
I've assigned both @fermadeiral and @khaes-kth as I think you may both want to have a look at this before we merge.

slarse:
I completely forgot to mention, statistics collection kicks in if you add option `--statsOutputFile path/to/some/file.json`

fermadeiral:
> This content is to be considered very much non-final, what I'm pitching with this PR is the approach to gathering statistics rather than the content of the stats file.

I think that the content itself is also already good because we really should start to have something like that as soon as possible.

About the approach, I honestly have never seen it, but I liked it! For me, this PR is ready to merge. I will wait for @khaes-kth to take a look at it.

slarse:
There is one little caveat that I'd like to add: with the current setup, there's no way to access anything but the type of an event from within an event handler without casting the event to its dynamic type. Currently, we have no need to do that, as we only need the events for printing output, and json.org does that with introspection. But, if we do need to access more granular data from events, say the key of a repair event, then there's a need for a slightly different architecture in order to avoid casting left and right.

khaes-kth:
Thank you @slarse . Seems very good to me.

fermadeiral:
> Implementing right now is very low-cost, but we might never need it at all.

If we don't have an usage right now, I think we could let it for the future.

fermadeiral:
Thanks for this feature, @slarse!

PR #236, opened 2020-11-23
fix(rule 2755): Remove handling of setting of FEATURE_SECURE_PROCESSING

### LINKED ISSUE #234: XxeProcessingProcessor does not add checked exception to method signature when using `XFactory.setFeature`
This is is something I overlooked. It appears as if the `setFeature` method always throws a checked exception. The factory creator method must therefore also throw that checked exception.

slarse:
Fix #234 

This PR removes the feature introduced by #219, which was found to be entirely broken (see #234). Unfortunately, #219 wasn't squashed, so simply reverting it was inconvenient.

We will probably want to revisit the topic of checked exceptions later, but I don't think it's a priority until we find the need to work with checked exceptions in multiple processors.


fermadeiral:
> We will probably want to revisit the topic of checked exceptions later, but I don't think it's a priority until we find the need to work with checked exceptions in multiple processors.

Ok then. Thanks for that, @slarse!

Issue #225, opened 2020-11-18
add support for targeted repair


monperrus:
* [x] the miner should be able to output a specific identifier per violation (see #193 )
* [x]  the repair module should be able to take this identifier

monperrus:
@khaes-kth works on point 1 (miner)

slarse:
@khaes-kth You will want to review/merge #221 first :)

slarse:
The idea we discussed at the meeting was the following:

* The miner outputs a JSON file with one entry for each violation, stating:
    - The key of the violation
    - A project-unique ID for the violation (probably something like `<key>:<filepath>:<line_nr>`
* The repair tool accepts violation ID's as a mutually exclusive alternative to rule keys
    - This requires a slight redesign of how the processors initialize resources (which I was considering doing anyway)

slarse:
I'm getting to work on prototyping the targeted repair now. Seems rather easy. @khaes-kth have you started on the miner JSON output thing?

monperrus:
there are actually two modes:

* single violation
* all violations of a single rule

monperrus:
For now, we agree that we do single violation.

slarse:
#269 is merged and good to go, so now it's up to me to integrate that with my work and finish #244 . Should be done tomorrow.

slarse:
This is done as of #257

PR #240, opened 2020-11-24
refactor: Extract discovery of rule violations from SoraldAbstractRepair

slarse:
This is step 1 of the targeted repair implementation (#225).

This refactor extracts the discovery of rule violations from the SoraldAbstractRepair class and pulls it up to an earlier level of the execution. It makes it much easier to replace the rule violations with something constructed from violation IDs for the targeted repair. See further comments in the code.

### DIFF
--- a/src/main/java/sorald/Repair.java
+++ b/src/main/java/sorald/Repair.java
@@ -10,6 +10,7 @@ import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 import org.apache.commons.lang3.tuple.Pair;
@@ -17,6 +18,9 @@ import sorald.processor.SoraldAbstractProcessor;
 import sorald.segment.FirstFitSegmentationAlgorithm;
 import sorald.segment.Node;
 import sorald.segment.SoraldTreeBuilderAlgorithm;
+import sorald.sonar.Checks;
+import sorald.sonar.ProjectScanner;
+import sorald.sonar.RuleViolation;
 import spoon.Launcher;
 import spoon.compiler.Environment;
 import spoon.processing.ProcessingManager;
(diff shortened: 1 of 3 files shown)


src/main/java/sorald/Repair.java
```diff
@@ -65,7 +69,13 @@ public void repair() {
...
             final Path outputDir = inOutPaths.getRight();
 
-            SoraldAbstractProcessor<?> processor = createProcessor(ruleKey);
+            Set<RuleViolation> ruleViolations =
+                    ProjectScanner.scanProject(
+                            inputDir.toFile(),
+                            FileUtils.getClosestDirectory(inputDir.toFile()),
+                            Checks.getCheckInstance(Integer.toString(ruleKey)));
+            SoraldAbstractProcessor<?> processor =
+                    createProcessor(ruleKey).setRuleViolations(ruleViolations);
```
slarse, line 78: We now set rule violations here, at the very top of the repair process. This makes it easy to swap these currently discovered violations for something we construct ourselves from parsing rule IDs for targeted repair.

src/main/java/sorald/processor/SoraldAbstractProcessor.java
```diff
@@ -31,49 +22,8 @@
...
-                try (Stream<Path> walk = Files.walk(Paths.get(node.getRootPath()))) {
-                    filesToScan.addAll(
-                            walk.map(x -> x.toFile().getAbsolutePath())
-                                    .filter(f -> f.endsWith(Constants.JAVA_EXT))
-                                    .collect(Collectors.toList()));
-                } catch (IOException e) {
-                    e.printStackTrace();
-                }
-            }
-        }
```
slarse, line 74: This method is not actually necessary. It's more efficient to scan the entire project for rule violations before splitting it into segments, than to scan the segments one-by-one. It doesn't matter that when repairing segment A there are also violations for segment B in the violations set of the processor, as they won't match anyway.

src/main/java/sorald/Repair.java
fermadeiral: Nice!

src/main/java/sorald/processor/SoraldAbstractProcessor.java
fermadeiral: This code was also duplicate, it's nice that it's gone!

fermadeiral:
Thanks for the refactoring!

PR #244, opened 2020-11-25
wip: Add targeted repair

slarse:
Repair part of #225 

Depends on #240 (which must be merged first). @khaes-kth I've implemented this to use violation IDs on the form `<rule-key>:<filepath>:<line-nr>`. All of that information is available to the miner, so that's the minimum we need to implement targeted repair. The miner is however missing a convenience method in `RuleViolation`, see the code comments.

We can discuss this further at the meeting.

### DIFF
--- a/src/main/java/sorald/sonar/RuleViolation.java
+++ b/src/main/java/sorald/sonar/RuleViolation.java
@@ -14,6 +14,9 @@ public abstract class RuleViolation {
     /** @return The name of the check class that generated this warning. */
     public abstract String getCheckName();
 
+    /** @return The key of the rule that is violated. */
+    public abstract String getRuleKey();
+
     @Override
     public boolean equals(Object obj) {
         if (!(obj instanceof RuleViolation)) {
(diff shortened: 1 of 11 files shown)


src/main/java/sorald/sonar/RuleViolation.java
```diff
@@ -1,28 +1,35 @@
...
-    }
+    public abstract String getFileName();
 
     /** @return The name of the check class that generated this warning. */
-    public String getCheckName() {
-        return message.getCheck().getClass().getSimpleName();
+    public abstract String getCheckName();
+
+    /** @return The key of the rule that is violated. */
+    public abstract String getRuleKey();
```
slarse, line 18: @khaes-kth The miner will most likely need this method. Perhaps I should make a separate PR with this functionality.

slarse:
This is on hold until #245 is fixed, the test failure in CI is related

slarse:
I will replace this with a new PR.

PR #248, opened 2020-11-25
wip: Fix Sorald repairing non-violating elements appearing on the same line as violation

### LINKED ISSUE #245: Sorald can attempt repair on element that does not violate rule if it appears on same line as element that does
The problem is in `SoraldAbstractProcessor::isToBeProcessedAccordingToSonar`. It only checks that a violation exists on the same line as the element. It doesn't check that the violation actually concerns the given processor (i.e. matches the rule key), nor that the element occurs in the correct span of the line.

slarse:
Fix #245 

This is a WIP, current fix is a bit too aggressive. There's the possibility for false positives (as evidenced by failing tests).

I think I need to redesign how processors are invoked in order to get around the problem.

### DIFF
--- a/src/main/java/sorald/FileUtils.java
+++ b/src/main/java/sorald/FileUtils.java
@@ -5,9 +5,13 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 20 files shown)


slarse:
Replaced by #253

Issue #210, opened 2020-11-11
Should Sorald crash if a processor crashes?


slarse:
Currently, if a processor crashes at any point, Sorald crashes. On the one hand, this is good, because if a processor crashes we want to know about it. On the other hand, a single point of failure in a large project could make all of it "unprocessable".

An alternative would be to just log errors, instead of letting the entire thing crash. The downside is that it makes crashes less noticeable.

monperrus:
Per our conversation today, we do not crash, we would count those errors and report them in the JSON file.

PR #250, opened 2020-11-26
feat: Redesign SoraldAbstractProcessor to get more control over single repairs

slarse:
This is related to #245 and #210 

To address those issues, we need more control over single repairs, and in particular, only attempting _one_ repair for each violation. This PR fixes just that.

`SoraldAbstractProcessor` now implements the Spoon methods `process` and `isToBeProcessed`, and they are `final` (meaning that concrete processors are not allowed to override them). Instead, the concrete processors implement `canRepair` and `repair`. These are almost identical to the old `process` and `isToBeProcessed` methods, except that processors no longer need (or should) call `super.process()` and `super.isToBeProcessedAccordingToStandards()`. This reduces code duplication, and also gives `SoraldAbstractProcessor` a lot more control over each repair. This alone allows us to address #210, and it will help in addressing #245 with a general solution (i.e. without fiddling in the individual processors). It also makes implementing new processors less error-prone, as the only two methods that need to be implemented (`canRepair` and `repair`) are abstract, and so cannot be missed.

Another side effect of this is that I found a processor, `UnclosedResourceProcessor`, that actually relied on being able to process each violation multiple times, thus inflating the amount of repairs it claimed to perform. In a test case with 3 violations, it reported 8 repairs. That effectively makes this PR a feature change, as end-user experience is affected, even though the majority is a refactor.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -41,7 +41,12 @@ So the name of your new processor is `CastArithmeticOperandCheck` replacing "Che
 2) Create the processor
 
 Once you have the name for the new processor, you can create a class using that name in `src/main/java/sorald/processor`.
-This new class must extend `SoraldAbstractProcessor` and implement the methods `isToBeProcessed` and `process` (check out real examples of processors [here](/src/main/java/sorald/processor)).
+This new class must extend `SoraldAbstractProcessor` and implement the methods `canRepair` and `repair`.
+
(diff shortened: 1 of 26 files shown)


slarse:
FYI I will be making more changes to how this works, but they require also adapting several of the processors. So I think this PR is a good first step.

fermadeiral:
> So I think this PR is a good first step.

It is. Thanks for it!

PR #253, opened 2020-11-30
fix: Add heuristic element-to-violation matching to prevent incorrect repairs

### LINKED ISSUE #245: Sorald can attempt repair on element that does not violate rule if it appears on same line as element that does

slarse:
Fix #245 

Replaces #248 

Depends on #250 to be merged first

This PR introduces a heuristic for matching Spoon elements to rule violations. This further increases the amount of control we have over repairs, as it decides on all elements to repair for any given rule violation before actually starting to enact repairs. Currently, the heuristic scans the Spoon model once for each rule (as we parse once for each rule), and for each element that the **associated processor says that it can repair**, does the following:

1. Finds all rule violations that intersect that element.
2. Finds all rule violations on the same line as that element.

When the scan is done, the matcher iterates over the rule violations and for each violation picks the first available matching Spoon element, with intersections being prioritized over same-line elements (the same-line elements are essentially a fallback mechanism for when Spoon and Sonar positions don't line up properly). Each Spoon element can only be matched to 1 violation, so the matching is 1:1.

The matching is guaranteed to be the same given the same input, as rule violations are sorted before the scan according to their position and rule type.

When Sorald then goes on to repair violations, it simply checks the best fits mapping for each element to see if there's a matching rule violation, and if there is it repairs that element.


fermadeiral:
This is really great, thanks @slarse!

PR #254, opened 2020-11-30
feat: Add in-place repair mode

### LINKED ISSUE #190: Add in-place repair mode
I just tried running Sorald on Spoon, and what annoyed me immediately was that I couldn't get it to do the repairs in-place. I had to copy the repaired files from the workspace into the Spoon project. As there's always a directory called "spooned" that's generated in the workspace, the semi-intuitive workaround of specifying the source directory as the workspace also doesn't work.

It would be very useful to just be able to run repairs in-place, because then inspecting the repairs made is a simple `git diff` away.

slarse:
Fix #190 

This PR adds an in-place repair mode, which writes output directly to the source files. It's literally in-place; if Sorald processes multiple rules the source files will be written to multiple times.

Can be activated by passing `--fileOutputStrategy IN_PLACE` to the `repair` command.

### DIFF
--- a/src/main/java/sorald/Repair.java
+++ b/src/main/java/sorald/Repair.java
@@ -236,7 +236,7 @@ public class Repair {
         try {
             Files.writeString(filepath, output);
         } catch (IOException e) {
-            // most convert to a runtime exception as this is used in writeModel, which in turn
+            // must convert to a runtime exception as this is used in writeModel, which in turn
             // is used in a stream (that can't have checked exceptions)
             throw new RuntimeException(e);
         }
--- a/src/main/java/sorald/Repair.java
+++ b/src/main/java/sorald/Repair.java
@@ -222,17 +222,23 @@ public class Repair {
                                     .getEnvironment()
                                     .createPrettyPrinter()
                                     .printTypes(type);
-                    try {
-                        Files.writeString(Paths.get(patchedFile.getKey()), output);
-                    } catch (IOException e) {
-                        e.printStackTrace();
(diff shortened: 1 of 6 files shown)


src/main/java/sorald/Repair.java
```diff
@@ -196,22 +202,46 @@ private void writeModel(CtModel model, Path outputDir) {
...
         }
     }
 
+    private static void writeString(Path filepath, String output) {
+        try {
+            Files.writeString(filepath, output);
+        } catch (IOException e) {
+            // most convert to a runtime exception as this is used in writeModel, which in turn
+            // is used in a stream (that can't have checked exceptions)
+            throw new RuntimeException(e);
```
slarse, line 241: Coverage drop is due to this case never being hit. We must convert the IOException to a RuntimeException (or any derivative of that) as this method is indirectly used in a stream. Streams cannot have checked exceptions thrown in them as they are lazy, and therefore can be evaluated somewhere other than where they are defined.

nharrand:
LGTM! (Plus I would like to use it now, so if no one objects I'll merge this ;-) )

nharrand:
Thanks for the feature @slarse !

slarse:
@nharrand You're welcome! I also really needed this as it just makes it so much more convenient to try out Sorald on real projects. In-place repair and the `git diff`, shows me exactly what changed with minimal effort.

khaes-kth:
Thanks @nharrand and @slarse.
Very useful feature.

PR #257, opened 2020-12-02
feat: Add targeted repair feature

slarse:
#225 

Depends on #253 being merged first

Adds the repair tool part of targeted repair. Currently works by specifying specific repairs on the command line with `--violationSpecs`, where each specifier is on the form `<rulekey>:<filepath>:<startline>:<startcol>:<endline>:<endcol>`. These specifiers can be generated with the `mine` command by adding the `--statsOutputFile` argument.

~I'm thinking that in a large amount of cases, simply specifying the line of the violation should be sufficient, i.e. `<rulekey>:<filepath>:<startline>`, and omitting the rest of the position. As long as there is no other violation of the same kind on that line, the heuristic matcher will find the correct violation.~ This is actually not a good idea, as elements of the same kind can then be mixed up (if just one of them is a violation).


slarse:
Violation id filepath _must_ be relative to `--originalFilesPath`

src/main/java/sorald/cli/Cli.java
```diff
@@ -69,25 +71,32 @@ public Integer call() {
...
+                    description =
+                            "One or more rule violation specifiers. Specifiers can be gathered "
+                                    + "with the '"
+                                    + Constants.MINE_COMMAND_NAME
+                                    + "' command using the "
+                                    + Constants.ARG_STATS_OUTPUT_FILE
+                                    + " option.",
+                    required = true,
+                    split = ",")
+            List<String> ruleViolationSpecifiers;
```
slarse: FYI this makes the options ARG_RULE_KEYS and ARG_RULE_VIOLATION_SPECIFIERS mutually exclusive in the CLI

src/main/java/sorald/cli/Cli.java
```diff
@@ -69,25 +71,32 @@ public Integer call() {
...
+                    description =
+                            "One or more rule violation specifiers. Specifiers can be gathered "
+                                    + "with the '"
+                                    + Constants.MINE_COMMAND_NAME
+                                    + "' command using the "
+                                    + Constants.ARG_STATS_OUTPUT_FILE
+                                    + " option.",
+                    required = true,
+                    split = ",")
+            List<String> ruleViolationSpecifiers = List.of();
```
slarse, line 99: FYI this makes the options ARG_RULE_KEYS and ARG_RULE_VIOLATION_SPECIFIERS mutually exclusive in the CLI

src/main/java/sorald/cli/Cli.java
```diff
@@ -171,11 +181,69 @@ private void validateArgs() {
...
+                        String.format(
+                                "Invalid violation ID '%s', no file '%s' in directory '%s'",
+                                violationSpecifier, rawFilename, originalFilesPath));
+            }
+
+            int startLine = Integer.parseInt(parts[2]);
+            int startCol = Integer.parseInt(parts[3]);
+            int endLine = Integer.parseInt(parts[4]);
+            int endCol = Integer.parseInt(parts[5]);
+            return new SpecifiedViolation(key, fileName, startLine, endLine, startCol, endCol);
```
slarse, line 240: I intend to refactor all of this specifier parsing into a separate class, but that'll be for a future PR.

src/main/java/sorald/event/models/WarningLocation.java
```diff
@@ -11,25 +12,13 @@
...
-                        + filePath
-                        + File.pathSeparator
-                        + startLine
-                        + File.pathSeparator
-                        + endLine
-                        + File.pathSeparator
-                        + startColumn
-                        + File.pathSeparator
-                        + endColumn;
+        this.violationSpecifier = violation.relativeSpecifier(projectPath);
```
slarse, line 21: This change wasn't strictly necessary, but it reduces code duplication and also fixes the problem of filepaths in mined specifiers being absolute (they are now relative to the project path).

src/main/java/sorald/cli/Cli.java
slarse: It can currently crash not-so-gracefully on poorly formatted specifiers (e.g. incorrect amount of elements in the specifier, numbers not actually being numbers, etc)

slarse:
@khaes-kth I think you're the best qualified to review this as you've just recently worked on the related mining feature. Note that this PR also fixes the problem we discussed about filepaths in specifiers being absolute. It's somewhat unrelated to the purpose of this PR, so I could remove it, but we're in a bit of a hurry AFAIK and it reduced duplication in specifier generation.

slarse:
@khaes-kth just fixed the merge conflict in `Consants.java` :)

src/main/java/sorald/event/models/WarningLocation.java
khaes-kth: Good. I just thought, logically speaking, we should be able to create a WarningLocation without specifying the project path as specified in the command line. But we can add a new constructor later whenever a new way of initialization of WarningLocation was required.

khaes-kth:
Awesome PR. Thanks @slarse

monperrus:
Thanks @slarse

slarse:
:)

PR #259, opened 2020-12-03
refactor: Remove Constants.ARG_SYMBOL

### LINKED ISSUE #242: Remove the Constants.ARG_SYMBOL constant
It was necessary as the JSAP library needed the arguments without argument symbol in order to extract data. This is not the case with picocli, and so we can remove the `Constants.ARG_SYMBOL` and incorporate it into the argument names. The reason we will want to do this is that it's incredibly annoying to write test code with invocations to Sorald, as you need to type e.g. `Constants.ARG_SYMBOL + Constants.SOME_ARGUMENT` for every single argument. A lot of clutter.

slarse:
Fix #242 

This is a simple refactor that removes the `Constants.ARG_SYMBOL` constant and puts `--` directly in each argument name. The split was necessary when using the old CLI library, as it required the argument names _without_ the `--` prefix to extract options. picocli doesn't, and so the split is purely an annoyance when writing tests.

### DIFF
--- a/src/main/java/sorald/Constants.java
+++ b/src/main/java/sorald/Constants.java
@@ -11,21 +11,20 @@ public class Constants {
     public static final String REPAIR_COMMAND_NAME = "repair";
     public static final String MINE_COMMAND_NAME = "mine";
 
-    public static final String ARG_SYMBOL = "--";
-    public static final String ARG_RULE_KEYS = "ruleKeys";
-    public static final String ARG_ORIGINAL_FILES_PATH = "originalFilesPath";
-    public static final String ARG_STATS_ON_GIT_REPOS = "statsOnGitRepos";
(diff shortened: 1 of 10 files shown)


khaes-kth:
Thanks @slarse

khaes-kth:
@slarse Do you know what causes this failure?

[https://github.com/SpoonLabs/sorald/runs/1498576225](https://github.com/SpoonLabs/sorald/runs/1498576225)

slarse:
@khaes-kth No, but just from looking at the output I'm thinking it might be a race condition in the printing to stdout. I'll redesign the failing test to not look at stdout. That's silly.

khaes-kth:
Great! Will you create a new PR with that modification?

slarse:
Yes. I think I'll have time to fix it in between meetings today. It shouldn't be hard. See #264

PR #262, opened 2020-12-04
fix: Re-implement segment mode's crash isolation

### LINKED ISSUE #258: Segment mode no longer isolates crashes within segments
Segment mode is supposed to ignore crashes within individual segments, and keep on processing the rest of the segments. After several refactorings over the past month, this no longer works as intended, and a crash will simply cause everything to fail.

This happened because:

1. I misunderstood the purpose of segment mode (thought it was for OOM errors only).
2. The behavior was not tested.

So, we need to reimplement the behavior (not hard) and test it (slightly harder).

slarse:
Fix #258 

This PR re-implements segment mode's intended behavior of ignoring crashes within individual segments, which I accidentally removed when redesigning the architecture. There's now also a test that verifies that segment strategy can actually withstand crashes in individual segments.

This still runs Sonar on the whole project before starting segment mode, and so it's possible that we'll need to move that back into the segments again. However, Sonar is very stable, so I don't think it should be an issue.


slarse:
@fermadeiral I was considering tagging Henry in this PR, but I don't want to distract him now that he's just about to present his thesis. You can determine if you think he needs to be in on this PR. I've discussed the changes with him in the related issue, so he's on board with the idea of the changes.

slarse:
Fixed merge conflict

fermadeiral:
> @fermadeiral I was considering tagging Henry in this PR, but I don't want to distract him now that he's just about to present his thesis. You can determine if you think he needs to be in on this PR.

Thank you for pointing this out. I would prefer that Henry indeed sees this PR, after his defense. Could we put it on hold? I will tag him by Monday.

slarse:
Sure, that's fine. It's relatively isolated from the rest of the things we're doing right now.

slarse:
@fermadeiral hope you don't mind that I tag Henry now, we need this merged.

@henry-lp could you have a look at this PR and see if there's anything you'd like changed/added?

slarse:
The reason I think this is somewhat urgent is that, given Henry's presentation, I don't think it's feasible to run the experiments for the paper in default mode. It will just crash too often.

henry-lp:
@slarse 
It looks great :) . I think I'll work now when you encapsulate the model building part a try-catch.

slarse:
@henry-lp Great, thanks! I also think it should work, at least the test case passes :).

@fermadeiral Feel free to merge if you're happy with this PR!

fermadeiral:
Great job! Merging now.

monperrus:
Excellent, thanks all.

PR #269, opened 2020-12-07
feat: generates json output for mining

khaes-kth:
Adds feature of generating json output for mining. The output would include "usableLocation" for warnings which will be used in a future PR to address #225.

Generally, it follows @slarse suggestion for #232.

Still needs testing. I have not tested it yet, because of some issues with running it from my IDE on windows. I will test and finalize it very soon.


src/main/java/sorald/FileUtils.java
```diff
@@ -114,6 +117,27 @@ public static void writeStatisticsJSON(
...
 
+    /**
+     * Write the statistics JSON file.
+     *
+     * @param statsOutputFile The file to write to.
+     * @param statsCollector A {@link SoraldEventHandler} containing stats.
+     * @param executionInfo The information of the execution.
+     * @throws IOException If the file can't be written to.
+     */
+    public static void writeStatisticsJSON(
```
khaes-kth, line 128: @slarse  Do you think we can merge this function and the old `writeStatisticsJSON` (which was used for repair mode)? The old one receives a `StatisticsCollector` instead of `SoraldEventHandler`.

slarse: @khaes-kth It could just receive `Object`, really. The method itself doesn't need to know anything about the object it receives.

slarse: @khaes-kth  I'll open up a new PR with a more generalized version of the `writeStatisticsJSON` method, such that we can both have it.

khaes-kth:
@slarse a sample output is this:

```
{
    "totalMiningTime": 1074,
    "executionInfo": {
        "soraldVersion": "92d377cc39f63be35f9bee00d372c8f1ca06be5e",
        "javaVersion": "13.0.1",
        "originalArgs": [
            "mine",
            "--originalFilesPath",
            "..\\tmp\\test\\hello-world",
            "--statsOutputFile",
            "..\\tmp\\test\\output.txt",
            "--tempDir",
            "..\\tmp\\test\\tmp",
            "--target",
            "hello-world"
        ],
        "target": "hello-world"
    },
    "miningStartTime": "Mon Dec 07 15:11:55 CET 2020",
    "minedRules": [{
        "warningLocations": [{
            "endLine": 6,
            "endColumn": 24,
            "startColumn": 22,
            "startLine": 6,
            "filePath": "C\\other\\daneshgah\\phd-kth\\projects\\sorald\\sorald\\..\\tmp\\test\\hello-world\\HelloWorld.java",
            "usableLocation": "4973###C\\other\\daneshgah\\phd-kth\\projects\\sorald\\sorald\\..\\tmp\\test\\hello-world\\HelloWorld.java###6###6###22###24"
        }],
        "nbFoundWarnings": 1,
        "ruleName": "CompareStringsBoxedTypesWithEqualsCheck",
        "ruleKey": "4973"
    }],
    "miningEndTime": "Mon Dec 07 15:11:56 CET 2020"
}
```

A couple of points to have in mind: for the `soraldVersion` attribute, I assumed that we will set it in the [Constants ](https://github.com/khaes-kth/sorald/blob/issue/225-generate-mining-output/src/main/java/sorald/Constants.java). I did not include the `command` attribute, because the `originalArgs` already gives us the command.

Most importantly, the `usableLocation` has this format: `<ruke-key>###<file-path>###<line-start>###<line-end>###<column-start>###<column-end>`. As you see in the sample, the `file-path` is an absolute path for now. If you think it should be changed to a relative path in this PR, give me a hint how to do that and I will fix it.

khaes-kth:
One more question and one note: question: how do you usually apply `spotless`? Do you use `mvn spotless:apply` every time? Or have you set up an automatic 'spotless applier'? note: There is an old issue that some tests do not pass locally on windows #273. I will fix this.

src/main/java/sorald/Constants.java
```diff
@@ -46,12 +49,15 @@
...
 
     public static final String STRING_QUALIFIED_NAME = "java.lang.String";
+    public static final String JAVA_VERSION_SYSTEM_PROPERTY = "java.version";
 
     public static final String TOSTRING_METHOD_NAME = "toString";
     public static final String HASHCODE_METHOD_NAME = "hashCode";
 
     public static final String PATH_TO_STATS_OUTPUT = "experimentation/stats/warnings";
 
+    public static final String VIOLATION_USABLE_PATH_SEPARATOR = "###";
```
slarse, line 59: This is not a good separator as hash symbols can be part of filepaths. It's also less convenient to split a path on multiple characters.

As we have ambitions to support both Windows and Unix-likes, we should use `File.pathSeparator` as the separator (`:` on most Unix-likes and `;` on Windows).

src/main/java/sorald/event/models/WarningLocation.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public Integer getStartColumn() {
+        return startColumn;
+    }
+
+    public Integer getEndColumn() {
+        return endColumn;
+    }
+
+    public String getUsableLocation() {
```
slarse, line 55: Could this be called something like `getViolationId` or `getViolationSpecifier` instead? It's not only a location, it defines an entire violation, rule key and all.

src/main/java/sorald/event/collectors/MinerStatisticsCollector.java
```diff
@@ -0,0 +1,63 @@
...
+                Optional<MinedRuleEvent> existingEvent =
+                        minedRules.stream()
+                                .filter(x -> x.getRuleKey().equals(minedRuleEvent.getRuleKey()))
+                                .findFirst();
+
+                if (!existingEvent.isEmpty()) {
+                    existingEvent.get().addWarningLocations(minedRuleEvent.getWarningLocations());
+                } else {
+                    minedRules.add(minedRuleEvent);
+                }
```
slarse, line 38: This I would advice against. Events should be immutable. If another event handler has also gotten this particular event, its caught event would be mutate as well, which would be a very high-risk side effect.

My advice: make `MinedRuleEvent` immutable and have it only contain which event was mined. Then keep track of the amounts of events in an internal data structure in this event handler.

src/main/java/sorald/FileUtils.java
```diff
@@ -105,7 +106,24 @@ public static void writeJSON(File file, Object coreObj, Map<String, Object> addi
...
+                                        (e.getValue() instanceof List
+                                                ? ((List) e.getValue()).toArray()
+                                                : e.getValue())))
+                .forEach(
+                        (k, v) ->
+                                jo.put(
+                                        k,
+                                        (v.getClass().isArray()
+                                                ? new JSONArray(v)
+                                                : new JSONObject(v))));
```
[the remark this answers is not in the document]
slarse: If it turns out to be necessary, prefer `map` over `collect` as an intermediate operation in a stream. `collect` terminates the stream and puts the results in a collection, so each `collect` you do will cause another iteration over all items. It won't matter much here as there are only a few items, but in general that's the way to go.

slarse (changes requested):
Looks really nice, I just have some minor notes and one major note on the mutable events.

slarse:
@khaes-kth 

> One more question and one note: question: how do you usually apply `spotless`? Do you use `mvn spotless:apply` every time? Or have you set up an automatic 'spotless applier'?

I have a pre-commit hook that runs spotless for me. I've also got Google's formatter (which is what we use in spotless) plugged in to my IDE. My pre-commit hook looks like so (although if you use windows it won't work as it's a bash script):

```bash
#! /bin/bash

mvn spotless:check
if [ $? != 0 ]; then
    mvn spotless:apply
    exit 1
fi

mvn clean compile
```

slarse:
> A couple of points to have in mind: for the `soraldVersion` attribute, I assumed that we will set it in the [Constants ](https://github.com/khaes-kth/sorald/blob/issue/225-generate-mining-output/src/main/java/sorald/Constants.java).

The version will be dynamically resolved, but we can have it as a constant for now until I get around to implement the version provider.

> Most importantly, the `usableLocation` has this format: `<ruke-key>###<file-path>###<line-start>###<line-end>###<column-start>###<column-end>`. As you see in the sample, the `file-path` is an absolute path for now. If you think it should be changed to a relative path in this PR, give me a hint how to do that and I will fix it.

Ideally the filepath would be relative to make the violation IDs portable, but we can fix that in an upcoming PR. No need to do everything at once :)

src/main/java/sorald/FileUtils.java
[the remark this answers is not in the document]
khaes-kth: Good, I agree with your point about `collect` vs `map` and made a change accordingly.

Regarding converting to json array, I know that it is required for json object (ex., it did not work without conversion for ExecutionInfo) but I think it works for array without conversion. Let's keep the conversion for array as well to be consistent.

src/main/java/sorald/event/collectors/MinerStatisticsCollector.java
```diff
@@ -0,0 +1,63 @@
...
+                break;
+            case MINING_END:
+                miningEndTime = System.currentTimeMillis();
+                break;
+            case MINED:
+                MinedRuleEvent minedRuleEvent = (MinedRuleEvent) event;
+                Optional<MinedRuleEvent> existingEvent =
+                        minedRules.stream()
+                                .filter(x -> x.getRuleKey().equals(minedRuleEvent.getRuleKey()))
+                                .findFirst();
```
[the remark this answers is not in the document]
khaes-kth: Yeah, I think some architectural changes can be made to improve it. We should think what kind of change is most appropriate.

src/main/java/sorald/Constants.java
khaes-kth: Done!

src/main/java/sorald/event/models/WarningLocation.java
khaes-kth: Done!

src/main/java/sorald/event/collectors/MinerStatisticsCollector.java
khaes-kth: Done! (have a look to see if it needs further fixes)

khaes-kth: Actually, this one is also related to the event-handler architecture. Maybe we can make it simpler by changing the architecture.

khaes-kth:
> I've also got Google's formatter (which is what we use in spotless) plugged in to my IDE

Nice, it works well.

src/main/java/sorald/FileUtils.java
```diff
@@ -105,7 +107,24 @@ public static void writeJSON(File file, Object coreObj, Map<String, Object> addi
...
+                                        (e.getValue() instanceof List
+                                                ? ((List) e.getValue()).toArray()
+                                                : e.getValue())))
+                .forEach(
+                        e ->
+                                jo.put(
+                                        (String) e.getKey(),
+                                        (e.getValue().getClass().isArray()
+                                                ? new JSONArray(e.getValue())
+                                                : new JSONObject(e.getValue()))));
```
slarse, line 126: This is just a nit-pick, but the reason you need to cast to `String` here is because you're using a raw collection. I also realized there's a bug: you should check for an instance of a `Collection`, not `List` (it could be e.g. a set or something like that as well, which should be a JSONArray).

```suggestion
        additionalData.entrySet().stream()
                .map(
                        e ->
                                new AbstractMap.SimpleEntry<>(
                                        e.getKey(),
                                        (e.getValue() instanceof Collection
                                                ? ((Collection<?>) e.getValue()).toArray()
                                                : e.getValue())))
                .forEach(
                        e ->
                                jo.put(
                                        e.getKey(),
                                        (e.getValue().getClass().isArray()
                                                ? new JSONArray(e.getValue())
                                                : new JSONObject(e.getValue()))));
```

Raw types really should be avoided if at all possible, they force casting too much.

Anyway, now that I'm looking at this, I think it would be much nicer to just define a helper method like so:

```java
    /**
     * Convert any collection or array into a JSONArray, and anything else into a JSONObject.
     */
    private static Object toJSONArrayOrObject(Object obj) {
        if (obj instanceof Collection) {
            return new JSONArray(((Collection<?>) obj).toArray());
        } else if (obj.getClass().isArray()) {
            return new JSONArray(obj);
        } else {
            return new JSONObject(obj);
        }
    }
```

And then make the whole thing be this:
```suggestion
        additionalData.forEach((k, v) -> jo.put(k, toJSONArrayOrObject(v));
```

Streams often get a bit unreadable if you define everything inline, so helpers methods can make it a lot more clear what's going on. But you decide what you wanna do here, I just find discussing this stuff to be really fun :)

src/main/java/sorald/cli/Cli.java
```diff
@@ -228,16 +237,42 @@ private SoraldConfig createConfig() {
...
+
+            if (statsOutputFile != null) {
+                Map<String, Object> additionalStatData = new HashMap<>();
+                additionalStatData.put(
+                        StatsMetadataKeys.EXECUTION_INFO,
+                        new ExecutionInfo(
+                                spec.commandLine().getParseResult().originalArgs(),
+                                Constants.SORALD_VERSION,
+                                javaVersion,
+                                target));
```
slarse, line 271: Just a tip, as of Java ... 9 I think, you can do stuff like this with `Map.of`

```suggestion
                Map<String, Object> additionalStatData = Map.of(
                        StatsMetadataKeys.EXECUTION_INFO,
                        new ExecutionInfo(
                                spec.commandLine().getParseResult().originalArgs(),
                                Constants.SORALD_VERSION,
                                javaVersion,
                                target));
```

That's nothing that you need to change, just something that you might not be aware of if you (like most people) have been stuck on Java 8 for an eternity. There's also `List.of` and `Set.of`.

src/main/java/sorald/event/collectors/MinerStatisticsCollector.java
```diff
@@ -0,0 +1,76 @@
...
+                                new MinedRule(
+                                        e.getKey().split(RULE_ID_SEPARATOR)[0],
+                                        e.getKey().split(RULE_ID_SEPARATOR)[1],
+                                        ImmutableList.copyOf(e.getValue())))
+                .collect(Collectors.toList());
+    }
+
+    private String violationToRuleId(MinedViolationEvent violation) {
+        return violation.getRuleKey() + RULE_ID_SEPARATOR + violation.getRuleName();
+    }
```
slarse, line 75: Is this only used as a key in the `ruleToViolations` map? If so, the rule key alone is just as unique as the rule key + rule name.

src/main/java/sorald/event/collectors/MinerStatisticsCollector.java
```diff
@@ -0,0 +1,76 @@
...
+            case MINED:
+                MinedViolationEvent minedViolationEvent = (MinedViolationEvent) event;
+
+                if (!ruleToViolations.containsKey(violationToRuleId(minedViolationEvent)))
+                    ruleToViolations.put(violationToRuleId(minedViolationEvent), new ArrayList<>());
+
+                ruleToViolations
+                        .get(violationToRuleId(minedViolationEvent))
+                        .add(minedViolationEvent.getWarningLocation());
+                break;
```
slarse, line 42: Exactly the way I was thinking, this is much clearer.

src/main/java/sorald/event/models/miner/MinedRule.java
```diff
@@ -0,0 +1,29 @@
...
+    private final String ruleKey;
+    private final String ruleName;
+    private final ImmutableList<WarningLocation> warningLocations;
+
+    public MinedRule(
+            String ruleKey, String ruleName, ImmutableList<WarningLocation> warningLocations) {
+        this.ruleKey = ruleKey;
+        this.ruleName = ruleName;
+        this.warningLocations = warningLocations;
+    }
```
slarse, line 16: Typically, what I do when I want to have immutable lists is to use standard library stuff. The input doesn't need to be immutable, either, it can be any list.

```suggestion
    public MinedRule(
            String ruleKey, String ruleName, List<WarningLocation> warningLocations) {
        this.ruleKey = ruleKey;
        this.ruleName = ruleName;
        this.warningLocations = Collections.unmodifiableList(warningLocations);
    }
```

src/main/java/sorald/event/models/miner/MinedRule.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public ImmutableList<WarningLocation> getWarningLocations() {
+        return warningLocations;
+    }
```
slarse, line 20: Alternatively, one can have the internal list be mutable, and then present an immutable copy of it.

```suggestion
   /** @returns An immutable list of warning locations. */
    public List<WarningLocation> getWarningLocations() {
        return Collections.unmodifiableList(warningLocations);
    }
```

slarse (approved):
This looks fine to me now. I have some minor suggestions here and there, and noted a minor bug that you might want to fix in `writeJSON`, but overall I think it looks great! Ping me when you are happy and I'll merge.

src/main/java/sorald/event/collectors/MinerStatisticsCollector.java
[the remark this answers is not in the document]
slarse: The only real choice we have is to have dedicated event methods for the complex events. It is the only way to avoid casting.

src/main/java/sorald/event/models/miner/MinedRule.java
slarse: Caveat: `Collections.unmodifiableList` provides a read-only view of a list, so if you want to make sure the list isn't modified elsewhere, it also needs to be copied. E.g. `Collections.unmodifiableList(new ArrayList<>(warningLocations))`

slarse:
I'm not sure what's going on with codecov not putting checks on PRs anymore ... will have to look into that. The coverage report looks fine, though.

src/main/java/sorald/event/collectors/MinerStatisticsCollector.java
slarse: I don't think it is related to the architecture. Only the casting is. I mean, we could send a _copy_ of each event to each handler, but that would entail having to make a copy constructor for each event as well, so it would be much easier to just keep to immutable events.

I mean, what would be architectural is to actually enforce immutable events. I don't think there's any way to enforce immutability natively in Java. We could make a Spoon processor to check it. Any ideas?

khaes-kth:
Thanks for your great suggestions @slarse. I applied almost all of them, except the one related to `ImmutableList` because I think we have some disagreements there. We can fix that one in a refactoring PR later.

So, I believe this PR should be ready to be merged.

slarse:
@khaes-kth The `ImmutableList` thing doesn't need to change, it's fine as is. I just personally prefer to use the standard library where possible, instead of external dependencies.

Right, then I'm merging, looks great! I'll integrate this with my own work on targeted repair, and hopefully I'll have a review-ready before the meeting tomorrow.

khaes-kth:
> hopefully I'll have a review-ready before the meeting tomorrow.

Great. Looking forward to it!

slarse:
By the way, reviewing this PR was a really nice change of pace for me. It's also important for my personal development to _read_ code, and not just write it, so I really hope you'll have time to do more dev work on Sorald in the near future.

khaes-kth:
It was joyful for me as well. Especially, because I'm an "old-school" developer who is not used to take advantage of all new features of Java 8+. I learned something and look forward to doing it again :)

PR #271, opened 2020-12-07
fix: Make main method exit non-zero instead of throwing

slarse:
The `Main.main` method currently throws an `IllegalStateException` when there's an error in executing a command. That's super strange, and it looks very odd to a user to see `IllegalStateException` written out in red letters when they mistype a CLI option, or something like that.

This PR fixes that by instead simply exiting non-zero. To make this behavior testable, it adds a custom security manager that throws a particular exception on non-zero exits. The security manager is registered before each test is executed, and unregistered after each test.

### DIFF
--- a/src/main/java/sorald/Main.java
+++ b/src/main/java/sorald/Main.java
@@ -3,13 +3,10 @@ package sorald;
 import sorald.cli.Cli;
 
 public class Main {
-    public static void main(String[] args) throws Exception {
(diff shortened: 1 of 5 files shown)


slarse:
It appears as if Jacoco can't measure a hit on a `System.exit` call. That line is absolutely being hit.

monperrus:
Interesting, good to know.

Issue #246, opened 2020-11-25
add a JSON Schema for the JSON output


monperrus:
@khaes-kth @fermadeiral will meet to create first alpha version

monperrus:
We would need two JSON schemas one for miner and one for repair

slarse:
This is related to #225. To be able to do targeted repair, I need the miner schema to contain at least the following for each rule violation:

* The key of the rule.
* Path to the file
    - For portability, should probably be relative to `--originalFilesPath`, i.e. the project root.
* The precise position of the violation (start line, end line, start column, end column)

fermadeiral:
> The precise position of the violation (start line, end line, start column, end column)

If I recall correctly, Sonar Java returns only the file name and the line number of the violation. Was that changed in newer versions?

slarse:
> If I recall correctly, Sonar Java returns only the file name and the line number of the violation. Was that changed in newer versions?

The exact location is semi-hidden in a nested data structure in the `AnalyzerMessage`, see https://github.com/SpoonLabs/sorald/blob/7e01bb9816f7b750464a821716a818ac148c761f/src/main/java/sorald/sonar/ScannedViolation.java#L13-L31

khaes-kth:
@fermadeiral and I prepared a first version of the json schema. The first link below is the schema/example for mining output and the second one is for repair output (including small changes that I made after we made a first draft with Fernanda). They are very close to what @slarse had proposed before.

[https://www.jsonschemavalidator.net/s/Glo83TcP](https://www.jsonschemavalidator.net/s/Glo83TcP)
[https://www.jsonschemavalidator.net/s/afS0RbsJ](https://www.jsonschemavalidator.net/s/afS0RbsJ)

To have an idea with a quick look, have a look at the examples (right hand side).

slarse:
@khaes-kth I've got a question on the mining schema, what's "miningTime" here?

```
            "rules": [
                {
                    "ruleNumber": "2111",
                    "ruleName": "BigDecimalDoubleConstructor",
                    "nbWarningsFound": 1,
                    "warnings": {
                        "location": {
                            "filePath": "relative/path/to/file",
                            "beginningLineNumber": 10,
                            "endLineNumber": 12,
                            "beginningColumnNumber": 3,
                            "endColumnNumber": 1
                        }
                    },
                    "miningTime": 1323
                }
            ],
```
[...]

PR #282, opened 2020-12-10
feat: Automatically resolve version info

slarse:
Related to #246 

This PR adds dynamic resolution of version info. The version number is only used if it is a release version (i.e. non-snapshot). Otherwise, falls back on the commit hash (shortened to 8 characters for readability and prefixed with `commit: ` to make it clear that it's a commit hash).

The version and commit hash are added to the MANIFEST.MF file in the jar, which is configured in `pom.xml`

What's also new here is that `sorald --version` now prints the version number, which it did not previously.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,6 +145,24 @@
                 </executions>
             </plugin>
             <plugin>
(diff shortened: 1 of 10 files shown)


khaes-kth:
Thanks @slarse

Issue #284, opened 2020-12-10
Terminology: warning or rule violation, which one?


slarse:
Throughout Sorald, we use the terms "warning" and "rule violation" interchangeably. To reduce confusion, we should pick one. I have no preference, I just want consistency. I can't find that there's any official terminology from SonarQube.

"Warning" is shorter than "rule violation", should we go with that? What terms do you use in the current paper?

@nharrand @monperrus @khaes-kth @fermadeiral

slarse:
There is actually a slight difference: a warning points out a violation of a rule. A false positive is a warning without violation, and false negative is a violation without a warning.

As such, I think the correct terminology for us would be "warning", as we are not concerned with false positives/negatives in Sorald.

fermadeiral:
> As such, I think the correct terminology for us would be "warning", as we are not concerned with false positives/negatives in Sorald.

I agree.

khaes-kth:
> What terms do you use in the current paper?

We are using "rule violation" but I think you're right about the tool, it's better to use warning there.

monperrus:
Both are correct and have pros and cons.

I would prefer "rule violation" because it's more precise, more explicit, and cannot be confused 
with compiler warnings, performance warnings, cost warnings, etc.

slarse:
Since you are using "rule violation" for the paper, I think we should stick to that in the tool as well. I will start gravitating toward that terminology.

khaes-kth:
I'm fine with both of them and actually, Martin made a good point. I just want to mention the terminology in the paper should not be the only reason for our choice, we may change it a hundred times before submitting the paper :-)

slarse:
I also thought @monperrus made a good point. "Warning" is pretty non-descriptive.

slarse:
I'm gonna close this now. We gravitate toward "violation" rather than "warning" as it's more specific.

PR #285, opened 2020-12-10
feat: Add more output to repair statistics

### LINKED ISSUE #246: add a JSON Schema for the JSON output

slarse:
Fix #246 

This turned out to be more work than I thought. The granularity of the data listed in the schema is high, and it was non-trivial to get put it together.

Still working on it, but I'll be done tomorrow morning.

### DIFF
--- a/src/main/java/sorald/cli/Cli.java
+++ b/src/main/java/sorald/cli/Cli.java
@@ -206,7 +206,7 @@ public class Cli {
                             SoraldVersionProvider.getVersionFromPropertiesResource(
                                     SoraldVersionProvider.DEFAULT_RESOURCE_NAME),
                             javaVersion,
-                            "N/A");
+                            target);
 
             List<RuleRepairStatistics> repairStats =
                     RuleRepairStatistics.createRepairStatsList(statsCollector, projectPath);
--- a/src/main/java/sorald/cli/Cli.java
+++ b/src/main/java/sorald/cli/Cli.java
@@ -195,7 +195,7 @@ public class Cli {
          */
         private void mineWarningsAfter(Repair repair, List<Integer> ruleKeys) {
             File projectPath = originalFilesPath.toPath().toAbsolutePath().normalize().toFile();
-            ruleKeys.forEach(key -> repair.mineWarnings(projectPath, key));
+            ruleKeys.forEach(key -> repair.mineViolations(projectPath, key));
         }
 
         private void writeStatisticsOutput(StatisticsCollector statsCollector, Path projectPath)
--- a/src/main/java/sorald/cli/Cli.java
+++ b/src/main/java/sorald/cli/Cli.java
@@ -67,7 +67,7 @@ public class Cli {
                 names = Constants.ARG_TARGET,
                 description =
                         "The target of this execution (ex. sorald/92d377). This will be included in the json report.")
-        String target = "N/A";
+        String target;
 
         @CommandLine.Option(
                 names = Constants.ARG_STATS_OUTPUT_FILE,
--- a/src/main/java/sorald/cli/Cli.java
+++ b/src/main/java/sorald/cli/Cli.java
@@ -58,16 +58,32 @@ public class Cli {
         }
     }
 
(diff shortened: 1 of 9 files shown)


slarse:
I suspect the failure is due to an overflow somewhere, but I can't find that I'm using an `int` anywhere ...

src/main/java/sorald/cli/Cli.java
```diff
@@ -238,7 +290,7 @@ private RuleViolation parseRuleViolation(String violationSpecifier) {
             int startCol = Integer.parseInt(parts[3]);
             int endLine = Integer.parseInt(parts[4]);
             int endCol = Integer.parseInt(parts[5]);
-            return new SpecifiedViolation(key, fileName, startLine, endLine, startCol, endCol);
+            return new SpecifiedViolation(key, fileName, startLine, startCol, endLine, endCol);
```
slarse, line 293: Targeted repair was apparently entirely broken. The stats collector tests failed miserably before I found this error.

slarse:
@khaes-kth I think this is ready for you now. It turned out to be quite hard to give the highly granular statistics that were requested, but I think I've gotten the most crucial parts.

Here is sample output from a repair:

```json
{
    "repairs": [{
        "nbPerformedRepairs": 1,
        "ruleName": "XxeProcessing",
        "crashedRepairsLocations": [],
        "nbCrashedRepairs": 0,
        "ruleKey": "2755",
        "performedRepairsLocations": [{
            "endLine": 15,
            "endColumn": 42,
            "startColumn": 31,
            "startLine": 15,
            "filePath": "2755_XxeProcessing/InsecureXMLInputFactoryUsage.java",
            "violationSpecifier": "2755:2755_XxeProcessing/InsecureXMLInputFactoryUsage.java:15:31:15:42"
        }],
        "nbViolationsBefore": 7,
        "nbViolationsAfter": 6
    }],
    "executionInfo": {
        "soraldVersion": "LOCAL",
        "javaVersion": "11.0.8",
        "originalArgs": [
            "repair",
            "--original-files-path",
            "/tmp/junit6634076807002230511/project",
            "--stats-output-file",
            "/tmp/junit6634076807002230511/stats.json",
            "--file-output-strategy",
            "IN_PLACE",
            "--violation-specs",
            "2755:2755_XxeProcessing/InsecureXMLInputFactoryUsage.java:15:31:15:42"
        ]
    },
    "totalTimeMs": 987,
    "repairTimeMs": 132,
    "startTimeMs": 1607682619618,
    "crashes": [],
    "endTimeMs": 1607682620605,
    "parseTimeMs": 466
}
```

I have logically divided repairs into two different kinds: _performed_ and _crashed_. A performed repair is not necessarily successful, it just means that the repair exited without errors. A _crashed_ repair is a repair in which the processor attempting the repair suffered an unrecoverable error. If everything is as it should, then `nbViolationsAfter = nbViolationsBefore - nbPerformedRepairs`. **This must be verified for each stats file**, as there may be bugs in our processors that cause them to fail repairs _without_ crashing.

Note that I have not included detailed running times for individual rules, as that would entail quite a lot more work, and it seems rather pointless at this time as we're doing targeted repair of single violations anyway.

The naming doesn't quite match the schema in #246, as I found that there were better names to use. It also follows @monperrus suggestion in #284 to prefer "violation" over "warning".

slarse:
@nharrand If you want to kick off a few experiments ASAP, you can just grab this branch. Usage won't change, and I suspect we still have bugs that need to be fixed before the actual experiments can take place.

khaes-kth:
Thank you @slarse for adding the desired stats to the output.

I think it is now ready to be used by @nharrand for the experiment. WDYT?

slarse:
@khaes-kth I think we will encounter a lot of crashes because of problems in Spoon, the sniper printer in particular. I'm working on it but I don't think we'll ever get it perfect. We might need something like #288 in order for the experiments to be reasonably successful.

Alternatively, we may just want to run in segment mode to begin with. Either way, I must add statistics collection to segment mode, so I'll get on that right away.

PR #291, opened 2020-12-14
refactor: Move and rename repair statistics collector

slarse:
This PR simply moves the repair statistics collector into the `sorald.event.collectors` package, and renames it to `RepairStatisticsCollector`. This is in order to be consistent with the `MinerStatisticsCollector`.


khaes-kth:
Thanks @slarse . After adding MinerStatCollector, we really had to rename this one.

PR #301, opened 2020-12-18
fix: Fix detection of protected repairInternal

slarse:
This PR makes all implementations of `repairInternal` and `canRepairInternal` protected (as they should be as they are protected in `SoraldAbstractProcessor`) to demonstrate that it doesn't actually work (833b84d). That is to say, if you implement `SoraldAbstractProcessor` with `protected` `repairInternal` and `canRepairInternal`, the processor won't work.

The problem is fixed in 50a8673 by using `Class.getDeclaredMethods()` (which finds all declared methods) instead of `Class.getMethods()` (which finds all public methods).


slarse:
@khaes-kth You will want this merged before you try to implement a processor, it was pretty broken :)

khaes-kth:
Thanks a lot @slarse .

PR #303, opened 2020-12-18
feat: Add almost all sonar-java checks

### LINKED ISSUE #302: Add all checks to the miner
Douglas (Verificatum maintainer) requested this. We can do this with a script fairly easily.

slarse:
Fix #302 

My final PR for the year! This PR adds almost all sonar-java checks to the `Checks.java` class, properly categorized. Deprecated rules and rules that don't provide source positions are excluded. This essentially makes the miner complete.

I did this in a semi-automated fashion with this script, applied to the `SonarSource/sonar-java` repo:

```python
import json
import pathlib
import collections
import sys
import re

from typing import Mapping, Tuple, Optional


def main():
    category = sys.argv[1]
    key_to_category = build_key_to_category_lookup(
        pathlib.Path("java-checks/src/main/resources/org/sonar/l10n/java/rules/java")
    )

    check_classes = pathlib.Path(".").rglob("*Check.java")
    check_class_names_in_category = (
        check_class.name
        for check_class in check_classes
        if (key := get_rule_key_from_check_class(check_class))
        and (
            (key_to_category.get(key) or key_to_category.get(check_class.name))
            == category
            or key not in key_to_category
            and category == "DEPRECATED"
        )
    )
    print("\n".join(check_class_names_in_category))


def get_rule_key_from_check_class(check_class: pathlib.Path) -> Optional[str]:
    """Get the rule key from the given class, if it exists. Some check classes
    are abstract and hence don't have a rule key.
    """
    content = check_class.read_text(encoding="utf8")
    key_search = re.search(r'@Rule\(key = "(.*)"\)', content)
    return key_search.group(1) if key_search else None


def build_key_to_category_lookup(rule_metadata_dir: pathlib.Path) -> Mapping[str, str]:
    """Maps the sqKey data from the rule json files in the rule metadata
    directory to their category.

    NOTE: Sometimes, sqKey is the name of the check class rather than the
    actual rulekey.
    """
    return dict(
        map(
            get_key_and_category,
            filter(is_rule_json, rule_metadata_dir.iterdir()),
        )
    )


def is_rule_json(path: pathlib.Path) -> bool:
    return re.match(r"\w\d+_java\.json", path.name) is not None


def get_key_and_category(rule_json_metadata: pathlib.Path) -> Tuple[str, str]:
    data = json.loads(rule_json_metadata.read_text(encoding="utf8"))
    return data["sqKey"], data["type"]


if __name__ == "__main__":
    main()
```

I'm just putting it there for the sake of documentation, it's not very useful in its current state and so I have not committed it to the repo. I'm thinking I'll revisit it next year and make it a neat little script we can use to find the check classes that correspond to a given rule key, and stuff like that.


monperrus:
LGTM, merging.

monperrus:
> My final PR for the year! 

And what a year! Thanks a lot for your deep contributions to Sorald ... and this is only the beginning :)

algomaster99:
@khaes-kth this was the script I was talking about. Thanks @slarse! This will be handy as far as we don't decouple Sorald with `Checks`.

PR #304, opened 2020-12-31
feat(rule 2225): Null-returning "toString()" functions should return empty string

khaes-kth:
It's an incomplete processor for [rule 2225](https://rules.sonarsource.com/java/RSPEC-2225). It only handles null-returning `toString()` and not `clone()`.

### DIFF
--- a/docs/HANDLED_RULES.md
+++ b/docs/HANDLED_RULES.md
@@ -19,7 +19,7 @@ Sorald can currently repair violations of the following rules:
     * [Synchronization should not be based on Strings or boxed primitives](#synchronization-should-not-be-based-on-Strings-or-boxed-primitives-sonar-rule-1860) ([Sonar Rule 1860](https://rules.sonarsource.com/java/tag/multi-threading/RSPEC-1860))
     * [Variables should not be self-assigned](#variables-should-not-be-self-assigned-sonar-rule-1656) ([Sonar Rule 1656](https://rules.sonarsource.com/java/type/Bug/RSPEC-1656))
     * ["Thread.run()" should not be called directly](#threadrun-should-not-be-called-directly-sonar-rule-1217) ([Sonar Rule 1217](https://rules.sonarsource.com/java/type/Bug/RSPEC-1217))
-    * ["toString()" and "clone()" methods should not return null](#tostring-should-not-return-null) ([Sonar Rule 2225](https://rules.sonarsource.com/java/tag/multi-threading/RSPEC-2225))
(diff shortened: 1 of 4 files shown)


slarse:
@khaes-kth Just FYI I'm back on the 5th and will review this then :)

docs/HANDLED_RULES.md
```diff
@@ -343,6 +344,27 @@ class SelfAssignement {
 
 -----
 
+#### "toString()" and "clone()" methods should not return null ([Sonar Rule 2225](https://rules.sonarsource.com/java/type/Bug/RSPEC-2225))
+
+For the return statements inside "toString()", this processor replaces the return expression with an empty string.
```
slarse, line 349: I think you should explicitly state here that the processor is incomplete and ignores `clone()` methods. Ideally, we'd pull that information from the source code annotation, but for now we just have to duplicate the information.

src/main/java/sorald/processor/ToStringReturningNullProcessor.java
```diff
@@ -0,0 +1,30 @@
...
+
+@IncompleteProcessor(description = "does not fix null returning clone()")
+@ProcessorAnnotation(
+        key = 2225,
+        description = "\"toString()\" and \"clone()\" methods should not return null")
+public class ToStringReturningNullProcessor extends SoraldAbstractProcessor<CtReturn<?>> {
+
+    @Override
+    protected boolean canRepairInternal(CtReturn<?> candidate) {
+        CtMethod parentMethod = candidate.getParent(new TypeFilter<>(CtMethod.class));
```
slarse, line 18: Two comments on this line: avoid raw generics, and `CtElement.getParent` has a convenience method for getting an element of a specific type.

```suggestion
        CtMethod<?> parentMethod = candidate.getParent(CtMethod.class);
```

src/main/java/sorald/processor/ToStringReturningNullProcessor.java
```diff
@@ -0,0 +1,30 @@
...
+        if (parentMethod.getSignature().equals(Constants.TOSTRING_METHOD_NAME + "()")
+                && candidate.getReturnedExpression().toString().equals("null")) {
+            return true;
+        }
+        return false;
```
slarse, line 23: This can be simplified. In general, anything on this form:

```java
if (something) {
    return true;
}
return false;
```
can be simplified like so:

```java
return something;
```
So, it would make it something like this:

```suggestion
        return parentMethod.getSignature().equals(Constants.TOSTRING_METHOD_NAME + "()")
                && candidate.getReturnedExpression().toString().equals("null");
```

slarse (changes requested):
Looks functionally A-okay, just have some minor comments that you may want to address before merge.

slarse (approved):
Looks good, nicely done!

PR #313, opened 2021-01-08
experiment: Add script for generating ACHIEVEMENTS.md

### LINKED ISSUE #255: add `docs/achievements.md` with links to PR
We will add `docs/achievements.md` with the list of merged PRs by Haris, Henry, Simon and the future ones to come.

slarse:
Related to #255 

This is a first version of the `ACHIEVEMENTS.md` generation script. Currently generates an `ACHIEVEMENTS.md` file that looks like so.

```markdown
# Achievements
This document presents an overview of the pull requests performed with Sorald.

## [redhat-developer/rsp-server#619](https://github.com/redhat-developer/rsp-server/pull/619)
This PR was opened at 2020-11-25 12:01:06 and merged at 2020-11-30 20:45:38.
The patch was generated fully automatically with Sorald.
It provided the following repairs:

* [Rule 2755](https://rules.sonarsource.com/java/RSPEC-2755)
    - Number of violations found: 4
    - Number of violations repaired: 4
```

There is some amount of variation in what this output looks like depending on the data available for the given PR. For example, if the PR isn't merged, the output won't say that it is. If there are manual edits, the output will say so instead of "fully automatically with Sorald".

See the added README file for usage.

The tests aren't included in the GitHub actions workflow at this point, I intend to include them once I have unified our support scripts as per #325. A new GitHub action for running this automatically each time we update (the currently non-existing) prs.json will also follow, which is why I haven't marked #255 as "fixed" in this PR.


slarse:
I want to discuss this during the meeting before putting more time into it.

experimentation/tools/achievements/README.md
```diff
@@ -0,0 +1,26 @@
+# Achievements script
+This is a script that can be used to generate an "achievements" file for
+Sorald, detailing the pull requests performed with it. It's essentally
```
fermadeiral, line 3: English is not my mother tongue, thus I'm sorry in advance if I'm wrong:
"essentally" -> "essentially"

experimentation/tools/achievements/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Install
+Install the packages listed in `requirements.txt`.
+
+```bash
+$ python3 -m pip install -r requirements.txt
+```
+
+> **Note:** You should do this in a virtual environment to avoid polluting
+> system-wide or user packages.
```
fermadeiral, line 17: I believe it can be useful to give some basic commands here, e.g.:

```bash
pip install virtualenv
virtualenv --python=python3.6 env
source ./env/bin/activate
```

WDYT?

slarse: I agree, but I think it would be better left for #326 where I bring all of the scripts together, as these particular instructions currently exist in several of the script directories (yay, duplication).

experimentation/tools/achievements/README.md
slarse: That's definitely a typo, fixed!

experimentation/tools/achievements/README.md
fermadeiral: Understood and agreed!

monperrus:
Nice!

PR #315, opened 2021-01-11
fix: Harden violation scanning against NullPointerException

### LINKED ISSUE #314: Must harden violation scanning against lack of primary location
After adding (almost) all checks in #303 , we've started getting some null pointer exception on accessing e.g. `ScannedViolation.getStartLine()` due to the fact that some violations don't have primary locations. We should harden the implementation against these NPEs, and if they do occur, the error message should be more informative than... NPE.

slarse:
Fix #314 

Hardens violation scanning against NPEs by filtering out any `AnalyzerMessage`s that don't have primary locations. If one were to somehow slip through anyway, `ScannedViolation` is now adapted to provide better error messages.


khaes-kth:
Thanks @slarse

PR #318, opened 2021-01-11
experiment: Consistently use camel case in output from PR recorder

slarse:
There was a mix of camel case and snake case in the output from the PR recorder script, as the sorald statistics are always camel-cased. This caused the whole file to look rather strange. With this PR, I'm simply making it consistent by using camel case for all attributes in the JSON output.

Also consolidating all JSON keys into a single module in the Python scripts so they can be reused across scripts. I'll probably create something DSL-like in the end to work with the PR data in Python, but for now this will due. I need this merged quickly so I can set up the automatic updates of the prs.json file and the achievements.md file.


fermadeiral:
Nice! Thank you, @slarse!

Issue #295, opened 2020-12-16
Add option to miner to only mine violations for which there is a corresponding processor in Sorald


slarse:
It would be convenient for me to be able to run the miner only for rules that we actually have processors for. We could add an option along the lines of `--handled-rules` to filter for rules that we have processors for.

@khaes-kth would you like to do this? Not urgent, but I'd really like to have it.

khaes-kth:
Sure, I'll do that.

slarse:
This was fixed in #319

PR #319, opened 2021-01-11
feat: Adds an option to mine only implemented rules

khaes-kth:
This PR is going to address #295 .

### DIFF
--- a/src/main/java/sorald/Constants.java
+++ b/src/main/java/sorald/Constants.java
@@ -25,6 +25,7 @@ public class Constants {
     public static final String ARG_REPAIR_STRATEGY = "--repair-strategy";
     public static final String ARG_MAX_FILES_PER_SEGMENT = "--max-files-per-segment";
     public static final String ARG_RULE_TYPES = "--rule-types";
+    public static final String ARG_HANDLED_RUES = "--handled-rules";
     public static final String ARG_RULE_VIOLATION_SPECIFIERS = "--violation-specs";
     public static final String ARG_TARGET = "--target";
 
--- a/src/main/java/sorald/Constants.java
+++ b/src/main/java/sorald/Constants.java
@@ -25,7 +25,7 @@ public class Constants {
     public static final String ARG_REPAIR_STRATEGY = "--repair-strategy";
     public static final String ARG_MAX_FILES_PER_SEGMENT = "--max-files-per-segment";
     public static final String ARG_RULE_TYPES = "--rule-types";
-    public static final String ARG_HANDLED_RUES = "--handled-rules";
+    public static final String ARG_HANDLED_RULES = "--handled-rules";
     public static final String ARG_RULE_VIOLATION_SPECIFIERS = "--violation-specs";
     public static final String ARG_TARGET = "--target";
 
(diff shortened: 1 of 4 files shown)


src/main/java/sorald/cli/MineCommand.java
```diff
@@ -56,10 +55,35 @@
...
+                                                                                    Rule.class))
+                                                            .map(Rule::key)
+                                                            .map(Checks::stripDigits)
+                                                            .map(Integer::parseInt)
+                                                            .map(Processors::getProcessor)
+                                                            .filter(Objects::nonNull)
+                                                            .count()
+                                                    > 0)
+                            .collect(Collectors.toList());
+        }
```
slarse, line 85: This can be simplified:

```suggestion
        if (handledRules) {
            checks =
                    checks.stream()
                            .filter(
                                    sc -> {
                                        int key =
                                                Integer.parseInt(Checks.getRuleKey(sc.getClass()));
                                        return Processors.getProcessor(key) != null;
                                    })
                            .collect(Collectors.toList());
        }
```

I would also move it into the `inferCheckInstances` method to not pollute the top level logic of the `call` method.

src/main/java/sorald/sonar/Checks.java
```diff
@@ -111,7 +111,7 @@ public static String getRuleKey(Class<? extends JavaCheck> checkClass) {
                                         checkClass.getName() + " does not have a key"));
     }
 
-    private static String stripDigits(String s) {
+    public static String stripDigits(String s) {
```
slarse, line 114: Given the suggestion above, this change can be reverted.

On a totally unrelated note: this method does the exact opposite of stripping digits (it strips everything BUT digits). Should be renamed. But that's for another PR :)

src/main/java/sorald/Constants.java
```diff
@@ -25,6 +25,7 @@
     public static final String ARG_REPAIR_STRATEGY = "--repair-strategy";
     public static final String ARG_MAX_FILES_PER_SEGMENT = "--max-files-per-segment";
     public static final String ARG_RULE_TYPES = "--rule-types";
+    public static final String ARG_HANDLED_RUES = "--handled-rules";
```
slarse, line 28: Typo

```suggestion
    public static final String ARG_HANDLED_RULES = "--handled-rules";
```

src/main/java/sorald/cli/MineCommand.java
slarse: Unrelated to this PR: It also seems to me like we should change the keys we use internally to match those in Sonar. It doesn't make sense for us to use integers instead of the strings that sonar-java uses, there's 0 benefit to that.

slarse (changes requested):
Seemingly works great, I just have some minor notes on the code.

src/main/java/sorald/cli/MineCommand.java
khaes-kth: > It doesn't make sense for us to use integers instead of the strings that sonar-java uses, there's 0 benefit to that

Great point!

slarse:
LGTM! This will help me a whole lot in my work to find stuff to repair, thanks @khaes-kth

slarse (approved):

PR #320, opened 2021-01-12
feat: Add support for ignoring rules with SuppressWarnings("squid:SXXXX")

### LINKED ISSUE #317: Add support for @SuppressWarnings("squid:SXXXX")
Sorald currently does not respect `@SuppressWarnings("squid:SXXXX")` annotations. It would be preferable if we did, as those represent things that the authors of a given project definitely don't want to repair.

slarse:
Fix #317 

This PR adds support for ignoring specific rules with `@SuppressWarnings` annotations, like so:

```java
public class WithMethodSuppressedS2116 {

    @SuppressWarnings("squid:S2116")
        String argStr = args.toString(); // Noncompliant (rule S2116, should be ignored)
        float f = 1.0f + 2.0f; // Noncompliant (rule S2164, should not be ignored)
    }
}
```


slarse:
Side note: To the extent possible, I've tried to avoid using internal components. However, as Sorald is to some extent pretending to be SonarLint, I don't think entirely avoiding internal SonarLint components is possible with our use case without implementing our own versions of a large amount of non-trivial of things.

src/test/java/sorald/miner/WarningMinerTest.java
```diff
@@ -176,7 +175,7 @@ public void extractWarnings_throwsException_whenCheckCrashes() {
                 };
 
         assertThrows(
-                AnalysisException.class,
+                IllegalArgumentException.class,
```
khaes-kth, line 178: @slarse Can you explain why the exception type is changed? I don't get it.

src/main/java/sorald/sonar/RuleVerifier.java
```diff
@@ -172,8 +211,48 @@ public void reportIssue(AnalyzerMessage analyzerMessage) {
...
+                    .filter(message -> postFilter.accept(getRuleKey(message), message))
+                    .collect(Collectors.toList());
+        }
+
+        public JavaClasspath getClasspath() {
+            return cp;
+        }
+
+        private static RuleKey getRuleKey(AnalyzerMessage message) {
+            return RuleKey.of("java", "S" + Checks.getRuleKey(message.getCheck().getClass()));
```
khaes-kth, line 235: Another sign that someday we should change our ruleKeys from String to Integer.

src/test/java/sorald/miner/WarningMinerTest.java
slarse: That does look a bit odd. It's probably a quick-fix I applied without thinking too much about it. I'll have a look

src/main/java/sorald/sonar/RuleVerifier.java
slarse: Indeed, I really think we should get on that sooner rather than later. It's getting rather annoying.

khaes-kth: > from String to Integer

* from Integer to String :D

src/test/java/sorald/miner/WarningMinerTest.java
slarse: It was testing entirely the wrong thing. With the use of the new higher-level components, I've not been able to figure out how to load and run a custom check. This isn't something that we actually have a need to do, and so I've simply removed the functionality that this test tested (i.e. that Sorald crashes if a Sonar check crashes).

src/main/java/sorald/sonar/RuleVerifier.java
slarse: I've turned this into a helper function `Checks.toSonarRuleKey` and added a TODO that we should remove it by simply using the Sonar keys in Sorald.

slarse:
@khaes-kth I think this is ready for review again.

slarse:
Noticed that this branch was behind master, so just merged now to check that there are no problems.

khaes-kth:
Thanks @slarse

PR #322, opened 2021-01-12
feat: Use literal suffixes for CastArithmeticOperandProcessor when applicable

### LINKED ISSUE #321: CastArithmeticOperandProcessor should append an L to literal integers instead of casting to long
I want to apply this processor to Verificatum, but it doesn't look good to have it change a literal int `1000` to `(long) 1000`. It should be `1000L`.

slarse:
Fix #321 

This PR changes `CastArithmeticOperandProcessor`'s transformations to use literal suffixes instead of casts when applicable. Here's an example:

```java
// original statement
long val = 1000 * someIntValue;

// old transformation
long val = (long) 1000 * someIntValue;

// new transformation
long val = 1000L * someIntValue;
```

Casts are still used where literal suffixes aren't applicable.

@fermadeiral WDYT, should we make this the default? I personally prefer literal suffixes, but as you noted in #321 none of the transformations or poor from an idiomatic standpoint. We could also hold off until we have the ability to supply options to processors.

### DIFF
--- a/docs/HANDLED_RULES.md
+++ b/docs/HANDLED_RULES.md
@@ -206,16 +206,18 @@ Example:
 
 #### Math operands should be cast before assignment ([Sonar Rule 2184](https://rules.sonarsource.com/java/RSPEC-2184))
 
-In arithmetic expressions, when the operands are `int` and/or `long`, but the result of the expression is assigned to a `long`, `double`, or `float`, the first operand is casted to the final type before the operation takes place.
+In arithmetic expressions, when the operands are `int` and/or `long`, but the result of the expression is assigned to
+a `long`, `double`, or `float`, the first left-hand is casted to the final type before the operation takes place.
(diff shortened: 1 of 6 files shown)


fermadeiral:
> @fermadeiral WDYT, should we make this the default? I personally prefer literal suffixes

I think we can make it the default for now.

> We could also hold off until we have the ability to supply options to processors.

I would proceed like this: we merge this PR, we close #321, we create a new issue with the "supply options to processors" thing and give issue #321 as an example. WDYT?

slarse:
> I would proceed like this: we merge this PR, we close #321, we create a new issue with the "supply options to processors" thing and give issue #321 as an example. WDYT?

Sounds good to me!

fermadeiral:
Thanks @slarse!

Could you handle the closing/opening of the issues, please?

slarse:
@fermadeiral Done, see #334

PR #323, opened 2021-01-12
feat: Respect NOSONAR comments

### LINKED ISSUE #316: Sorald does not respect NOSONAR comments
Neither miner nor repairer respects `// NOSONAR` comments. The `NoSonarTest` is broken as it makes the post-test check against the original file rather than the repaired file.

slarse:
Fix #316 

Re-implements feature for respecting `// NOSONAR` comments in the source code. The test to check this functionality has also been fixed, and a new one added.


slarse:
The poor diff hit is because of the defensive programming on this line:

https://codecov.io/gh/SpoonLabs/sorald/pull/323/diff?src=pr&el=tree#D1-237

I'm pretty sure the input component can be something else, so I want that check there. But I've not been able to reproduce it artificially, and hence there's a partial hit on that expression (it always evaluates to true).

khaes-kth:
Thanks @slarse

PR #326, opened 2021-01-14
experiment: Unify Python support scripts into a single package

### LINKED ISSUE #325: Unify supporting scripts into a single Python package
It will be a lot easier to maintain, and in particular to test, the support scripts if they are all in a single package. As several of them are closely related (e.g. the PR recorder and achievements scripts), it makes a lot of sense to have them live more closely together.

slarse:
Fix #325 

~Depends upon #313 being merged first.~

This PR unifies all of our Python scripts into a single Python package. This makes testing, usage and development a whole lot easier. It also collects all of the documentation for the scripts into a single [README](https://github.com/slarse/sorald/blob/issue/325-unify-python-scripts/experimentation/tools/README.md), which should help a lot with reducing duplicated instructions. I haven't gone for full best-practices Python development here as that might be cumbersome to Sorald contributors who perhaps don't know Python well, but have instead cherry-picked practices that makes maintenance easier.

This setup also allows for code reuse between scripts. Although I have not done so in this PR, the PR recorder and achievements scripts are prime candidates for reusing code (especially in the metadata keys they both define).

It's still very easy to add a new script even for those who aren't used to writing Python packages, as described in the [README](https://github.com/slarse/sorald/blob/issue/325-unify-python-scripts/experimentation/tools/README.md#adding-a-new-script).


slarse:
As per the discussion in #313, the little note about using a virtual environment in the install instructions should be expanded with instructions on how to actually do that.

slarse:
@fermadeiral This is ready for review! Have a look at the [README](https://github.com/slarse/sorald/blob/issue/325-unify-python-scripts/experimentation/tools/README.md) for detailed usage.

What do you think?

experimentation/tools/README.md
```diff
@@ -63,7 +147,7 @@ with statistics collection.
 The command executed above creates a file `prs.json` with an initial record
 like this:
 
-```
+```json
```
fermadeiral, line 150: I think there is a " that should be removed in line 158, where "merged_at" is.

slarse: Yep, wonder how that got there. Removed.

fermadeiral:
That is great, Simon! Thanks!

monperrus:
Indeed, thanks!

PR #328, opened 2021-01-15
refactor: Commit generated code instead of generating on compile

### LINKED ISSUE #294: Adopt Spoon's approach of committing generated sources to the repository
My dev environment (IntelliJ IDEA) started acting up last Friday, crashing left and right. I've lost about half a workday since trying to work around it.

I've also tried switching to a different environment, but it's _really_ hard to get an arbitrary development environment to properly recognize the generated sources. As such, I think we should adopt Spoon's approach of generating sources and then committing them to the repository. It just saves so much headache when stuff like this happens.

slarse:
Fix #294 

This PR redesigns the code generation such that we commit the code to the repository, instead of relying on annotation processing during compile time. While annotation processing at compile time works great for Maven, I've found that it's in general very tricky to get various IDEs and editors to recognize the generated code as part of the code base. We've only gotten it to work somewhat well with IntelliJ IDEA so far.

So, this PR is meant to make Sorald development friendlier for more development environments. There's one unfortunate side effect, and that is that adding a new processor requires one to manually run the code generation. This isn't hard, but it is an extra step. It's described in CONTRIBUTING.md. There's also a new GitHub Actions job that checks that the generated code is up-to-date.

Overall, this also greatly simplifies interaction with the `Processors` class, as it is now possible to add/remove/edit methods and fields by just editing the source file, as long as one does not touch the generated fields.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -66,9 +66,9 @@ with `ProcessorAnnotation`.
 
 As soon as you have created a skeleton for the processor, with the
 `ProcessorAnnotation` properly filled in, you should update the `Processors`
-class. You do this by running the main method of
-`sorald.annotations.ProcessorsClassGenerator`, which you can either do with
-your favorite IDE, or by running the following shell commands in `bash`:
+class. You do this by running the main method of `sorald.CodeGenerator`, which
+you can either do with your favorite IDE, or by running the following shell
+commands in `bash`:
 
 ```bash
 $ mvn clean compile dependency:build-classpath -Dmdep.outputFile=cp.txt
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -2,22 +2,6 @@
 
 Pull requests are very welcome by the Sorald team!
 
-### A note on using an IDE to develop Sorald
-
-Due to a somewhat unconventional annotation processor setup that includes a
-precompile step to compile the processor, IDEs sometimes have trouble building
-Sorald. There are currently two workarounds for this.
-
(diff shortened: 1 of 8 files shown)


docs/CONTRIBUTING.md
```diff
@@ -72,19 +56,34 @@ So the name of your new processor is `CastArithmeticOperandCheck` replacing "Che
...
-When you have created your processor, you must also add the check class to one
-of the four categories of check classes in the static code block in
-[Checks.java](/src/main/java/sorald/sonar/Checks.java), if it is not already
-present. To find out which category to add your check to, look it up on the
-[SonarSource website](https://rules.sonarsource.com/java), the category is
-listed just below the rule title.
+As soon as you have created a skeleton for the processor, with the
+`ProcessorAnnotation` properly filled in, you should update the `Processors`
+class. You do this by running the main method of
+`sorald.annotations.ProcessorsClassGenerator`, which you can either do with
```
khaes-kth, line 70: > by running the main method of `sorald.annotations.ProcessorsClassGenerator`

or by running the main method of `sorald.CodeGenerator`?

slarse: Yep, good catch :)

slarse: Fixed

khaes-kth:
Thanks @slarse . Making contributions to Sorald will be easier after this :)

slarse:
I agree. The compile-time annotation processing was cool and all, but it was very annoying to deal with in practice.

monperrus:
Per the saying, __"In Theory There Is No Difference Between Theory and Practice, While In Practice 
There Is"__

PR #336, opened 2021-01-19
fix: Fix printing of compilation units with multiple top-level types

### LINKED ISSUE #333: Sorald fails to print unedited top-level types in compilation unit with multiple types
Sniper printer throws an exception on printing InterruptedExceptionForTesting.java. It doesn't matter if Sorald repairs anything in the file, there's a crash anyway. I think it's related to the fact that there are multiple type definitions in the file. Will investigate in Spoon.

slarse:
Fix #333 

Fixes the bug in Sorald described in #333, which solves the problem for us. The related bug in Spoon that's mentioned is not a problem for us, although I'll fix that too eventually. Here's the gist of the bug in Sorald:

> In Sorald, we only print top-level types that have been edited. This means that a CU with multiple top-level types may be printed with fewer types than it should, if any of the top-level types have not been edited. This bug is severe, as even when running with the default printer, there are types missing from the output. This also shows the dire need for expected output files for all test cases.

This PR changes the printing logic to work on compilation units instead of individual types. This both fixes the crash in the sniper printer, and the incorrect output of the normal printer.

In the future, the `UniqueTypesCollector` should probably be rewritten as a "compilation unit collector" instead, as that's the level of granularity we are concerned with when writing output.

### DIFF
--- a/src/main/java/sorald/CompilationUnitHelpers.java
+++ b/src/main/java/sorald/CompilationUnitHelpers.java
@@ -4,11 +4,9 @@ import java.io.File;
 import java.nio.file.Path;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.IdentityHashMap;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.Function;
 import spoon.reflect.declaration.CtCompilationUnit;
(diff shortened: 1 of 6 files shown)


slarse:
Skip printing CU without well-defined top-level type.

PR #342, opened 2021-01-26
fix: completely exclude xml-apis dependency from project

### LINKED ISSUE #330: The package javax.xml is accessible from more than one module
I get the error in the title in both Eclipse and VS Code (VS code uses an Eclipse backend for Java stuff), and the project refuses to compile. Apparently, a lot of stuff in `javax.xml` is distributed with the standard library (where it's in a module), but it appears as if dependencies that also have `javax.xml` in it. The dependency likely pre-dates Java 9, and so `javax.xml` is not in a module (it says "unknown module"), while it is in the standard library.

The only IDEA that I've found to work around this is IntelliJ, possibly due to better Maven integration than the others. Maven has no problem building the project.

slarse:
Fix #330 

At long last, a fix for this very annoying problem. To summarize #330, the problem is that the `javax.xml` package is both defined in a module in the standard library, _and_ in the `xml-apis` bundle in an unnamed module (i.e. pre java 9 style). This is technically not allowed, a package can only appear in one module, and so _most_ dev environments scream bloody murder. It also causes problems at runtime as JDT has trouble resolving references to `javax.xml.*` classes.

The problem appears to be with version 1.4.01 of xml-apis in particular, whereas 2.0.2 appears to work fine. Strangely enough, 1.4.01 (released in 2011) appears to be much newer than 2.0.2 (released in 2005). I'm not quite sure what's going on there.

Now, we can exclude this dependency from `java-checks-testkit` and `java-checks`, but it turns out that it's also bundled in the packaged version of `sonar-java-api`. We therefore remove our dependency on `sonar-java-plugin`, and exclude `xml-apis` from the remaining sonar dependencies.

Coincidentally, the refactoring I had to perform to make it possible to remove the `sonar-java-plugin` dependency gives us more control over how files are scanned, and it's not more complicated code either. This is probably what it should have looking like initially, I just didn't know that at the time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,6 +83,13 @@
             <groupId>org.sonarsource.java</groupId>
             <artifactId>java-checks-testkit</artifactId>
             <version>6.9.0.23563</version>
+            <exclusions>
+                <exclusion>
+                    <!-- included in stdlib -->
+                    <groupId>xml-apis</groupId>
+                    <artifactId>xml-apis</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.sonarsource.java</groupId>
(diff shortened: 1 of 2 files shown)


monperrus:
Thanks a lot; nice debugging!

Issue #205, opened 2020-11-10
Add possibility to run Sorald in classpath mode (and specify the classpath)


slarse:
Right now, Sorald runs in noclasspath mode. That's a problem for complex repairs, as they may become impossible to perform without complete typing information. For example, right now I'm having trouble with the following crashing with a `NullPointerException`:

```java
    @Override
    public boolean isToBeProcessed(CtInvocation<?> candidate) {
        String typeName = candidate.getExecutable().getType().getSimpleName();
```

The type of the executable is not known, and so `null` is returned.

slarse:
Providing classpath can also improve Sonar's precision.

monperrus:
This may benefit/impact Sonar itself which uses the classpath in some cases.

monperrus:
We could use the MavenLauncher to solve this problem.

slarse:
For the time being, we'll make this an exclusive feature of Maven mode.

slarse:
Here's a very simple code snippet to test classpath mode (using stuff from `gumtree-spoon-ast-diff`):

```java
Diff diff = new AstComparator().compare(leftPath.toFile(), rightPath.toFile());
double ratio = diff.getRootOperations().size() / diff.getAllOperations().size();
```

This _should_ result in a violation of rule 2184 (cast arithmetic operator), but without the classpath, Sonar can't deterimne that `diff.get{Root/All}Operations().size()` is an int, and thus won't flag a violation.

PR #347, opened 2021-01-28
feat: Add new MAVEN mode that uses Spoon's MavenLauncher

### LINKED ISSUE #297: Use Spoon's Maven launcher in Maven projects
As suggested by @nharrand , we may want to use the Maven Spoon launcher for Maven projects. We could either add this as an auto-detection feature (look for pom.xml in project root directory), or as a flag (`--maven`).

This could potentially make default mode more reliable and avoid some of the issues pointed out in #287. It probably won't work at all for segment mode, though.

Thoughts?

slarse:
Fix #297 

This also relates to #205, but we don't provide the classpath to Sonar yet.

This PR adds a new repair strategy, `MAVEN` (i.e. run with `--repair-strategy MAVEN` to activate it), that uses Spoon's `MavenLauncher` to build the model. ~Test code is ignored for now. Sonar does have separate treatment of test code and production code, so we could potentially expand upon this in the future, but let's make it work for production code first.~

Note the following:

* Runs Spoon in classpath mode, which gives better type resolution, but it can also cause crashes due to failing to resolve Maven dependencies.
* Maven mode only supports in-place repair when repairing multiple rules, as pom files and other supporting resources aren't printed in the intermediate directories for the other modes.
    - Single-rule repair works for all output strategies.
* ~Test code is ignored at the moment, but _only for repairs_. We still mine violations from all code, meaning that we may get many violations that Maven mode can't possibly repair  as it doesn't operate on test code.~
* Even though we can now resolve the full classpath, we don't yet provide it to Sonar. This PR is large enough as it is!

So, as is evident by this PR, "just using the MavenLauncher" wasn't really the whole story :)


slarse:
Coverage drop due to this statement (can't really be tested): https://codecov.io/gh/SpoonLabs/sorald/pull/347/diff?src=pr&el=tree#D11-157...158

slarse:
@khaes-kth OOPS, I totally forgot I was supposed to refactor a thing here before requesting a review. Sorry, hope you didn't have time to start, don't merge yet!

slarse:
There we go, much better. Ready for review again.

khaes-kth:
> Ready for review again

Great, I'll review it soon.

src/main/java/sorald/Repair.java
```diff
@@ -162,10 +167,19 @@ CtModel defaultRepair(
...
         repairModelWithInitializedProcessor(model, processor, violations);
-        EventHelper.fireEvent(EventType.REPAIR_END, eventHandlers);
+        return model;
+    }
 
+    CtModel mavenRepair(
+            Path inputDir, SoraldAbstractProcessor<?> processor, Set<RuleViolation> violations) {
+        EventHelper.fireEvent(EventType.PARSE_START, eventHandlers);
+        MavenLauncher launcher =
+                new MavenLauncher(inputDir.toString(), MavenLauncher.SOURCE_TYPE.APP_SOURCE);
```
monperrus, line 178: change to ALL_SOURCE?

slarse: Done. I'm not quite sure why I decided to only process production code.

src/test/resources/processor_test_files/2111_BigDecimalDoubleConstructor/BigDecimalDoubleConstructor.java.expected
```diff
@@ -26,7 +26,7 @@ public class BigDecimalDoubleConstructor {
     // Aditional tests
     public void foo(String[] args) {
         double d = 1.1;
-        float f = 2.2;
+        float f = 2.2f;
```
khaes-kth, line 29: @slarse Seems good to me.

Just one question: why did you have to change many of the tests like this one?

khaes-kth:
Seems there is a conflict here, can you resolve it @slarse ?

slarse:
@khaes-kth Fixed

khaes-kth:
> @khaes-kth Fixed

Good. My only question is what I asked about the tests. I don't quite understand why many test files have changed.

src/test/resources/processor_test_files/2111_BigDecimalDoubleConstructor/BigDecimalDoubleConstructor.java.expected
slarse: tl;dr: I needed the files to compile.

The Maven launcher test grabs all of the test files from `resources/processor_test_files` and creates a maven project out of them, and then runs a repair. Many of those files did not compile, either because they were syntactically invalid (found a few methods nested in methods and stuff like that), semantically invalid (such as the above, you can't assign a double literal to a float variable), or needlessly contained type references that were not on the classpath.

slarse:
@khaes-kth Totally missed your comment, sorry. I've added a response.

src/test/resources/processor_test_files/2111_BigDecimalDoubleConstructor/BigDecimalDoubleConstructor.java.expected
khaes-kth: Hmm, I thought so. Seems that we had many noncompilable test files.

PR #349, opened 2021-02-02
experimentation: Add new prs.json file

slarse:
#255 

This PR adds the new prs.json file with two entries (rsp-server and verificatum-vmn PRs), generated with the `sorald.prrecorder` script.

PR #350, opened 2021-02-02
doc: Add achievements.md

slarse:
#255 

Adds the long-awaited ACHIEVEMENTS.md file to the docs directory. This is generated from the `prs.json` file in #349, so we should merge that first.


slarse:
Actually, hold off on merging this one, it might be more fun to do the automatic generation immediately!

PR #351, opened 2021-02-02
fix: fix an IndexOutOfBoundsException in CastArithmeticOperandProcessor

### LINKED ISSUE #125: Array out of bound exception rule key 2184

fermadeiral:
This PR addresses #125 partially.

The thing is: when the arithmetic operation is inside a method call, and an inner binary operator of the arithmetic operation is being analyzed, we don't find it in the list of arguments of the method call, because that binary operator is part of something more. Not finding it in the list of arguments causes an `IndexOutOfBoundsException` because we have index `-1`. This PR fixes that. Note that such a case was added in the test, which triggers the exception if the patch added in this PR does not exist.

### DIFF
--- a/src/test/resources/processor_test_files/2184_CastArithmeticOperand/CastArithmeticOperand.java
+++ b/src/test/resources/processor_test_files/2184_CastArithmeticOperand/CastArithmeticOperand.java
@@ -11,6 +11,7 @@ class CastArithmeticOperand {
     long bigNegNum =  Integer.MIN_VALUE-1; // Noncompliant, gives a positive result instead of a negative one.
     int seconds;
     Date myDate = new Date(seconds * 1000); // Noncompliant, won't produce the expected result if seconds > 2_147_483
+    Date myDate2 = new Date(seconds * 1000 * 10); // Noncompliant
 
     public long compute(int factor){
         return factor * 10000; // Noncompliant, won't produce the expected result if factor > 214_748
--- a/src/test/resources/processor_test_files/2184_CastArithmeticOperand/CastArithmeticOperand.java
+++ b/src/test/resources/processor_test_files/2184_CastArithmeticOperand/CastArithmeticOperand.java
@@ -11,7 +11,6 @@ class CastArithmeticOperand {
     long bigNegNum =  Integer.MIN_VALUE-1; // Noncompliant, gives a positive result instead of a negative one.
     int seconds;
     Date myDate = new Date(seconds * 1000); // Noncompliant, won't produce the expected result if seconds > 2_147_483
-    Date myDate2 = new Date(seconds * 1000 * 10); // Noncompliant
 
     public long compute(int factor){
         return factor * 10000; // Noncompliant, won't produce the expected result if factor > 214_748
--- a/src/test/resources/processor_test_files/2184_CastArithmeticOperand/MultipleBinaryOperatorsInMethodCall.java
+++ b/src/test/resources/processor_test_files/2184_CastArithmeticOperand/MultipleBinaryOperatorsInMethodCall.java
@@ -3,6 +3,9 @@ is being analyzed, we don't find it directly in the list of arguments of the met
 is part of something more. Not finding it in the list of arguments causes an IndexOutOfBoundsException because
 we have index -1. This test checks that that exception doesn't happen.
  */
+import java.util.Date;
+
 class MultipleBinaryOperatorsInMethodCall {
+    int seconds = 1000;
     Date myDate = new Date(seconds * 1000 * 10); // Noncompliant
 }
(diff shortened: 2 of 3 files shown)


src/test/resources/processor_test_files/2184_CastArithmeticOperand/CastArithmeticOperand.java
```diff
@@ -11,6 +11,7 @@
     long bigNegNum =  Integer.MIN_VALUE-1; // Noncompliant, gives a positive result instead of a negative one.
     int seconds;
     Date myDate = new Date(seconds * 1000); // Noncompliant, won't produce the expected result if seconds > 2_147_483
+    Date myDate2 = new Date(seconds * 1000 * 10); // Noncompliant
```
slarse, line 14: Like we discussed at the meeting, would you mind putting this line in a separate test file, name it something informative, and just put a small comment in the source of that test file to clarify what it's testing?

As the test source files _are_ the test cases, I think we should try to document them as such as well (I've not been great at that myself, but it's time to get better! :)

slarse (changes requested):
See comment

fermadeiral:
> As the test source files are the test cases, I think we should try to document them as such as well (I've not been great at that myself, but it's time to get better! :)

I agree with you. However, I'm not sure I can do that every time that I fix something (since my time is limited and the deadline is approaching). From a project on which Sorald has bad behavior until to arrive to that fix and that little, minimal line to trigger the wrong behavior already took some time. I will do that now, but not sure I will be able to do that for the short future.

fermadeiral:
@slarse, there are some test failures but I think that I did not cause them. Do you have some idea?

src/test/resources/processor_test_files/2184_CastArithmeticOperand/MultipleBinaryOperatorsInMethodCall.java
```diff
@@ -0,0 +1,8 @@
+/* When the arithmetic operation is inside a method call, and an inner binary operator of the arithmetic operation
+is being analyzed, we don't find it directly in the list of arguments of the method call, because that binary operator
+is part of something more. Not finding it in the list of arguments causes an IndexOutOfBoundsException because
+we have index -1. This test checks that that exception doesn't happen.
+ */
+class MultipleBinaryOperatorsInMethodCall {
+    Date myDate = new Date(seconds * 1000 * 10); // Noncompliant
+}
```
slarse, line 8: I'm pretty sure the test failures are because this file isn't complete, and so neither Spoon nor Sonar can properly resolve all types. Try this:

```suggestion
import java.util.Date;

class MultipleBinaryOperatorsInMethodCall {
    int seconds = 1000;
    Date myDate = new Date(seconds * 1000 * 10); // Noncompliant
}
```

slarse:
> > As the test source files are the test cases, I think we should try to document them as such as well (I've not been great at that myself, but it's time to get better! :)
> 
> I agree with you. However, I'm not sure I can do that every time that I fix something (since my time is limited and the deadline is approaching). From a project on which Sorald has bad behavior until to arrive to that fix and that little, minimal line to trigger the wrong behavior already took some time. I will do that now, but not sure I will be able to do that for the short future.

Of course, maintainability is always a tradeoff between what you've got time to do now and how much you expect it to pay off in the future. I personally highly prioritize minimal test cases such that I know that I've actually reproduced and fixed the error, but if you don't have the time to do it, then you don't have the time to do it.

A not quite as good, but still sufficient solution is to put a big comment in the test file at the exact point where the expected behavior is triggered, and then someone can possibly refactor that at a later time.

> @slarse, there are some test failures but I think that I did not cause them. Do you have some idea?

See my comment on the new test file.

src/test/resources/processor_test_files/2184_CastArithmeticOperand/MultipleBinaryOperatorsInMethodCall.java
fermadeiral: Ooops, my bad, I will fix that! Thank you!

fermadeiral:
> I personally highly prioritize minimal test cases such that I know that I've actually reproduced and fixed the error

Me too, which is why I took time to open this PR and probably will take time to open new ones, because I always aim adding a minimal test case to reproduce the bug.

src/test/resources/processor_test_files/2184_CastArithmeticOperand/MultipleBinaryOperatorsInMethodCall.java
slarse: Super easy to miss, I just happened to fix a bunch of these issues a few days ago and so spotted it immediately :)

fermadeiral: I would take forever.

slarse:
> > I personally highly prioritize minimal test cases such that I know that I've actually reproduced and fixed the error
> 
> Me too, which is why I took time to open this PR and probably will take time to open new ones, because I always aim adding a minimal test case to reproduce the bug.

I know :). I just wanted to clarify that I'm well aware that it is a tradeoff, and it's not always possible to take the ideal path as time is a factor.

Anyway, this looks great to me now, happy to merge?

slarse (approved):

fermadeiral:
> Anyway, this looks great to me now, happy to merge?

Yes! Thank you!

PR #353, opened 2021-02-02
chore: Add missing parenthesis in support workflow

slarse:
I thought that `expr1 || expr2 && expr3` would implicitly be `expr1 || (expr2 && expr3)` as `&&` typically binds "harder" than `||`. That's apparently not the case in bash :D


slarse:
I'm just going to merge this right away

PR #356, opened 2021-02-02
feat: Add support for projects with named modules

### LINKED ISSUE #339: Spoon Java compliance level set too low, causes crash on encountering module-info.java
Failure 1 in #337 

>  [`spoon.SpoonException: Modules are only available since Java 9. Please set appropriate compliance level`](https://github.com/khaes-kth/Sorald-experiments/blob/master/repair-results/argparse4j/sorald_repair-2225.err)

We could set the default compliance level to e.g. 14 to fix this (that's what we're using for Sonar)

slarse:
Fix #339 

I labeled this as a feature, as previously we really did only support Java 8 and below. Now we have support for Java 9+, with modules. I've set the default compliance level to Java 11, as that's what we decided upon for the current paper.

Note that this also sets the Spoon version to the snapshot, as we need that for the module bugfix that went in today (see INRIA/spoon#3771).


khaes-kth:
Thanks @slarse

PR #363, opened 2021-02-04
fix: Rework type collector into compilation unit collector

### LINKED ISSUE #251: Global state in UniqueTypesCollector is causing issues in tests
When execution crashes in a test (because we test for a crash), the state in `UniqueTypesCollector` is not reset in the program, thus causing follow-up failures in subsequent executions depending on test order.

I think we should remove the global state of the `UniqueTypesCollector`, make it a return-value from the repair methods. For now, I'll work around it by resetting the collector _before_ execution, as opposed to after as it currently is.

### LINKED ISSUE #348: Sorald throws (and catches) exceptions during repair
@slarse I think you are aware of most of these, but here is a list of common types of crashes that we face:

[Crash in sorald.processor.IteratorNextExceptionProcessor::process](https://raw.githubusercontent.com/khaes-kth/Sorald-experiments/master/repair-results/joshua/repair-2272.json)

[Crash in sorald.processor.CastArithmeticOperandProcessor::process](https://raw.githubusercontent.com/khaes-kth/Sorald-experiments/master/repair-results/jMetal/repair-2184.json)

[Crash in sorald.processor.InterruptedExceptionProcessor::process](https://github.com/khaes-kth/Sorald-experiments/blob/master/repair-results/aesh/repair-2142.json)

[Crash in sorald.processor.UnclosedResourcesProcessor::process](https://github.com/khaes-kth/Sorald-experiments/blob/master/repair-results/aesh/repair-2095.json)

slarse:
Fix #251 
Fix #348 

The `UniqueTypesCollector` has three problems: 1) global state (#251), 2) collecting types instead of compilation units (which is the unit of granularity that we actually print), and 3) that it's simply incorrectly implemented (#348).

This PR entirely replaces the `UniqueTypesCollector` with the `CompilationUnitCollector`. It solves 1) by making it an event handler which is passed in, no more global state! It solves 2) by simply collecting compilation units, which is in fact a whole lot easier. It solves 3) by hopefully being correctly implemented :)

Note that there's still a slight problem with the collector: it will crash if whatever is passed into it is on a higher level than a type (e.g. a package or a module). We don't have any processors that deal with such high-level elements, so I'll simply fix that separately later.

### DIFF
--- a/src/main/java/sorald/CompilationUnitCollector.java
+++ b/src/main/java/sorald/CompilationUnitCollector.java
@@ -12,18 +12,18 @@ import spoon.reflect.declaration.CtElement;
 import spoon.reflect.declaration.CtType;
 
 /* Only add the CtType object if it does not exist in the map yet */
-public class UniqueTypesCollector {
-    private static UniqueTypesCollector uniqueTypesCollector;
+public class CompilationUnitCollector {
+    private static CompilationUnitCollector uniqueTypesCollector;
 
     private final Map<Path, CtCompilationUnit> pathToCu;
     private final SoraldConfig config;
 
-    private UniqueTypesCollector(SoraldConfig config) {
+    private CompilationUnitCollector(SoraldConfig config) {
         this.config = config;
         pathToCu = new HashMap<>();
     }
 
-    public static UniqueTypesCollector getInstance() {
+    public static CompilationUnitCollector getInstance() {
         if (uniqueTypesCollector == null) {
             throw new IllegalStateException("must call reset first");
         }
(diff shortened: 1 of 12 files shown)


fermadeiral:
@slarse, since this PR fixes one issue that we talked about before, is it ok if I review it?

khaes-kth:
@fermadeiral 

I had a look and I do not see any special problem with the PR. That would be good if you have a look as well and merge it if you see no problem.

fermadeiral:
@khaes-kth LGTM too!

slarse:
@fermadeiral I was certain I had assigned you to review this as well. My intention was for both you and @khaes-kth to review it as it's a rather significant change, but apparently that worked itself out on its own :)

PR #373, opened 2021-02-08
fix: fix a NPE in CastArithmeticOperandProcessor

fermadeiral:
This PR partially addresses #125.

The thing is: when we cannot resolve the type of the left or right hand operand in a binary operator, we get a `NullPointerException`. This PR adds a test case that triggers that exception and the fix for it.

### DIFF
--- a/src/test/resources/processor_test_files/2184_CastArithmeticOperand/NOCOMPILE_NullOperandType.java
+++ b/src/test/resources/processor_test_files/2184_CastArithmeticOperand/NOCOMPILE_NullOperandType.java
@@ -4,5 +4,6 @@ This test checks that such an exception doesn't happen.
 */
 class NOCOMPILE_NullOperandType {
     float twoThirds = 2/3; // Noncompliant; THIS LINE WAS ONLY ADDED TO MAKE THE TEST FIND A VIOLATION IN THIS FILE SO THE NEXT LINE CAN BE ACTUALLY TESTED!
-    long var = System.currentTimeMillis() - message.getJMSTimestamp();
+    long rhsUnknown = System.currentTimeMillis() - message.getJMSTimestamp();
+    long lhsUnknown = message.getJMSTimestamp() - System.currentTimeMillis();
 }
--- a/src/test/resources/processor_test_files/2184_CastArithmeticOperand/NOCOMPILE_NullOperandType.java
+++ b/src/test/resources/processor_test_files/2184_CastArithmeticOperand/NOCOMPILE_NullOperandType.java
@@ -4,7 +4,7 @@ This test checks that such an exception doesn't happen.
 */
 import javax.jms.Message;
 
-class NullOperandType {
+class NOCOMPILE_NullOperandType {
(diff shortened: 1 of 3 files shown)


fermadeiral:
@slarse, I'm not able to create a failing test case easily. I mean, I have it, and it works for the purpose of testing the processor in question, but then it make other tests fail. I think it's the same problem as the other PR I opened these days and you explained me how to fix it. However, I don't really see a way to fix this situation now in a minimal way.

This bug is happening a lot. Is it ok if I just open a PR with the fix for it without test? I feel that soon the data for the paper will need to be consolidated and I would like to have the fix for this problem incorporated in the next round of experiments.

slarse:
@fermadeiral This is my bad.

You're absolutely right that there are test failures because the file doesn't compile. However, here we don't _want_ the file to compile, that's the entire test. So the problem here is that I've not told anyone, nor documented, how to make that work. Oops.

**Solution:** Just prefix the test source file with `NOCOMPILE_`, and it will be ignored by tests that require compilation. **Important:** The class in the test file also needs to be prefixed with `NOCOMPILE_`, as per normal Java naming rules of files and classes (I forget that all the time, so I thought I'd mention it).

fermadeiral:
@slarse I am happy that there is a solution for that! I will do it!

fermadeiral:
@slarse, ready for review!

src/test/resources/processor_test_files/2184_CastArithmeticOperand/NOCOMPILE_NullOperandType.java
```diff
@@ -0,0 +1,8 @@
+/*
+When we cannot resolve the type of the left or right hand operands in a binary operator, we get a NullPointerException.
+This test checks that such an exception doesn't happen.
+*/
+class NOCOMPILE_NullOperandType {
+    float twoThirds = 2/3; // Noncompliant; THIS LINE WAS ONLY ADDED TO MAKE THE TEST FIND A VIOLATION IN THIS FILE SO THE NEXT LINE CAN BE ACTUALLY TESTED!
+    long var = System.currentTimeMillis() - message.getJMSTimestamp();
```
slarse, line 7: I suggest adding the "crashy statement" in the opposite configuration as well, so we're confident that the fix works wherever the unknown type appears in the binary operator.

```suggestion
    long rhsUnknown = System.currentTimeMillis() - message.getJMSTimestamp();
    long lhsUnknown = message.getJMSTimestamp() - System.currentTimeMillis();
```

slarse (changes requested):
Fix looks good, I only have a single suggestion for the test to test "both sides" of a binary operator. Currently, the test would pass even if only the right operand's type is checked.

src/test/resources/processor_test_files/2184_CastArithmeticOperand/NOCOMPILE_NullOperandType.java
fermadeiral: Done!

slarse (approved):

slarse:
Great PR, I'm totally satisfied. Merging!

PR #375, opened 2021-02-09
fix: fix another IndexOutOfBoundsException in CastArithmeticOperandProcessor

fermadeiral:
This PR partially addresses #125.

When a binary operator is in a method call that can receive different number of parameters (e.g., `void info(String var1, Object... var2);`), we can get an `IndexOutOfBoundsException` when trying to find the type expected for the argument (i.e., the binary operator in this case). This PR adds a test case that triggers that exception and the fix for it.

### DIFF
--- a/src/main/java/sorald/processor/CastArithmeticOperandProcessor.java
+++ b/src/main/java/sorald/processor/CastArithmeticOperandProcessor.java
@@ -60,40 +60,7 @@ public class CastArithmeticOperandProcessor extends SoraldAbstractProcessor<CtBi
         CtTypeReference ctTypeReference = null;
 
         if (ctBinaryOperator.getParent(CtAbstractInvocation.class) != null) {
-
-            CtAbstractInvocation ctAbstractInvocation =
-                    ctBinaryOperator.getParent(CtAbstractInvocation.class);
-            List<CtExpression> arguments = ctAbstractInvocation.getArguments();
-
-            CtElement argToBeFound = ctBinaryOperator;
-            while (argToBeFound.getParent() != ctAbstractInvocation) {
-                argToBeFound = argToBeFound.getParent();
-            }
-
-            int indexInInvocation = -1;
-            for (int i = 0; i < arguments.size(); i++) {
-                if (arguments.get(i) == argToBeFound) {
-                    indexInInvocation = i;
-                    break;
-                }
-            }
-
-            CtExecutableReference ctExecutableReference = ctAbstractInvocation.getExecutable();
-            if (ctExecutableReference != null && ctExecutableReference.getParameters() != null) {
-                List<CtTypeReference> parameters = ctExecutableReference.getParameters();
-
-                if (indexInInvocation > parameters.size()
-                        && parameters.get(parameters.size() - 1).isArray()) {
-                    ctTypeReference =
-                            ((CtArrayTypeReferenceImpl) parameters.get(parameters.size() - 1))
-                                    .getComponentType();
-                } else {
-                    ctTypeReference =
-                            (CtTypeReference)
-                                    ctExecutableReference.getParameters().get(indexInInvocation);
-                }
-            }
-
+            ctTypeReference = getExpectedTypeOfMethodCallArgument(ctBinaryOperator);
         } else if (ctBinaryOperator.getParent(CtField.class) != null
                 || ctBinaryOperator.getParent(CtLocalVariable.class) != null) {
             CtField ctField = ctBinaryOperator.getParent(CtField.class);
@@ -109,6 +76,44 @@ public class CastArithmeticOperandProcessor extends SoraldAbstractProcessor<CtBi
             ctTypeReference = ctReturn.getParent(CtMethod.class).getType();
         }
 
+        return ctTypeReference;
+    }
+
+    private CtTypeReference getExpectedTypeOfMethodCallArgument(CtBinaryOperator ctBinaryOperator) {
+        CtTypeReference ctTypeReference = null;
+
+        CtAbstractInvocation ctAbstractInvocation =
+                ctBinaryOperator.getParent(CtAbstractInvocation.class);
+        List<CtExpression> arguments = ctAbstractInvocation.getArguments();
+
+        CtElement argToBeFound = ctBinaryOperator;
+        while (argToBeFound.getParent() != ctAbstractInvocation) {
+            argToBeFound = argToBeFound.getParent();
+        }
+
+        int indexInInvocation = -1;
+        for (int i = 0; i < arguments.size(); i++) {
+            if (arguments.get(i) == argToBeFound) {
+                indexInInvocation = i;
+                break;
+            }
+        }
+
+        CtExecutableReference ctExecutableReference = ctAbstractInvocation.getExecutable();
+        if (ctExecutableReference != null && ctExecutableReference.getParameters() != null) {
+            List<CtTypeReference> parameters = ctExecutableReference.getParameters();
+
+            if (indexInInvocation > parameters.size()
+                    && parameters.get(parameters.size() - 1).isArray()) {
+                ctTypeReference =
+                        ((CtArrayTypeReferenceImpl) parameters.get(parameters.size() - 1))
+                                .getComponentType();
+            } else {
+                ctTypeReference =
+                        (CtTypeReference)
+                                ctExecutableReference.getParameters().get(indexInInvocation);
+            }
+        }
         return ctTypeReference;
     }
 
--- a/src/main/java/sorald/processor/CastArithmeticOperandProcessor.java
+++ b/src/main/java/sorald/processor/CastArithmeticOperandProcessor.java
@@ -11,6 +11,7 @@ import spoon.reflect.declaration.CtMethod;
 import spoon.reflect.reference.CtExecutableReference;
 import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.visitor.filter.TypeFilter;
+import spoon.support.reflect.reference.CtArrayTypeReferenceImpl;
 
 @ProcessorAnnotation(key = 2184, description = "Math operands should be cast before assignment")
 public class CastArithmeticOperandProcessor extends SoraldAbstractProcessor<CtBinaryOperator> {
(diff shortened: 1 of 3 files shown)


src/test/resources/processor_test_files/2184_CastArithmeticOperand/NOCOMPILE_BinaryOperatorInMethodCallThatCanReceiveDifferentNumberOfParameters.java
```diff
@@ -0,0 +1,19 @@
+/*
+When a binary operator is in a method call that can receive different number of parameters (e.g., `void info(String var1, Object... var2);`),
+we can get an IndexOutOfBoundsException when trying to find the type expected for the argument (i.e., the binary operator in this case).
+This test checks that such an exception doesn't happen.
```
slarse, line 4: These comments you put in are excellent. We should document in the contributing guidelines that _this_ is how to document a targeted test case like this.

src/test/resources/processor_test_files/2184_CastArithmeticOperand/NOCOMPILE_BinaryOperatorInMethodCallThatCanReceiveDifferentNumberOfParameters.java
```diff
@@ -0,0 +1,19 @@
...
+import org.slf4j.LoggerFactory;
+
+class NOCOMPILE_BinaryOperatorInMethodCallThatCanReceiveDifferentNumberOfParameters {
+    float twoThirds = 2/3; // Noncompliant; THIS LINE WAS ONLY ADDED TO MAKE THE TEST FIND A VIOLATION IN THIS FILE SO THE NEXT LINES CAN BE ACTUALLY TESTED!
+
+    Logger LOGGER = LoggerFactory.getLogger(getClass());
+
+    void method() {
+        LOGGER.info("{} {}", 1, 1 + 1); // This is ok, see the method signature: void info(String var1, Object var2, Object var3);
+        LOGGER.info("{} {} {}", 1, 2, 1 + 2); // This can cause an IndexOutOfBoundsException, see the method signature: void info(String var1, Object... var2);
```
slarse, line 17: Here, I think it would be better to define the methods you call directly in the test file. This is for three reasons.

1. `slf4j` is not actually a dependency of Sorald, it's a transitive dependency through Spoon and Sonar. If Spoon/Sonar were to remove it, this test case would suddenly fail as the type can't be resolved.
2. If `slf4j` changes either signature (it's moving to 2.0, so it could happen), the test case may break.
3. It would make the comments with the signatures redundant, as the declarations would be right there.

But this works now, so I think we can leave it for later.

src/main/java/sorald/processor/CastArithmeticOperandProcessor.java
```diff
@@ -79,9 +80,18 @@ private CtTypeReference getExpectedType(CtBinaryOperator ctBinaryOperator) {
...
+                if (indexInInvocation > parameters.size()
+                        && parameters.get(parameters.size() - 1).isArray()) {
+                    ctTypeReference =
+                            ((CtArrayTypeReferenceImpl) parameters.get(parameters.size() - 1))
+                                    .getComponentType();
+                } else {
+                    ctTypeReference =
+                            (CtTypeReference)
+                                    ctExecutableReference.getParameters().get(indexInInvocation);
+                }
```
slarse, line 94: This could be a method in its own right, would make the `getExpectedType` easier to read.

slarse (approved):
Coverage hit is due to this expression never evaluating to false in the tests, which I think is fine: https://codecov.io/gh/SpoonLabs/sorald/pull/375/diff#D1-86

The fix looks correct to me. I've left a couple of comments on the diff, but I think that in the interest of time **we should merge now and refactor later**. Agree? If so I'll just open issues for it and then merge this.

fermadeiral:
@slarse, I would prefer to fix your two last comments in this PR. Is it ok to wait for my updates until tomorrow morning?

slarse:
@fermadeiral That's fine by me, I doubt we'll be running the next round of experiments before tomorrow afternoon as I have a couple of Spoon bugs to look at first.

src/test/resources/processor_test_files/2184_CastArithmeticOperand/NOCOMPILE_BinaryOperatorInMethodCallThatCanReceiveDifferentNumberOfParameters.java
fermadeiral: Done.

src/main/java/sorald/processor/CastArithmeticOperandProcessor.java
fermadeiral: Done.

fermadeiral:
@slarse, it's ready for review again.

fermadeiral:
@slarse, no, it's not. Wait a moment, I think I introduced another problem with this PR.

fermadeiral:
@slarse, the changes in this PR triggers a different `IndexOutOfBoundsException`, but apparently this is on Spoon. I'm a little bit in a rush and I really don't want to discuss this by text. Do you have some time to discuss about it over Skype tomorrow after 2pm?

slarse:
@fermadeiral Sure, 2pm is good. Send me a meeting invite so I don't forget :)

fermadeiral:
@slarse done! See you tomorrow.

slarse:
@fermadeiral I've managed to drastically reduce the complexity of the `CastArithmeticOperandProcessor`, while still passing all existing tests. The two key improvements are:

1. Passing the `RuleViolation` to `canRepair` and `repair`, with the message available.
2. Improving the matching heuristic such that it a) only passes elements that are very close to the violation to `canRepair`, and b) ranks candidate matches between elements and violations by how large a fraction of the element intersects with the violation.

The improvements to the matching heuristic means that, in general, elements passed to `canRepair` _are_ violating the rule (there may be exceptions to this that we'll have to discover through experimentation). Therefore, only incomplete processors need to implement `canRepairInternal`, whereas complete processors can simply return `true` (we can have a default implementation for that).

Below is the simplified `CastArithmeticOperandProcessor`.

```java
package sorald.processor;

import java.util.List;
import java.util.Map;
import java.util.regex.Matcher;
import java.util.regex.Pattern;
import java.util.stream.Collectors;
import java.util.stream.Stream;

import org.apache.commons.lang3.tuple.Pair;
import sorald.Constants;
import sorald.annotations.ProcessorAnnotation;
import sorald.sonar.RuleViolation;
import spoon.reflect.code.*;
import spoon.reflect.factory.TypeFactory;
import spoon.reflect.reference.CtTypeReference;

@ProcessorAnnotation(key = 2184, description = "Math operands should be cast before assignment")
public class CastArithmeticOperandProcessor extends SoraldAbstractProcessor<CtBinaryOperator> {

    private static final Map<String, BinaryOperatorKind> toOpKind =
            Map.of(
                    "division", BinaryOperatorKind.DIV,
                    "addition", BinaryOperatorKind.PLUS,
                    "subtraction", BinaryOperatorKind.MINUS,
                    "multiplication", BinaryOperatorKind.MUL);

    @Override
    protected boolean canRepairInternal(CtBinaryOperator candidate, RuleViolation violation) {
        return true;
    }

    @Override
    protected void repairInternal(CtBinaryOperator element, RuleViolation violation) {
        CtTypeReference<?> typeToBeUsedToCast = getOpKindAndType(violation).getRight();
        CtExpression<?> lhs = element.getLeftHandOperand();
        CtExpression<?> rhs = element.getRightHandOperand();

        if (isIntLiteral(lhs)) {
            repairWithLiteralSuffix((CtLiteral<?>) lhs, typeToBeUsedToCast);
        } else if (isIntLiteral(rhs)) {
            repairWithLiteralSuffix((CtLiteral<?>) rhs, typeToBeUsedToCast);
        } else {
            repairWithCast(element, typeToBeUsedToCast);
        }
    }

    private static void repairWithCast(
            CtBinaryOperator element, CtTypeReference<?> typeToBeUsedToCast) {
        CtCodeSnippetExpression newBinaryOperator =
                element.getFactory()
                        .createCodeSnippetExpression(
                                "("
                                        + typeToBeUsedToCast.getSimpleName()
                                        + ") "
                                        + element.getLeftHandOperand());
        element.setLeftHandOperand(newBinaryOperator);
        // A nicer code for the casting would be the next line. However, more parentheses are added
        // in
        // the expressions when using such a solution.
        // element.getLeftHandOperand().addTypeCast(typeToBeUsedToCast.clone());
    }

    private static void repairWithLiteralSuffix(
            CtLiteral<?> literalInt, CtTypeReference<?> typeForSuffix) {
        Integer value = (Integer) literalInt.getValue();
        CtCodeSnippetExpression<?> literalWithSuffix =
                literalInt
                        .getFactory()
                        .createCodeSnippetExpression(value + getLiteralSuffix(typeForSuffix));
        literalInt.replace(literalWithSuffix);
    }

    private Pair<BinaryOperatorKind, CtTypeReference<?>> getOpKindAndType(RuleViolation violation) {
        String message = violation.getMessage();
        Pattern p = Pattern.compile(".*?(\\w+)( operation)? to a \"(\\w+)\".*");
        Matcher m = p.matcher(message);

        if (!m.matches()) {
            throw new IllegalStateException(
                    "Message '" + message + "' did not contain expected match");
        }

        String rawOpKind = m.group(1);
        String rawExpectedType = m.group(3);

        TypeFactory tf = getFactory().Type();
        Map<String, CtTypeReference<?>> nameToPrimitiveType =
                Stream.of(tf.floatPrimitiveType(), tf.longPrimitiveType(), tf.doublePrimitiveType())
                        .collect(Collectors.toMap(CtTypeReference::getSimpleName, ref -> ref));

        return Pair.of(toOpKind.get(rawOpKind), nameToPrimitiveType.get(rawExpectedType));
    }

    private static boolean isIntLiteral(CtExpression<?> expr) {
        return expr instanceof CtLiteral
                && expr.getFactory().Type().INTEGER_PRIMITIVE.equals(expr.getType());
    }

    private static String getLiteralSuffix(CtTypeReference<?> floatDoubleOrLong) {
        String simpleName = floatDoubleOrLong.getSimpleName().toLowerCase();
        assert List.of(Constants.FLOAT, Constants.DOUBLE, Constants.LONG).contains(simpleName);
        return Map.of(Constants.FLOAT, "f", Constants.DOUBLE, "d", Constants.LONG, "l")
                .get(simpleName);
    }
}
```

fermadeiral:
@slarse, how should we proceed? Should I chose this PR?

slarse:
@fermadeiral I think we should keep it open until we have a solution implemented and merged. This PR still serves as important documentation of our thought process.

slarse:
Right, this is fixed as of #443. Closing!

PR #378, opened 2021-02-11
feat: Only invoke SoraldAbstractProcessor.canRepair on elements close to violations

### LINKED ISSUE #346: Optimize the GreedyBestFitScanner
It currently scans the entire model, while in reality it only needs to scan the compilation units in which we've detected violations.

slarse:
Fix #346 

Currently, the GreedyBestFitScanner scans the entire model, and invokes `SoraldAbstractProcessor.canRepair` on _all_ elements that have the correct type. This is bad for several reasons:

1. Inefficient, we both scan too much and invoke `canRepair` too often.
2. The vast majority of elements passed to `canRepair` are not close to any violation, or even in the same file. This is an unnecessary source of crashes.
3. It's not clear to implementers of processors exactly what needs to be checked in `canRepair`, as it's invoked all over the place.

This redesign does two things. First, it does not scan type declarations that don't contain any violations. Second, it only passes elements to `canRepair` after verifying that said elements are _very close_ to a considered violation. Currently, we do it the other way around: first call `canRepair`, and then check if the element is close.

The new guarantee that `canRepair/canRepairInternal` are only called on elements that are very close to a relevant violation are also added to the docs.

This isn't really a refactor as it fundamentally affects how Sorald behaves, so I labeled it as a feature instead.

### DIFF
--- a/src/main/java/sorald/processor/SoraldAbstractProcessor.java
+++ b/src/main/java/sorald/processor/SoraldAbstractProcessor.java
@@ -55,6 +55,8 @@ public abstract class SoraldAbstractProcessor<E extends CtElement> extends Abstr
      * example, when repairing something involving a method call, there may be nested calls that are
      * not actually the violating parties, but still appear in the correct vicinity.
      *
+     * <p>This method is only to be called on elements that appear very close to a violation.
+     *
      * <p><b>This method does not mutate the state of the processor</b>.
      *
      * <p>This method never crashes, instead returning false if there is a problem in the concrete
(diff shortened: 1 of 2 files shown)


slarse:
This PR quite drastically reduces coverage as many of the checks performed by processors in `canRepairInternal` are now never utilized, as the elements that triggered these checks weren't actually violations and thus aren't scanned anymore. Not quite sure what the best way to deal with that is.

slarse:
@fermadeiral This somewhat affects the fixes you're trying to implement right now. With this PR, far fewer elements would be scanned with the `CastArithmeticProcessor`, and so far fewer crashes would be reported. Only crashes on actual violating elements, or elements very close to violations, will be a problem after this is merged.

slarse:
I think this should be merged as-is, without addressing the coverage drop on the processors. The reason the coverage drops is that many of the checks are made redundant, as `canRepair` is no longer invoked on _every_ element, but only on the ones that are close to violations.

In my further prototyping, I've improved the heuristic to the point where elements passed to `canRepair` are almost certain to be violating elements, which will render `canRepair` implementations redundant for most processors (not for the incomplete ones, though). I think further improving the matching heuristic is a better idea than having intricate `canRepairInternal` implementations for each and every processor.

fermadeiral:
> Only crashes on actual violating elements, or elements very close to violations, will be a problem after this is merged.

That is great! Thank you!

PR #385, opened 2021-02-16
feat(2142): Add thread interrupt before first throw or return

### LINKED ISSUE #384: InterruptedExceptionProcessor should add the interrupt to the start of the catch
Currently, `InterruptedExceptionProcessor` adds the interrupt to the end of the catch. This can cause issues, for example in the patch I show in #382 .

```diff
diff --git a/src/java/com/verificatum/protocol/mixnet/ShufflerElGamalSession.java b/src/java/com/verificatum/protocol/mixnet/ShufflerElGamalSession.java
index 05aa99d..48183b3 100644
--- a/src/java/com/verificatum/protocol/mixnet/ShufflerElGamalSession.java
+++ b/src/java/com/verificatum/protocol/mixnet/ShufflerElGamalSession.java
@@ -25,6 +25,8 @@
  */
 
 package com.verificatum.protocol.mixnet;
+import static com.verificatum.protocol.elgamal.ProtocolElGamal.getCiphPGroup;
+import static com.verificatum.protocol.elgamal.ProtocolElGamal.getWidePublicKey;
 
 import java.io.File;
 
@@ -939,6 +941,7 @@ public final class ShufflerElGamalSession extends ProtocolElGamal {
                 nextOutputThread.join();
             } catch (final InterruptedException ie) {
                 throw new ProtocolError("Unable to join threads!", ie);
+                Thread.currentThread().interrupt();
             }
             nextOutputThread = null;
         }
```

Here, there's a throw in the catch statement, and `Thread.currentThread().interrupt()` is added _after_ the throw. This makes the interrupt unreachable, and this won't even compile. It would be a much safer bet to add the interrupt at the start of the catch.

slarse:
Fix #384 

This PR modifies the processor for rule 2142 to place the interrupt as late as possible in the catch block, but before any throw or return statement. See #384 for why this is necessary.

### DIFF
--- a/src/test/resources/processor_test_files/2142_InterruptedException/InterruptedExceptionForTesting.java.expected
+++ b/src/test/resources/processor_test_files/2142_InterruptedException/InterruptedExceptionForTesting.java.expected
@@ -21,8 +21,8 @@ class InterruptedExceptionForTesting {
 				throw new InterruptedException();
 			}
 		}catch (InterruptedException e) { // Noncompliant; logging is not enough
-		    Thread.currentThread().interrupt();
 			LOGGER.log(Level.WARN, "Interrupted!", e);
+			Thread.currentThread().interrupt();
 		}
 	}
 
@@ -85,10 +85,10 @@ class InterruptedExceptionForTesting {
 				throw new InterruptedException();
 			}
 		} catch (InterruptedException e) { // Noncompliant {{Either re-interrupt this method or rethrow the "InterruptedException".}}
-			Thread.currentThread().interrupt();
 			LOGGER.log(Level.WARN, "Interrupted!", e);
 			// clean up state...
 			new Foo().interrupt();
+			Thread.currentThread().interrupt();
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


slarse:
Actually, I need to think about this a bit more

src/test/resources/processor_test_files/2142_InterruptedException/InterruptedExceptionForTesting.java.expected
```diff
@@ -0,0 +1,116 @@
...
+				// do stuff
+				throw new InterruptedException();
+			}
+		} catch (InterruptedException e) {
+			LOGGER.log(Level.WARN, "Interrupted!", e);
+			// clean up state...
+			throw e;
+		} catch (ThreadDeath threadDeath) { // Noncompliant {{Either re-interrupt this method or rethrow the "ThreadDeath".}}
+			Thread.currentThread().interrupt();
+			throw new IOException();
```
slarse, line 67: Here's an example of where the interrupt must come at the start of the block

src/test/resources/processor_test_files/2142_InterruptedException/InterruptedExceptionForTesting.java.expected
```diff
@@ -0,0 +1,116 @@
...
+		try {
+			while (true) {
+				// do stuff
+				throw new InterruptedException();
+			}
+		} catch (InterruptedException e) { // Noncompliant {{Either re-interrupt this method or rethrow the "InterruptedException".}}
+			Thread.currentThread().interrupt();
+			LOGGER.log(Level.WARN, "Interrupted!", e);
+			// clean up state...
+			new Foo().interrupt();
```
slarse, line 91: Here's an example of where it looks quite funky

slarse:
I think we'll want to place the interrupt as late as possible in the catch block. We scan each statement in the statement list and see if it contains a return or a throw, and if it does we place the interrupt before that statement (otherwise last).

slarse:
@fermadeiral Ready for review now.

fermadeiral:
@slarse thank you!

PR #387, opened 2021-02-16
feat: Improve Spoon element to Sonar violation position matching

### LINKED ISSUE #383: Further improve the position matching between Spoon elements and Sonar violations
We should further improve the position matching between Spoon elements and Sonar violations to get better results. I propose a "ranking" of position matchings according to how large a fraction of the element a given violation position encompasses. First, some terminology:

* Source position: A source position given by character start/end. E.g. "starts at char 32 and ends at char 48". Note that lines and columns are only for readability, the source position can be fully specified in one dimension.
* Element position: The source position of a Spoon element.
* Violation position: The source position of a Sonar violation.
* Position size: `source_pos.end - source_pos.start`

Now, assume that we have two Spoon elements `e1` and `e2` and a violation `v`, and we want to determine which of `e1` and `e2` is the best position match for `v`. In short, we do this by picking the element with the largest fractional intersection with `v`. If both elements have the exact same intersection fraction (typically only happens for 0% and 100% intersection), we pick the element with the largest source position, as the absolute value of the intersection is then larger.

In pseudocode, it looks like so:

```python
def best_position_match(e1: CtElement, e2: CtElement, v: RuleViolation) -> CtElement:
    e1_intersect = intersect_fraction(e1, v)
    e2_intersect = intersect_fraction(e2, v)
    if abs(e1_intersect - e2_intersect) < INTERSECT_TOLERANCE:
        return e1 if e1.source_pos.size >= e2.source_pos.size else e2
    else:
        return e1 if e1_intersect >= e2_intersect else e2

def intrersect_fraction(e: CtElement, v: RuleViolation) -> double:
    return v.source_pos.intersection(e.source_pos).size / e.source_pos.size
```

There are two subtle but important details here: 1) `intersect_fraction` being normalized by Spoon element size, and  2) that we pick the larger element if the intersect fraction is equal.

### Normalizing intersection by element size
Note that we normalize the intersection size by the size of the Spoon element. This is an important detail for nestable elements, as it will prioritize "tighter fits". For example, consider the nested class `Serial` that violates [rule 2057](https://rules.sonarsource.com/java/RSPEC-2057) (also see #368):

```java
import java.io.Serializable;

public class NonSerial {
    public static class Serial implements Serializable {
    }
}
```

The violation position will be contained inside the header of `Serial`. As all of `Serial` is contained in `NonSerial`, the violation position is fully encompassed by both classes. However, `Serial` is a smaller element (in terms of source position size), and so the intersection fraction calculated by `intersect_fraction` will be slightly larger. I.e., it's a "tighter fit". As nested elements always have a smaller source position than their encompassing elements, this should be a pretty good heuristic.

It _might_ backfire depending on how Sonar provides positions for specific elements, but we have `canRepair` as a fallback for that.

### Picking the larger element when intersection fractions are (roughly) equal
In general, the intersection fraction of `e1` and `e2` can only be equal if `intersect_fraction(e1) == intersect_fraction(e2) == 1` (i.e. violation encompasses both elements) or `intersect_fraction(e1) == intersect_fraction(e2) == 0` (i.e. neither element intersects). I'm not sure the latter can happen, but the former very much can with nestable expressions. For example, consider this expression that violates [rule 2164](https://rules.sonarsource.com/java/RSPEC-2164).

```java
double d = 1.0F + 2.0F / 3.0F;
```

The violation will be for the entire expression `1.0F + 2.0F / 3.0F`, which consists of two binary operators `plus(1.0F, div(2.0F, 3.0F))`. Now, the intersection fraction for the entire expression is 1.0, and so is the intersection fraction for the subexpression `2.0F / 3.0F`. Picking the entire expression is a better fit than picking the subexpression, and so we pick the _larger_ element this time.

This heuristic is potentially flawed due to position mismatching: if the larger element's intersection fraction is 0.999999 and the smaller element's fraction is 1.0, we'll still pick the smaller element. So, we use a threshold value `INTERSECT_TOLERANCE` for comparison, but I've no idea what a good value for it would be.

slarse:
Fix #383 

Implements improved position matching as described in #383. The only complete processor for which we can't remove the `canRepairInternal` method is `MathOnFloatProcessor`, but that's because it's incorrectly implemented, and not because the matching makes a mistake. It's quite easy to break the current implementation even while retaining the `canRepairInternal` implementation. I'll open a separate issue for that.

I've benchmarked this with the benchmark in #404, and there's no degradation. I'll add in the benchmark results once #404 is merged. **Merge #404 before this one, and then ping me**.

This PR can still be reviewed in the mean time, as the code is all there.

@khaes-kth @fermadeiral Whichever one of you has time can review this.


slarse:
Currently, the only processor that fails tests when removing `canRepairInternal` is `MathOnFloatProcessor`. But that's not because the wrong element is provided to it, it's because its `repairInternal` implementation is not correct (it's easy to break it).

slarse:
I want to setup a benchmark before merging this, so we can judge if it's at least as good as before.

fermadeiral:
#404 merged.

slarse:
> #404 merged.

Thanks! I ran the benchmark again, and we can see some improvements across the board. The ratios improved as follows:

```
Old total_repair_ratio: 0.8226181254841208
New total_repair_ratio: 0.8641141141141141
total_repair_ratio is unchanged or improved

Old attempted_repair_ratio: 0.9178910976663786
New attempted_repair_ratio: 0.9282258064516129
attempted_repair_ratio is unchanged or improved
```

For the precise updates, see 614d35d.

I'm happy with this PR now, ready for merge in my opinion!

fermadeiral:
@slarse thanks.

PR #392, opened 2021-02-19
feat(2184): Use upper case letter for literal suffixes

### LINKED ISSUE #379: Literal suffixes for double, long and float should use uppercase letter
This is required in some cases, see https://github.com/AddstarMC/Prism-Bukkit/pull/247/files

slarse:
Fix #379 

There's no test to verify that this is the case, as we currently don't have support for testing literal output in the test suite, but I've verified it manually. I've opened an issue about augmenting the test suite to allow for testing that arbitrary text is present in the output file, see #393.


fermadeiral:
LGTM.

PR #391, opened 2021-02-19
experiment: Add complete record for spoonlabs/sorald PR 381

slarse:
Adds the complete record for spoonlabs/sorald#381 to prs.json.

This includes the initial data, the "manual" edit (autoformatter removed an import statement), and the final data.

**Merge #394 before this one**

PR #394, opened 2021-02-19
chore: Use github-actions bot user as committer for ACHIEVEMENTS update

slarse:
I'm not sure if anyone's noticed, but we have a `GitHub Web Flow` contributor. That's because of this commit: https://github.com/SpoonLabs/sorald/commit/cfc009279b03b63cb2ceca0ad355fcc55dc525fa

Which lists it as a co-author for the ACHIEVEMENTS file update. That's automatically put there because the auto-updater commits with that user.

This PR makes the support workflow commit with the github-actions bot user instead. It will still generate a "co-authored by", but it will be with an alias of the actions bot, which does not count as a contributor.

### DIFF
--- a/.github/workflows/support.yml
+++ b/.github/workflows/support.yml
@@ -26,8 +26,8 @@ jobs:
         run: python -m sorald.achievements -p experimentation/prs.json -o docs/ACHIEVEMENTS.md
       - name: Create pull request
         run: |
-          git config --local user.email noreply@github.com
-          git config --local user.name GitHub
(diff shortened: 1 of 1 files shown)


slarse:
@fermadeiral @khaes-kth Could either of you have a quick look at this and merge it, and then merge #391 so we can see that it worked? I tested it on my fork and it appears to work as expected, but I wanna see it work in "the correct place" as well :)

fermadeiral:
LGTM, Merging...

PR #401, opened 2021-02-23
feat(1854): Fix cases of local variable deletion in dead store elimination

### LINKED ISSUE #400: DeadStoreProcessor (rule 1854) carelessly deletes local variables that are used later
Currently, the `DeadStoreProcessor` which is supposed to repair dead stores has a serious bug: it just deletes any element flagged as a violation. That causes problems if the element that's deleted is a variable initialization, and the variable is used later in the program. For example:

```java
int a = 2; // Noncompliant
a = 3;
```

is repaired by deleting `int a = 2;`, but leaves `a = 3`, which makes for a program that does not compile.

The _simple_ way to fix this would be to only delete the initializer (i.e. repair to this):

```java
int a;
a = 3;
```

The _nicer_ way to fix this would be to move the initialization into as tight a scope as possible. Here, it would mean this:

```java
int a = 3;
```

There are some corner cases where this becomes difficult, but in general it should be doable.

slarse:
Fix #400 

Fixes the problem described in #400 with a pretty cool mechanism for moving the declaration into appropriate scope. See [the updated handled rules](https://github.com/slarse/sorald/blob/issue/400-fix-dead-store-delete-local-variable/docs/HANDLED_RULES.md#unused-assignments-should-be-removed-sonar-rule-1854) for a concise description of the new mechanism.

Note: There are still cases where this may crash and burn if there are multiple dead stores in a row in the same variable. That can cause issues, but I'll leave that for a later PR. We need some mechanism to indicate which elements have been repaired, so we don't attempt to enact repairs on already modified elements.

And yes, I spent way too much time on this, but it was a super fun problem to solve :)


slarse:
@khaes-kth @fermadeiral I need this merged for the demo tomorrow, so if either of you have time to review anything before then please prioritize this PR.

fermadeiral:
LGTM.

@slarse, sorry for the delay, last week was crazy and I didn't notice this PR.

slarse:
That's OK, it's in for the demo so we're all good :)

PR #404, opened 2021-02-25
experiment: Add benchmark script

### LINKED ISSUE #389: Need a benchmark
Our test suite is pretty good but it's far from sufficient to assess sweeping changes. For example, right now, I'm working on a (supposedly) much improved version of the spoon/sonar matching heuristic (see #387). It passes our tests, but so did the prior version, and so I'd like to benchmark this to see if there's any difference.

@nharrand How involved is the experiment you're running for the paper? Would it be feasible to automate it such that it runs, extracts some data (most importantly, `num_attempted_repairs / num_violations` and `num_successful_repairs / num_attempted_repairs`), and the presents that in sufficiently readable way? Or would it be easier for me to start from scratch with a new benchmark?

slarse:
Fix #389 

Adds a benchmark script along with the current benchmark results for Sorald.

Don't worry about the massive diff: the `benchmark_baseline.csv` file accounts for almost 800 of those lines.


slarse:
It turned out to be a wee bit difficult to get this entirely stable. I think I've managed now, but it takes a while to run my current benchmark and so we'll probably have to wait until Monday to merge this.

Issue #166, opened 2020-10-23
Use Path/File for paths and files instead of String


slarse:
Throughout sorald, there is heavy use of raw `String` values to represent paths. This is really hard to deal with consistently, especially when relative paths start getting mixed with absolute paths. Simple operations such as checking if a path exists, is a file or directory, or resolving a new path relative to some other path, are needlessly tedious and error-prone.

We should replace all raw `String` values with either `Path` and/or `File` values.

PR #406, opened 2021-03-01
refactor: Make Constants.PATH_TO_RESOURCES_FOLDER an actual Path

slarse:
#166 

First PR to start using Path/File instead of String to represent paths. Title says it all about the changes in this PR.

### DIFF
--- a/src/main/java/sorald/Constants.java
+++ b/src/main/java/sorald/Constants.java
@@ -1,6 +1,8 @@
 package sorald;
 
 import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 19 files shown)


khaes-kth:
Great!

@slarse: do you think it is now ready to be used on windows or should we make more changes before that? I can give it a try on a number of projects if you think it is ready.

slarse:
> Great!
> 
> @slarse: do you think it is now ready to be used on windows or should we make more changes before that? I can give it a try on a number of projects if you think it is ready.

Don't think so yet, there are still a number of `String` usages I need to remove

khaes-kth:
> Don't think so yet, there are still a number of `String` usages I need to remove

Good, we'll wait until they are removed as well.

Issue #409, opened 2021-03-03
Removing little used or redundant features (multi-rule repair, out-of-place printing and Git patch generation)


slarse:
I broached this topic during the last meeting and thought I'd put it in text. There are two features in Sorald that don't seem very useful to me, yet significantly complicate the architecture of the tool and implementation of new features.

First is the ability to repair multiple rule keys in one invocation. This requires us print-and-reparse in between each repair to avoid different rules (that might overlap) from interfering with each other. This doesn't work at all in MAVEN mode, as we don't grab supporting files (e.g. pom files). While we _could_ try to fix all corner cases causing MAVEN mode not to work, it's a not insignificant hassle. It doesn't make sense for targeted repair as source code positions may shift around from one parse to another, causing the violation specifiers to become invalid. It also significantly complicates both #358 and #395. Add to that that I've personally never used multi-rule repair. If one for some reason wants to repair multiple rules, just creating a loop in e.g. a bash script for it is trivial.

Second is the ability to do out-of-place repair (i.e. any mode but IN_PLACE). We can't reliably gather before/after statistics on a project if we print the output to a separate location, and so that's a large feature that just doesn't work. For all our experiments, we've used in-place repair since I implemented it, and I don't see much value in keeping the other modes.

Removing these features would make for a much simpler project, and we can go ahead with some much needed improvements, especially #358. We can also remove the `--file-output-strategy` option from the CLI.

slarse:
@monperrus @khaes-kth @fermadeiral @nharrand thoughts?

monperrus:
+1. In general, I love debloat and simplification.

khaes-kth:
As far as the current paper is concerned, we only use IN_PLACE mode and repair violations of one rule in each execution. So, I'm fine with both suggestions. In the long-term, I think it is better to provide the option for multiple-rule fixes. As you said in the meeting, it can be done by a script that runs Sorald several times.

slarse:
I think we're in agreement to remove these features for the time being. I'm not convinced multi-rule fixes will ever be a feature that users actually want in the tool itself, but if that day comes we can quite easily build such a feature on top of single-rule repair (just run Sorald multiple times internally).

slarse:
I found another completely redundant feature: generating Git patches. With in-place repair, we can just run `git diff` in a Git repo instead.

khaes-kth:
> generating Git patches

[...]

PR #410, opened 2021-03-04
break: Remove Git patch generation

slarse:
#409 

Breaking change that removes the Git patch generation functionality. Since we have in-place repair, this feature is redundant. We can currently achieve roughly the same behavior like so:

```bash
# in some Git repo
$ sorald repair --file-output-strategy IN_PLACE [... options ...]
$ git diff > sorald_repair.patch
```

The only difference is that we don't get one patch file per changed source file, but I really don't see why we'd want that either (and it could be done with some more scripting).


fermadeiral:
@slarse, this feature was implemented by @henry-lp, and before anything else I would like to hear from him.

@henry-lp, if this feature is removed, does something get broken on the Repairnator side?

slarse:
@fermadeiral It does break repairnator (see https://github.com/eclipse/repairnator/blob/0cff4102b6792fe209d6f890589364dfb3bfa014/src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/repair/Sorald.java), but it can be fixed by using in-place repair instead. Right now, Sorald is essentially used like this:

1. Perform patch out-of-place.
2. Generate patches.
3. Apply patches to repo.

That can be cut down to.

1. Perform patch in-place.

There's one extra thing in that Sorald "notifies" something (a listener I presume?) about the individual patch files, of which I don't quite know the purpose.

@henry-lp Feel free to correct me if I'm wrong.

slarse:
Note also that we're removing out-of-place repair as well to make it possible to improve other features. And Repairnator also uses the old camel case version of the CLI. So already, the usage in Repairnator needs to be patched, but it's all easy fixes.

slarse:
@fermadeiral Regardless of what @henry-lp answers, we can defer a decision on this PR until the next meeting (on the 9th).

fermadeiral:
@slarse thanks.

slarse:
@fermadeiral According to @khaes-kth , we don't need the built-in Git patch generation:

https://github.com/SpoonLabs/sorald/issues/409#issuecomment-791405658

fermadeiral:
@slarse, if nothing will be broken on Repairnator, I'm merging.

slarse:
@fermadeiral That's not quite what we meant, Repairnator integration is already broken by the updates to the CLI. We need to patch Repairnator with the new functionality, but it's not hard, and will simplify the code overall. We'll discuss this during the next meeting, it'll be up to myself and @khaes-kth to make it work :)

slarse:
So, to be clear, it _is_ currently used, but it doesn't actually need to be used.

khaes-kth:
> So, to be clear, it _is_ currently used, but it doesn't actually need to be used.

Agree!

fermadeiral:
@slarse thank you for the clarification.

In this case, do we already have an issue somewhere for us to not forget that Repairnator is broken because of several changes in Sorald and so we need to fix it?

slarse:
I think my clarification wasn't that clear. Repairnator, right now, isn't broken. It's just using an older version of Sorald. Once someone updates the version of Sorald in Repairnator, they'll also have to deal with the changes in the CLI. To me, that's just common sense when updating a dependency; check for changes and make sure that it works.

While we _could_ be "nice" and open an issue in Repairnator about it, I don't think it's a good idea to set a precedent that we notify dependents of such changes, as that's something we may forget to do at another time. WDYT?

fermadeiral:
> I think my clarification wasn't that clear.

Indeed, because "Repairnator integration is already broken by the updates to the CLI" is very different from "Repairnator, right now, isn't broken".

But I got your point now, and I agree with you. No issue to be created in Repairnator.

PR #411, opened 2021-03-05
break: Remove multi-rule repair

slarse:
#409 

WIP


slarse:
1ed0159 is a perfect example of why I always try to add at least a sanity check test for support scripts: that script was merged today and I already broke it :)

src/main/java/sorald/SoraldConfig.java
```diff
@@ -2,17 +2,10 @@
...
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
 import java.util.Optional;
-import java.util.stream.Collectors;
-import sorald.sonar.RuleViolation;
 
 /* All config settings of Sorald should be gathered here */
 public class SoraldConfig {
-    private final List<Integer> ruleKeys = new ArrayList<>();
```
fermadeiral, line 14: Why does the rule key not stored in `SoraldConfig`?
I thought that this class was supposed to hold all config values.

slarse: Short answer: because it doesn't need to be there.

The somewhat longer answer: First of all, the rule key isn't necessarily explicitly configured with a command line argument anymore, as we have targeted repair (in which the violation IDs contain rule key). This means that it's not truly a configured value at all times, but rather is resolved at runtime.

Second, and this is a big one, the `SoraldConfig` is a pretty big hassle to work with. It makes testing harder as you need to properly configure a config object to call certain methods, but there's no way to figure out what values need to actually be configured in order to test a method other than by going into the source code and looking for config accesses. It's much easier to just pass the required values directly to constructors or methods. This goes double for when we actually provide an API for Sorald, it would be very unintuitive for users to have to try and figure out what are the required values to set in the config to be able to call some given method.

Depending on how the API turns out in the end, `SoraldConfig` may not make sense to keep at all.

slarse: Here's an example of what I mean:

```java
// strange API
SoraldConfig config = new SoraldConfig();
config.setRuleKey(1854);
config.setOriginalFilesPath(somePath);
new Repair(config).repair();

// more intuitive, much more easily tested API
Repair.repair(1854, somePath);
```

PR #414, opened 2021-03-08
wip: chore: Add macOS build

slarse:
Adds a macOS build. The commit message is a lie :)


slarse:
So, we have some failures here. I think they may all be related to temporary directories. In other projects, I've had trouble with temporary directories on macOS. When you create a temporary directory, it ends up somewhere in `/var`, but `/var` is for some arcane reason symlinked to `/private/var`. So, I've had stuff like this happen:

```java
File tempDir = createTemporaryDirectory(); // /var/some/path

// stuff happens

File fileInTempDir = ...; // /var/some/path/file.txt
File normalizedFileInTempDir = fileInTempDir.normalize(); // /private/var/some/path/file.txt 
if (tempDir.containsFile(fileInTempDir.normalize()) { // false, because /private/var is not in /var

}
```

It might be something like that.

slarse:
Just like with Windows, this will most likely be resolved if we consistently use _normalized_ Path/File objects everywhere.

slarse:
Support script tests are failing due to GitHub rate limiting. We should use the workflow's token to avoid that problem.

slarse:
Argh, fix one thing, break another. Paths are such a bother ...

PR #415, opened 2021-03-08
Update pom.xml

gothius:
Update Maven Javadoc Plugin from version 3.0.0-M1 to version 3.2.0 to fix NullPointerException during build. (for Java openjdk 11.0.10 on MacOS 10.15.7).


slarse:
@gothius That's great, thanks!

Also points out that we should be building the Javadoc in CI, and also have a macOS build (because this NullPointerException does not occur on any of my Linux machines). Working on the macOS build in #414, should stabilize in the coming days.

Issue #359, opened 2021-02-03
Provide rule violation to canRepair and repair


slarse:
I think that in combination with #358 , the `canRepair` methods may have use of the rule violation itself. For example, the message for [rule 1939](https://rules.sonarsource.com/java/RSPEC-1939) contains the name of the redundant implementation/extension, which otherwise would require quite a lot to come by in Spoon itself.

slarse:
I'm going to push this back, because I want #383 done first, and for that I need then benchmark.

slarse:
As rule violations are now available to the `repairInternal` method, I'm currently satisfied. I've no use case for providing rule violations to `canRepairInternal` at the moment.

PR #420, opened 2021-03-15
fix(1854): Avoid merging declaration with another dead store

### LINKED ISSUE #418: DeadStoreProcessor sometimes does not attempt all repairs
When running the DeadStoreProcessor (rule 1854) on verificatum/verificatum-vmn@945b620e4c8228a6ad10169a017d58aac699aa1e, only 8 of 9 violations are repaired. I suspect that this is caused by one violation overwriting another dead store when merging a declaration with the next assignment.

To fix this, we need to check that the assignment we want to merge with is not actually a dead store itself. To be able to check that, the processor must know about the locations of _all_ violations.

slarse:
Fix #418 

This PR fixes a problem where the `DeadStoreProcessor` would merge a declaration with a dead initializer, with _another_ dead store (/facepalm). The fix is pretty simple: don't do that.

Here's an example of how moving the declaration without considering dead stores can go wrong:

```diff
public class MultipleDeadStoresInARow {
    public static void main(String[] args){
-       int a = 2; // Noncompliant
-       a = 3; // Noncompliant
+       int a = 3;
        a = 4;
        System.out.println(a);
    }
} 
```

Note that the initializer for `a` is still a dead store. The problem is that the dead store that's mapped to `a = 3` will simply disappear, as that element is replaced with the declaration.

Here's the same fix after the patch in this PR:

```diff
public class MultipleDeadStoresInARow {
    public static void main(String[] args){
-       int a = 2; // Noncompliant
-       a = 3; // Noncompliant
-       a = 4;
+       int a = 4;
        System.out.println(a);
    }
} 
```

**How does this work?** When repairing the violation on `int a = 2;` by moving the declaration to another store, it simply skips over all dead stores and finds the first non-dead store. The skipped-over dead stores are then deleted by further invocations of the processor, on those particular elements. Note that if there is _no_ non-dead store, the declaration is simply deleted.

**Why this much work for such a simple repair?** This presents a good stepping stone for making Sorald's repairs more sophisticated. It's a simple repair in principle that's complicated by our choice of moving the declaration to create a better-looking patch.

For this to be possible, the `DeadStoreProcessor` must be able to access the best-fits mapping between Spoon elements and rule violations. As it turns out, this was trivial: just add a getter to `SoraldAbstractProcessor`. Therefore, this PR _also_ fixes part of #359, in that `repair` now has access to said mapping. It's not however accessible to `canRepair`, as that method is in fact used to compute the mapping. If we want `canRepair` to have access to the rule violation for an element, a position-matched rule violation and Spoon element pair must be passed in to `canRepair`.

### DIFF
--- a/src/main/java/sorald/processor/DeadStoreProcessor.java
+++ b/src/main/java/sorald/processor/DeadStoreProcessor.java
@@ -57,11 +57,7 @@ public class DeadStoreProcessor extends SoraldAbstractProcessor<CtStatement> {
      */
     private boolean isDeadStore(CtVariableAccess<?> varAccess) {
         CtStatement parentStatement = varAccess.getParent(CtStatement.class);
-        if (parentStatement == null) {
-            return false;
-        } else {
-            return getBestFits().containsKey(parentStatement);
-        }
+        return getBestFits().containsKey(parentStatement);
     }
 
     /**
--- a/src/main/java/sorald/processor/DeadStoreProcessor.java
+++ b/src/main/java/sorald/processor/DeadStoreProcessor.java
@@ -43,11 +43,11 @@ public class DeadStoreProcessor extends SoraldAbstractProcessor<CtStatement> {
      */
     private void retainDeclarationOnVariableUse(CtLocalVariable<?> localVar) {
         CtStatementList statementList = localVar.getParent(CtStatementList.class);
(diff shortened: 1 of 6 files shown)


fermadeiral:
LGTM!

PR #422, opened 2021-03-17
chore: Separate building and testing into different steps

### LINKED ISSUE #423: Clean up CI output
There are thousands of lines of "downloading dependency" stuff in the test output of CI. We can clean that up by splitting building and testing into different steps.

slarse:
Fix #423 

This cleans up the CI output by separating building the project and running tests into separate steps. Thus, the "Run tests" step is not polluted by thousands of lines of "Downloading some dependency".

Previously, the "Build project" step both fetched dependencies and ran tests. Now, the new "Run tests" step only runs tests, which makes it easier to find the test output. Arguably, we should also turn off the logger or set it to warnings only, but that's for another PR.

* [See this build for a "before"](https://github.com/SpoonLabs/sorald/actions/runs/659434600)
* [See this build for an "after"](https://github.com/SpoonLabs/sorald/runs/2128668398)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,6 +32,10 @@ jobs:
         run: |
           mvn -B package -DskipTests
           mvn -B test-compile
+      - name: Fetch final dependencies
+        # this is a hack to fetch some test runtime dependencies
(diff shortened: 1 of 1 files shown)


fermadeiral:
Thank you, Simon.

PR #431, opened 2021-03-18
wip: chore: Update sonar-java

### LINKED ISSUE #331: Update to sonar-java-7.0.0.26422
This is a bit tricky as they've changed some of the internals. All incompatibilities are confined within `RuleVerifier`, however, so the problems are closely contained. It should be manageable.

Currently using this issue to keep track of compatibility problems. Here are the currently known problems with different versions of sonar-java:

* 6.10+: Breaks issue filtering mechanisms, see https://github.com/SpoonLabs/sorald/issues/331#issuecomment-802678077
* 6.13.0.25138
    - Breaks rule 2755 (XxeProcessingProcessor), it just doesn't detect anything anymore
    - Gets a lot more grumpy about files that don't compile than previously (some NOCOMPILE test cases fail)

slarse:
#331 

Currently, updating to any recent version of Sonar breaks our handling of `NOSONAR` and `@SuppressWarnings`. Not bothering with it at the moment, see https://github.com/SpoonLabs/sorald/issues/331#issuecomment-802678077

I will periodically update this PR to ensure that we stay somewhat up-to-date with sonar-java, but won't try to fix the issue filtering at this time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,6 +10,14 @@
     <description>An automatic repair system for static code analysis warnings from Sonar Java.</description>
     <url>https://github.com/SpoonLabs/sorald</url>
 
+    <properties>
+        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.source>11</maven.compiler.source>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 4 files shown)


slarse:
accidentally closed

monperrus:
see also #331

PR #432, opened 2021-03-19
wip: break: Remove out-of-place repair modes

### LINKED ISSUE #409: Removing little used or redundant features (multi-rule repair, out-of-place printing and Git patch generation)

slarse:
Final part of #409 , so fixes #409!


slarse:
Replaced by #434 and #436

PR #433, opened 2021-03-19
refactor: Move PATH_TO_RESOURCES_FOLDER to TestHelper

slarse:
The PATH_TO_RESOURCES_FOLDER constant should not be available in production, as it's only used for tests. This PR moves it into an appropriate position.


khaes-kth (commented):
LGTM!

PR #434, opened 2021-03-19
break: Remove --workspace argument

slarse:
#409 

This is in preparation of removing out-of-place repair. With that gone, there is no need for a separate workspace directory, and consequently there is no need for a CLI argument to specify it.

Note that this PR only removes the workspace argument, the workspace is still used for out-of-place repair but is hard-coded to `sorald-workspace`. I will follow up on this PR with a couple more such that, in the end, out-of-place repair is gone. See #432 for the complete (but pretty messy) vision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,14 +59,16 @@ The full list of options is as follows (and can also be found by running
 `sorald repair --help`):
 
 ```bash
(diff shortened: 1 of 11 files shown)


fermadeiral:
Thanks, @slarse!

PR #436, opened 2021-03-22
break: Remove out-of-place repair

### LINKED ISSUE #409: Removing little used or redundant features (multi-rule repair, out-of-place printing and Git patch generation)

slarse:
Fix #409 

This PR removes out-of-place repair, so that Sorald now only operates in-place. It is the final part of removing the mostly redundant features that hinder or slow down further improvement of Sorald.

Changes to production code are small. The only significant changes to production code are the removal of the `--file-output-strategy` argument and some related modifications to `SoraldConfig`, `Repair` and `RepairCommand`.

The rest of the changes are in the tests, and that is entirely to accommodate for the fact that the `sorald-workspace` no longer is created automatically (all repair is in-place!). Any test that repairs code must be executed on a copy of the test resources, so as not to directly modify the test resources themselves. Utility functions for this is provided in `TestHelper`.

Ps. Sorry about the size of this PR. Splitting it further is possible but would result in a large amount of extra work for me, as all changes are pretty tightly related.


fermadeiral:
I love removal of code. Thanks @slarse!

slarse:
> I love removal of code. Thanks @slarse!

It's the best!

For the record, if we ever find a need to re-implement out-of-place repair, we can do so very easily on top of the current architecture:

* Out-of-place with all files preserved: Just copy the input directory/file to the desired destination and run in-place on the copy.
    - This is however so easily done by the user of sorald that I find it kind of pointless (`$ cp -r target target-copy; sorald repair --original-files-path target-copy`)
* Out-of-place with only changed files preserved: Copy the input directory/file to the desired destination, run in-place on the copy, delete all files that were not written to.

Just putting that here in case someone decides this functionality is desirable.

Issue #358, opened 2021-02-03
Add message to RuleViolation


slarse:
The message emitted by Sonar may be useful for repairs, and should be added to the API of `RuleViolation`. It should also be printed in the "warning locations" output.

slarse:
This essentially faces the same problem as #395. To get the messages when using specified rule violations, we need to mine for the violations.

PR #442, opened 2021-03-22
feat: Validate violation specs

### LINKED ISSUE #395: Validate user-specified violation specs (targeted repair)
Right now, there's no check that violation specifiers are actually real violations. This is a great potential source of user error: we want to be absolutely certain that rule violations sent to process are actually flagged as rule violations.

The fix for this is simple: always scan for violations, even when they are specified. Then match the specified violations to the scanned violations.

slarse:
Fix #395 

Adds validation to user-specified rule violations by checking that the precise violations specified actually exist. To be clear, it validates that violations specified like so exist:

```
$ sorald repair --violation-specs 2111:2111_BigDecimalDoubleConstructor/BigDecimalDoubleConstructor.java:21:25:21:45 [...]
```

 Previously, this was limited to the specified file existing. This new behavior is made possible by _always_ mining violations, which is something we'll need to do to implement #358 as well.

A downside is that this adds a performance overhead to specifying rule violations, as it requires a scan with Sonar that would otherwise not be necessary as we'd just assume that they are real. But, I think I'm pretty much the only one who currently uses targeted repair (I use it when developing), and I always run with stats output, which also requires a scan with Sonar, and so it doesn't matter to my use case.

The error message looks like so:

```
No actual violation matching violation spec: '2111:2111_BigDecimalDoubleConstructor/BigDecimalDoubleConstructor.java:21:25:21:45'
```

### DIFF
--- a/src/main/java/sorald/cli/RepairCommand.java
+++ b/src/main/java/sorald/cli/RepairCommand.java
@@ -137,7 +137,11 @@ class RepairCommand extends BaseCommand {
             specifiedRuleViolations.forEach(
                     specifiedViolation -> {
                         if (!minedViolations.contains(specifiedViolation)) {
-                            throw new IllegalStateException();
+                            throw new CommandLine.ParameterException(
+                                    spec.commandLine(),
+                                    "No actual violation matching violation spec: "
+                                            + specifiedViolation.relativeSpecifier(
+                                                    originalFilesPath.toPath()));
                         }
                     });
 
--- a/src/main/java/sorald/cli/RepairCommand.java
+++ b/src/main/java/sorald/cli/RepairCommand.java
@@ -3,6 +3,7 @@ package sorald.cli;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


khaes-kth (commented):
This new feature totally makes sense. But why shouldn't we have the same check when a user passes a file as the target?

slarse:
> This new feature totally makes sense. But why shouldn't we have the same check when a user passes a file as the target?

I don't understand, could you elaborate?

khaes-kth:
Sure. When we do not pass specified violations, we are asking Sorald to look for violations in "--original-files-path". In that case, if there is no violations of the given rule, we do not throw an exception. Why don't we throw an exception in that case, while we do it here?

slarse:
Oh, now I see what you mean.

There's a large semantic difference. If you specify `--original-files-path some/nice/path --rule-key some-rule-key`, and `some/nice/path` and `some-rule-key` both exist, then Sorald will scan for violations and repair any that it finds. As long as both exist, there is no great reason the believe there's user error, because there's no requirement to first run `sorald mine` before running `sorald repair`. I.e. a user might very well just run `sorald repair` both to see if there are violations, and to potentially repair them. I do that from time to time.

As for explicitly specifying a rule violation that does not exist, that is essentially guaranteed to be user error. It's about as much of an error as specifying a rule key that does not exist, because Sorald can't possibly do anything with it. It's also leaps and bounds easier to mistype or fail in copy/pasting a violation specification than a filepath.

Finally, these are two separate issues. Even if we would like to treat no violations found when specifying rule key as an error, that'd be a different PR :)

khaes-kth:
I understand your argument and I feel one can also argue the other way. Overall, I agree that these things can be considered separately. So, I'll merge this PR.

slarse:
> I understand your argument and I feel one can also argue the other way

Feel free to do so! I don't see a good reason to treat repair with rule key that doesn't find violations as an error, but if you do then share it.

PR #443, opened 2021-03-22
feat: Add message to rule violation

### LINKED ISSUE #358: Add message to RuleViolation

slarse:
Fix #358 

This adds a method `RuleViolation.getMessage()` that can be used to fetch the message emitted by Sonar. The reason we want this message is that it sometimes contains useful information. As a driver for the implementation and to illustrate that it's useful, I've refactored `CastArithmeticOperandProcessor` to use the message to parse the expected type.

See #375 for why it's proven inconvenient to try to get the type ourselves. Unfortunately, the test case provided there no longer causes a failure, as SonarJava does not flag a violation of rule 2184 inside method calls, it's only for assignments. The new, better position matching thus makes it impossible to trigger the bug, but it's still guaranteed to be gone as the method in which it was triggered `getExpectedtype()` has been removed.

### DIFF
--- a/src/main/java/sorald/processor/CastArithmeticOperandProcessor.java
+++ b/src/main/java/sorald/processor/CastArithmeticOperandProcessor.java
@@ -2,21 +2,32 @@ package sorald.processor;
 
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.commons.lang3.tuple.Pair;
 import sorald.Constants;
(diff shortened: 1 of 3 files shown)


slarse:
@fermadeiral Diff coverage isn't quite there due to not testing the throwing of two exceptions. I think testing them is more effort than it's worth at this point.

`getOpKindAndType` could be easily tested in isolation if there was an easy way to create a "fake" RuleViolation without having to implement the abstract class, so I'll open an issue for that.

fermadeiral:
Nice feat! Thanks @slarse!

Issue #445, opened 2021-03-22
Problems applying SerialVersionUidProcessor (rule 2057) to Spoon


slarse:
I tried to apply the SerialVersionUidProcessor to INRIA/Spoon@2823e58503e62d863ec75fe26ddfb9bf8d900cb4, which has 9 violations, but there were two problems.

1. One NPE (this is a problem in the processor)
2. For almost every patch, the added field appears on the same line as the class header (this is a problem in the sniper printer, missing newline)

I'll look into this more tomorrow. Most likely, 1) can be fixed in Sorald, while 2) must be fixed in Spoon.

slarse:
Confirmed that 2) is a problem in Spoon, working on a fix. See inria/spoon#3855

slarse:
Here's the stack trace for 1)

```
java.lang.NullPointerException
	at sorald.processor.SerialVersionUidProcessor.repairInternal(SerialVersionUidProcessor.java:63)
	at sorald.processor.SerialVersionUidProcessor.repairInternal(SerialVersionUidProcessor.java:14)
	at sorald.processor.SoraldAbstractProcessor.repair(SoraldAbstractProcessor.java:87)
	at sorald.processor.SoraldAbstractProcessor.process(SoraldAbstractProcessor.java:164)
	at spoon.support.visitor.ProcessingVisitor.scan(ProcessingVisitor.java:72)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:170)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:143)
	at spoon.reflect.visitor.CtScanner.visitCtClass(CtScanner.java:336)
	at spoon.support.reflect.declaration.CtClassImpl.accept(CtClassImpl.java:58)
	at spoon.reflect.visitor.CtScanner.scan(CtScanner.java:178)
	at spoon.support.visitor.ProcessingVisitor.scan(ProcessingVisitor.java:68)
	at spoon.support.QueueProcessingManager.process(QueueProcessingManager.java:118)
	at sorald.Repair.repairModelWithInitializedProcessor(Repair.java:198)
	at sorald.Repair.defaultRepair(Repair.java:128)
	at sorald.Repair.repair(Repair.java:105)
	at sorald.Repair.repair(Repair.java:90)
	at sorald.cli.RepairCommand.call(RepairCommand.java:113)
	at sorald.cli.RepairCommand.call(RepairCommand.java:33)
[...]

PR #447, opened 2021-03-23
feat(2057): Handle supertype implementing serializable

slarse:
#445 

This PR lets the `SerialVersionUidProcessor` handle the case where a class `B` extends a class `A` that is serializable, but `B` does not declare a `serialVersionUID` field. It is primarily a feature addition as this case was documented as ignored, but it also fixes a crash caused by the supertype `A` not being in the input sources. This is why the added test case subtypes `ArrayList`, it guards both against the crash and verifies the correct behavior.

This PR doesn't fully address issue #445, there's a pending PR in Spoon that fixes the other aspect.


khaes-kth (commented):
LGTM! Hard to understand why we did not fix this case before. Seems we almost had this feature implemented.

slarse:
The PR in which this processor was added (#368) was a bit drawn-out due to underlying problems in Sorald, and so we just wanted to merge it and then deal with additional problems like this one later :)

PR #449, opened 2021-03-24
doc: add paper information

monperrus:


slarse:
Nice, thanks @monperrus!

PR #451, opened 2021-03-24
break: Rename --original-files-path arg to --source

### LINKED ISSUE #429: Rename `--original-files-path` to `--source`
I propose to rename the `--original-files-path` argument to `--source`. Reason: I mistype it _all_ the time. In every demo I've ever made, I mess up on that argument. I can never quite remember if its `--original-files-path` or `--original-file-path`, and even when I do I still mistype it.

slarse:
Fix #429 

This PR renames the argument, and also updates all corresponding variable names in the source code, as well as updating docs and scripts.

I swear that I did not mistype the argument in `Rename --origina-files-path arg to --source` on purpose; it's just really error prone! :)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,8 +54,8 @@ jobs:
         env:
           SORALD_JAR_PATH: "target/sorald-1.1-SNAPSHOT-jar-with-dependencies.jar"
         run: |
-          java -jar "$SORALD_JAR_PATH" repair --original-files-path src/test/resources/ArrayHashCodeAndToString.java --rule-key 2184
-          java -jar "$SORALD_JAR_PATH" repair --original-files-path src/test/resources/ArrayHashCodeAndToString.java --rule-key 2116
+          java -jar "$SORALD_JAR_PATH" repair --source src/test/resources/ArrayHashCodeAndToString.java --rule-key 2184
+          java -jar "$SORALD_JAR_PATH" repair --source src/test/resources/ArrayHashCodeAndToString.java --rule-key 2116
       - name: Upload coverage report to Codecov
         uses: codecov/codecov-action@239febf655bba88b16ff5dea1d3135ea8663a1f9 # v1.0.15
         with:
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ and are specified by their key. For example, to repair violations of the rule
 `some/project/path`, one can invoke Sorald like so.
 
 ```bash
-$ sorald repair --original-files-path some/project/path --rule-key 2111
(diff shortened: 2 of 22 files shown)


slarse:
@fermadeiral Now that I'm looking at this, perhaps we should even make the source path a positional argument? It's always required, after all. For example, this:

```
java -jar "$SORALD_JAR_PATH" repair --source src/test/resources/ArrayHashCodeAndToString.java --rule-key 2184
```

Would the become this:

```
java -jar "$SORALD_JAR_PATH" repair src/test/resources/ArrayHashCodeAndToString.java --rule-key 2184
```

It would help in highlighting the fact that the source path must _always_ be provided. What do you think?

khaes-kth:
> It would help in highlighting the fact that the source path must _always_ be provided. What do you think?

I can imagine cases where we do not provide source path. What if Sorald takes a Github repo url as input and fixes violations in that repo?

slarse:
> > It would help in highlighting the fact that the source path must _always_ be provided. What do you think?
> 
> I can imagine cases where we do not provide source path. What if Sorald takes a Github repo url as input and fixes violations in that repo?

We could try to infer the type of the source path (check if it exists as a local file/directory, check if it has the extension `.git` for Git url), and provide an additional option for users to force a specific kind of source type. E.g. `sorald repair https://github.com/spoonlabs/sorald.git --source-type giturl`.

khaes-kth:
> The point is, we always need a source input for repair.

Fair enough!

khaes-kth:
Do you think we should merge this PR then? Or should we migrate to a "positional argument"?

slarse:
> Do you think we should merge this PR then? Or should we migrate to a "positional argument"?

I don't know :). It's all very subjective, neither is inherently better. Let's see what @fermadeiral thinks.

fermadeiral:
I have no preference.

However, I do think that we could merge this PR as it is and open an issue with Simon's proposal (https://github.com/SpoonLabs/sorald/pull/451#issuecomment-806423317) for improvements in the future.

fermadeiral:
Actually, I think I have a slight preference for having `--source` in the command line. I think it makes things more explicit.

slarse:
I think we should in the future make `--source` optional. It would make sense to have it default to the current directory, would it not? Keeping it as an option is simply more flexible, so let's go with that. I'm happy to have this merged!

fermadeiral:
> Keeping it as an option instead of positional is simply more flexible

Ok then.

slarse:
With all of this done, I'm happy to do the 0.1 release. I'll do it tomorrow morning!

PR #454, opened 2021-03-26
chore: Move Spoon snapshot repo into a profile

slarse:
We shouldn't declare the snapshot repository willy-nilly in the pom-file, see discussion over at inria/spoon#3216. This PR puts the snapshot repo in a profile, such that we can still easily access it when we want like so:

```
$ mvn -P spoonSnapshot ...
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,16 +18,6 @@
         </license>
     </licenses>
 
-    <repositories>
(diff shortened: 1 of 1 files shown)


fermadeiral:
Thanks @slarse.

PR #455, opened 2021-03-26
fix: Handle symlinks in input (add macOS support)

### LINKED ISSUE #416: Fix test errors on macOS
Probably related to filepaths.

### LINKED ISSUE #428: Sorald acts strangely around symbolic links
There's strangeness when using symbolic links anywhere in the input paths. This is what's causing the build errors on macOS (see #414).

slarse:
Fix #416
Fix #428

This PR adds a macOS build and fixes all test errors on macOS. They're all related to symbolic links. When creating a temporary directory on macOS (which we do e.g. in `TestHelper.createTemporaryTestResourceWorkspace()`), it's created in the `/var` directory. However, that's actually a symlink to `/private/var`, but the path you get back starts with `/var`. This causes problems, as Spoon always fully resolves symlinks while Sorald otherwise does not.

In order to debug and test this on a non-macOS system, I also added a hidden part to the test suite: if you set the environment variable `SORALD_TEST_SYMLINK_WORKSPACE=true`, then `TestHelper` returns a symlink to the temporary workspaces it creates, rather than a real path to the workspace. This does not however need to be active in tests, as it's covered by the new macOS build.

### DIFF
--- a/src/main/java/sorald/FileUtils.java
+++ b/src/main/java/sorald/FileUtils.java
@@ -16,14 +16,21 @@ import org.json.JSONObject;
 public class FileUtils {
 
     /**
-     * Compare the two given paths as absolute, normalized paths.
+     * Compare the two given paths as real paths, resolving any symbolic links.
+     *
+     * <p>Note that this method returns false if either path does not exist, even if the paths are
+     * otherwise equal!
      *
      * @param lhs A path.
      * @param rhs A path.
-     * @return Whether or not the paths are equal as absolute, normalized paths.
+     * @return true iff both paths exist and point to the same real file
      */
-    public static boolean pathAbsNormEqual(Path lhs, Path rhs) {
-        return lhs.toAbsolutePath().normalize().equals(rhs.toAbsolutePath().normalize());
+    public static boolean realPathEquals(Path lhs, Path rhs) {
+        try {
+            return lhs.toRealPath().equals(rhs.toRealPath());
+        } catch (IOException e) {
+            return false;
+        }
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/sorald/FileUtils.java
```diff
@@ -16,14 +16,21 @@
...
-     * Compare the two given paths as absolute, normalized paths.
+     * Compare the two given paths as real paths, resolving any symbolic links.
+     *
+     * <p>Note that this method returns false if either path does not exist, even if the paths are
+     * otherwise equal!
      *
      * @param lhs A path.
      * @param rhs A path.
-     * @return Whether or not the paths are equal as absolute, normalized paths.
+     * @return true iff both paths exist and point to the same real file
```
fermadeiral, line 26: Double f in "if".

slarse: That's intentional, it's stronger than "if": https://en.wikipedia.org/wiki/If_and_only_if

Perhaps this isn't as commonly accepted as KTH's math courses lead you to believe? Change it to "if"?

fermadeiral: > it's stronger than "if": https://en.wikipedia.org/wiki/If_and_only_if

Oh, I didn't know about that. Let's keep it like it is then!

slarse: > Oh, I didn't know about that. Let's keep it like it is then!

I mean, if you didn't know about "iff", that's a bit of a red flag in terms of using it :D

Then again, to anyone who hasn't seen "iff" before it will just look like a typo and so it doesn't really matter here. But in general, if you or anyone else reviewing a PR doesn't understand a term, then it's likely it's not the best term to use!

PR #458, opened 2021-03-29
feat(2057): Make newly added serialVersionUID field private

### LINKED ISSUE #457: Rule 2057: When adding a new serialVersionUID field, it should be private
When modifying an existing serialVersionUID field, we keep the existing visibility. When adding a new field, it makes most sense to make the field private.

slarse:
Fix #457 

Adds the `private` visibility to a `serialVersionUID` field if the field is _created_ by the processor, rather than the processor modifying an existing field. It makes most sense to make new fields private, as they're guaranteed not to be accessed anywhere else in the project (as they didn't exist prior to the repair!).

### DIFF
--- a/docs/HANDLED_RULES.md
+++ b/docs/HANDLED_RULES.md
@@ -480,7 +480,7 @@ The repair consists of add a serialVersionUID to classes implementing Serializab
 Example:
 ```diff
 public class NoUID implements Serializable {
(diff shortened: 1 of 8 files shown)


fermadeiral:
LGTM.

khaes-kth (approved):
LGTM.

PR #461, opened 2021-03-30
test: Extend test suite to be able to test exact literal matches

### LINKED ISSUE #393: Add functionality for testing literal output in test suite
Right now, the test suite tests the expected output only based on the AST. It would be great to be able to test the actual text, for example asserting the files match exactly, or that a certain text pattern is present in the output file.

slarse:
Fix #393 

This PR adds the long-awaited functionality to test exact, literal output from Sorald. See the updated `CONTRIBUTING.md` file (cc4ccd4) for how to use it.


fermadeiral:
Thanks @slarse!

Issue #440, opened 2021-03-22
Sorald in CI: Dogfood SoraldBot, breaking the build


slarse:
We should setup SoraldBot for this repo, and have it always use the latest snapshot version (see #439)

monperrus:
See #448

slarse:
On hold for now

slarse:
So, there's a bit of a snag with SonarCloud. The secret required to run the SonarCloud analysis is not available to a pull request triggered from a fork. That's reasonable: we don't want untrusted users to have access to secrets with code with haven't vetted.

The problem then is how to actually activate the analysis. [This blog post suggests a few alternatives](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). One pretty reasonable one is to allow a pull request from a fork access to secrets if it's labeled with a specific label, say `safe`, or something like that. Only repository collaborators are allowed to use labels, and so this would mean that code must be inspected before analysis runs.

Another alternative would be to run SonarCloud analysis on the main branch only, and use it reactively rather than proactively. We could then adopt the pragmatic approach of only breaking the build for issues that Sorald can fix. We could do something like this:

* Run SonarCloud only on the main branch, use it to keep track of issues.
* On pull requests, we do a simplified version of option 2: break the build if there's _ANY_ violation that Sorald can fix.
    - As long as we make sure Sorald is clear of any violations it can repair, this will be equivalent to breaking the build for newly introduced violations.

Thoughts?

slarse:
I think we can close this now as we have sorald-buildbreaker running on this repo.

PR #464, opened 2021-03-30
chore: Add SonarCloud configuration

slarse:
#440 

Adds SonarCloud configuration to the pom and GitHub Actions workflow. I think this should be sufficient to trigger SonarCLoud analyses.


slarse:
For future reference, the problem here causing SonarCloud not to run is that the token is not available to a PR from a fork (for security reasons).

slarse:
Replaced with #465

PR #466, opened 2021-03-31
chore: Add developer information to pom.xml

slarse:
Sonatype requires that the pom file contains [developer information](https://central.sonatype.org/pages/requirements.html#developer-information). Here's for example the corresponding entry in Depclean: https://github.com/castor-software/depclean/blob/add262810f6d1b93b44d31431d69c7d14cad1cd6/pom.xml#L62-L70

I figured we'll just put the currently active developers, which would be at least myself, and then possibly @fermadeiral and @khaes-kth. @fermadeiral @khaes-kth would you like to be added, and if so, with what details (email, organization etc)?

What should we put as organization? Since we're publishing under Castor's group id, I suppose we put Castor there? Perhaps it doesn't really matter, we could just put "KTH" for everyone.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,6 +23,8 @@
             <id>slarse</id>
             <name>Simon Larsén</name>
             <email>slarse@kth.se</email>
+            <organization>KTH Royal Institute of Technology</organization>
+            <organizationUrl>https://www.kth.se</organizationUrl>
         </developer>
     </developers>
 
--- a/pom.xml
+++ b/pom.xml
@@ -18,6 +18,14 @@
         </license>
     </licenses>
 
+    <developers>
+        <developer>
+            <id>slarse</id>
+            <name>Simon Larsén</name>
+            <email>slarse@kth.se</email>
+        </developer>
+    </developers>
+
     <dependencies>
         <dependency>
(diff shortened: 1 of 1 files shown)


khaes-kth:
Of course, I would like to be added as a developer. With my KTH email (khaes at kth dot se) and having KTH as my organization.

slarse:
@khaes-kth Added, please have a look at your info so it's all good :)

khaes-kth:
@slarse Yes, it is all good. Thanks!

slarse:
@fermadeiral Do you want your info added? If so, could you supply the email address I should use? Keep in mind that this will be published on Maven Central.

fermadeiral:
@slarse sure:

```
<developer>
    <id>fermadeiral</id>
    <name>Fernanda Madeiral</name>
    <email>fer.madeiral@gmail.com</email>
    <organization>KTH Royal Institute of Technology</organization>
    <organizationUrl>https://www.kth.se</organizationUrl>
</developer>
```

slarse:
@fermadeiral Added! Please review and merge at your leisure :)

fermadeiral:
Thanks @slarse.

PR #468, opened 2021-04-09
fix(2184): Make CastArithmeticOperand use uppercase literal suffix

### LINKED ISSUE #462: CastArithmeticOperand uses lower-case suffixes again
There was a regression in #443, and lowercase suffixes are again being used, instead of uppercase like we want. This regression wasn't caught as we do not have tests for the literal output from processors.

I'm waiting for #461 to be merged before I fix this issue, as that allows us to actually test the literal output, such that a regression can be caught by the tests next time.

slarse:
Fix #462


fermadeiral:
LGTM. Thank you!

PR #469, opened 2021-04-09
fix: Fix violations of Sonar rule 1118

slarse:
Hi,

This PR fixes 11 violations of [Sonar Rule 1118: 'Utility classes should not have public constructors'](https://rules.sonarsource.com/java/RSPEC-1118).

The patch was generated automatically with the tool [Sorald](https://github.com/SpoonLabs/sorald). For details on the fix applied here, please see [Sorald's documentation on rule 1118](https://github.com/SpoonLabs/sorald/blob/master/docs/HANDLED_RULES.md#utility-classes-should-not-have-public-constructors-sonar-rule-1118).

### DIFF
--- a/src/main/java/sorald/Constants.java
+++ b/src/main/java/sorald/Constants.java
@@ -3,8 +3,8 @@ package sorald;
 import java.io.File;
 
 public class Constants {
-    private Constants() {
-    }
+    private Constants() {}
+
     public static final String REPAIR_COMMAND_NAME = "repair";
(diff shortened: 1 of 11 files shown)


slarse:
Note that there are two "manual" edits on this PR: one for formatting with Spotless/Google Auto Format (4b78c5f), and one where the code generation of the Processors class added an extra newline (cd786f3). Both of these edits are automatic in practice, but they count as manual as they weren't performed by Sorald itself.

khaes-kth (approved):
LGTM!

PR #470, opened 2021-04-09
refactor: Fix violations of Sonar rule 3032

slarse:
Hi,

This PR fixes 1 violations of [Sonar Rule 3032: 'JEE applications should not "getClassLoader"'](https://rules.sonarsource.com/java/RSPEC-3032).

The patch was generated automatically with the tool [Sorald](https://github.com/SpoonLabs/sorald). For details on the fix applied here, please see [Sorald's documentation on rule 3032](https://github.com/SpoonLabs/sorald/blob/master/docs/HANDLED_RULES.md#jee-applications-should-not-getclassloader-sonar-rule-3032).

### DIFF
--- a/src/main/java/sorald/cli/SoraldVersionProvider.java
+++ b/src/main/java/sorald/cli/SoraldVersionProvider.java
@@ -28,7 +28,9 @@ public class SoraldVersionProvider implements CommandLine.IVersionProvider {
         Properties props = new Properties();
         try {
             props.load(
-                    Thread.currentThread().getContextClassLoader().getResourceAsStream(resourceName));
+                    Thread.currentThread()
+                            .getContextClassLoader()
+                            .getResourceAsStream(resourceName));
         } catch (Exception e) {
             return LOCAL_VERSION;
         }
--- a/src/main/java/sorald/cli/SoraldVersionProvider.java
+++ b/src/main/java/sorald/cli/SoraldVersionProvider.java
@@ -28,7 +28,7 @@ public class SoraldVersionProvider implements CommandLine.IVersionProvider {
         Properties props = new Properties();
         try {
             props.load(
-                    SoraldVersionProvider.class.getClassLoader().getResourceAsStream(resourceName));
+                    Thread.currentThread().getContextClassLoader().getResourceAsStream(resourceName));
         } catch (Exception e) {
             return LOCAL_VERSION;
         }


slarse:
The likelihood of Sorald ever being executed with a JEE runtime is pretty much non-existent, so this fix is somewhat pointless. But we should either fix the violation, or remove the processor, in my opinion.

I wouldn't be opposed to removing the processor entirely, JEE-specific fixes seems too niche to me.

fermadeiral:
> But we should either fix the violation, or remove the processor, in my opinion.

Why? Is this not like a false positive or something similar?

slarse:
> Is this not like a false positive or something similar?

Not really. If we were to execute Sorald with a JEE runtime, then this would be applicable. It's just pretty unlikely that will ever happen, unless Sorald starts to get adopted by enterprises. For the vast majority of projects that use Sonar, this rule isn't active as it is not part of the default quality profile (Sonar Way).

Of course, there's no way for Sonar to know which runtime will be used to run the application, and so it will flag all uses of the `getClassLoader` method as violations.

fermadeiral:
About accepting/rejecting this patch, I have no strong feelings. However, I don't think we should remove this processor.

slarse:
Then we should merge it to be able to easily use Sorald in CI (i.e. break the build if there's a violation that Sorald can repair). This is the last violation in Sorald that it itself has a processor to fix.

fermadeiral:
> Then we should merge it to be able to easily use Sorald in CI

Can't we just mark the line with NOSONAR, if the patch is not sound for this project?

# About not removing the processor #

My thought is that static analyzers can always find rule violations in specific contexts that possibly no problem will ever happen. Having said that, any patch for fixing rule violations can be like this one. This is why I don't think we should remove this processor; otherwise, we should start to remove more processors, maybe all.

Do you follow my point? I feel my explanation might be a bit confusing.

slarse:
> > Then we should merge it to be able to easily use Sorald in CI
> 
> Can't we just mark the line with NOSONAR, if the patch is not sound for this project?

We can, but using `NOSONAR` is generally considered bad practice as it disables all current and possibly future violations on that line. There's even a [Sonar rule that flags violations on finding NOSONAR comments](https://rules.sonarsource.com/java/RSPEC-1291).

In any case, I've realized there's a much easier "fix": just don't explicitly call `getClassLoader()`. It will be called in the background, as it should be. See #475. 

> My thought is that static analyzers can always find rule violations in specific contexts that possibly no problem will ever happen.

That is of course true, but the extent of this problem varies by rule. This particular rule is much worse in this regard as Sonar has no way of knowing what runtime will be used to run the program in the end. I'd imagine the use of JEE in the projects we can consider (open source projects) is tiny, and so 

So, I basically have three arguments for why we should not have this processor at all:

1. It is not part of the default SonarWay profile, and so will not be considered by the vast majority of projects that use Sonar.
2. It specifically targets JEE applications, which is very niche, and we don't know how to identify
    - I should note here that a prior version of this processor looked for `javax` imports. That's not a viable way to identify a JEE application. There are many `javax` packages in JDK 11+.
3. [This is a very complex topic](https://www.infoworld.com/article/2077344/find-a-way-out-of-the-classloader-maze.html). Which classloader to use is heavily context dependent, and we have no measures to determine the context.
    - Can anyone on this project argue for when to use which classloader? We should not have fixes for problems we don't understand.

That being said, after reading up more on how to use classloaders, we should not merge this PR. It's badness.

fermadeiral:
> There's even a Sonar rule that flags violations on finding NOSONAR comments.

I didn't know that, thank you for the info!

> In any case, I've realized there's a much easier "fix": just don't explicitly call getClassLoader(). It will be called in the background, effectively side-stepping the false positive. See #475.

I will see the issue.

> This particular rule is much worse in this regard as Sonar has no way of knowing what runtime will be used to run the program in the end.

I understand your point, and honestly haven't thought about it before. However, I still continue thinking that several rules could result in a lot of false positives, which is why we could potentially study the rules more prone to false positives and not include support for them in Sorald, or to add the support but with some sort of special treatment (I'm basically talking about a new approach to handle this type of situation, like the patch in this current PR). Currently, I have no idea, but this is something to think for the future.

About your three arguments, I agree with you.

> after reading up more on how to use classloaders, we should not merge this PR. It's badness.

Ok.

> We can keep the processor as it's easy enough to circumvent within Sorald, but in the future I think that we should try to only add processors that are generally applicable

I agree, but there is a challenge here, which is to know which processors are generally applicable. This requires a study.

> and can be explained by someone on the team

I don't understand this last part.

slarse:
> I agree, but there is a challenge here, which is to know which processors are generally applicable. This requires a study.

For a definitive relative ranking of rules, then sure. But some rules are easy to label as _not_ generally applicable, such as this one that only applies to JEE. Another good rule of thumb is to only pick rules that are part of the default SonarWay profile for Java, as these are the rules that SonarSource have found to be the most generally applicable.

> > and can be explained by someone on the team
>
> I don't understand this last part.

What I mean is that at least one person working on Sorald should be able to justify the fix. That requires understanding of the underlying problem, which I don't think we have for this classloader issue.

PR #471, opened 2021-04-12
chore: Use setup-java v2.0.0

### LINKED ISSUE #467: Update setup-java in GitHub Actions workflow to v2.0.0
New release: https://github.com/actions/setup-java/releases/tag/v2.0.0

There's a breaking change in that we must now also specify the distribution to use (Zulu or Adopt).

slarse:
Fix #467


fermadeiral:
Thanks @slarse!

Issue #150, opened 2020-10-16
Make a release of sorald?


slarse:
The project has both had a name change and a major uplift to the UI since the last release. With a new release, we could also update the README with usage instructions for the pre-built jar. It's a bit clunky to have to build the project from source.

@fermadeiral

fermadeiral:
Yes, we can do that!

What about to schedule the release for next week, so we have time to finish some of the ongoing changes?

slarse:
Sure, that'll be great!

slarse:
@fermadeiral Could you set up a new release? The current PRs are only test related and so aren't relevant. I think it's a good idea to get a release out the door before we upgrade to sonar-java 6.9, as that may cause further issues.

fermadeiral:
> @fermadeiral Could you set up a new release?

Yes, I will do that tomorrow!

slarse:
@monperrus @fermadeiral 

We discussed publishing a release during the meeting, and touched upon publishing to Maven Central. I checked and there's no release of Sorald (or sonarqube-repair) up there, there's just [this one snapshot of sonarqube-repair](https://oss.sonatype.org/#nexus-search;quick~sonarqube-repair). At least that's what I could find.

Now, I don't think it's a bad thing we don't have any releases on Central, because that's a repository primarily for _libraries_, and Maven is a tool for dependency management rather than application distribution. Sorald is designed and documented as an application, and it has neither a well-defined public API nor any user documentation to go with it, and so using it as a library is ill-advised at the moment.

I think that for the time being, we shouldn't publish on Central, and simply publishing releases of the application here on GitHub (https://github.com/SpoonLabs/sorald/releases) is fully sufficient. Of course, in the future, we could create a nice API for Sorald and make it usable both as a library, and as an application. Then, I think it'd be appropriate to publish to Central.

What do you think?

monperrus:
Per our discussion today, we release on Central, esp for Repairnator which uses Sorald as library.

slarse:
@monperrus I've been looking around a bit now, and I've got no idea of where to create an account. Any pointers? :)

monperrus:
AFAIR https://jira.sonarsource.com/secure/Signup!default.jspa

monperrus:
[...]

PR #472, opened 2021-04-12
chore: Set version to 0.1.0-SNAPSHOT

slarse:
#150 

This PR sets the version number to 0.1.0-SNAPSHOT and updates support scripts that had the old version number hard coded.


slarse:
@fermadeiral This is the final PR before I can start experimenting with snapshot and release deployments :)

fermadeiral:
@slarse so let's merge it!

slarse:
Awesome, now let's see if I can get that automatic deployment working.

PR #475, opened 2021-04-13
refactor: Avoid explicitly accessing getClassLoader()

slarse:
This avoids a false positive for rule 3032 discussed in #470.

### DIFF
--- a/src/main/java/sorald/cli/SoraldVersionProvider.java
+++ b/src/main/java/sorald/cli/SoraldVersionProvider.java
@@ -27,7 +27,7 @@ public class SoraldVersionProvider implements CommandLine.IVersionProvider {
     public static String getVersionFromPropertiesResource(String resourceName) {
         Properties props = new Properties();
         try {
-            props.load(SoraldVersionProvider.class.getResourceAsStream(resourceName));
+            props.load(SoraldVersionProvider.class.getResourceAsStream("/" + resourceName));
         } catch (Exception e) {
             return LOCAL_VERSION;
         }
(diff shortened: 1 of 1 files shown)


slarse:
We can also use `@SuppressWarnings("squid:S3032")` to suppress this rules in the given method. The problem with the suppression is that we won't get any indication of if it becomes redundant, i.e. if the source of the rule violation is removed, the suppression will silently remain and do nothing. I think it's better to just do this little workaround, since it's so easy.

@fermadeiral what do you think, suppress the warning or use this workaround?

fermadeiral:
> This avoids a false positive for rule 3032

But in https://github.com/SpoonLabs/sorald/pull/470#issuecomment-817624519 you said that this is not a false positive. Maybe we should align our understanding on false positives in the context of static analyzers. To be discussed in the next meeting I will attend?

> I think it's better to just do this little workaround, since it's so easy.

LGTM. Merging...

slarse:
> > This avoids a false positive for rule 3032
> 
> But in [#470 (comment)](https://github.com/SpoonLabs/sorald/pull/470#issuecomment-817624519) you said that this is not a false positive. Maybe we should align our understanding on false positives in the context of static analyzers. To be discussed in the next meeting I will attend?

Ah yes that is inconsistent of me. It's a matter of perspective. It's a false positive if Sorald never executes in a JEE runtime, and a true positive of it ever is. The problem is that it's impossible to determine (because it implies foreseeing the future), which is one of the reasons I think it's such a poor rule to implement a processor for.

So, it isn't theoretically a false positive, but in practice it probably is.

Issue #478, opened 2021-04-13
create our Github Action for breaking the build based on sorald


monperrus:
Then, we can use this Github Action in Spoon and other repos.

(per our conversation at today's meeting)

monperrus:
(similar task in Repairnator: https://github.com/eclipse/repairnator/issues/836)

slarse:
Things that we need to be able to configure in the action (for a first version):

* Source input path
* Version of Sorald to use (we fetch the jar directly from Maven Central)
* Whether or not to only consider changed code

slarse:
I've been looking at creating a GitHub Action out of the scripts in #481. It's possible to create a GitHub Action only using shell scripts, but the support libraries are all in NodeJS, and so using NodeJS seems like the best path.

As I'm not all too familiar with NodeJS, putting it all together is a little bit slow, but I'll probably have a basic prototype ready sometime tomorrow.

slarse:
Quick update on the status here.

I've managed to more or less translate the scripts in #481 into an easy-to-use GitHub Action over in https://github.com/slarse/test-github-action. See #486 for an example use of the action. It's really very concise, and creating an action was much easier than I anticipated.

Next step is to move the prototype from slarse/test-github-action into a SpoonLabs repository, and then we'll take it from there.

@monperrus @fermadeiral @khaes-kth We need a name for the action, any suggestions? My working title is `sorald-buildbreaker`, but that sounds so negative. We don't need to settle on a name until we decide to publish the action on the marketplace (if we do), but I thought I'd seed the idea of coming up with one.

slarse:
Name suggestions:

* `sorald-buildbreaker`
* `sorald-watch`
* `sorald-ci`

slarse:
See [SpoonLabs/sorald-buildbreaker](https://github.com/SpoonLabs/sorald-buildbreaker)

monperrus:
I like the name "sorald-buildbreaker" very much!

PR #481, opened 2021-04-14
experimentation: Add buildbreaker script and workflow

slarse:
This is a first step toward #478, and also related to #440

The `sorald.buildbreaker` script is a crude prototype of breaking the build when Sorald finds a repairable violation. It works as follows.

1. Mine violations for all handled rules
2. Perform targeted repair of the mined violations
    - This will in the future allow us to only select violations that are within edited parts of the code!
3. If Sorald succesfully repairs any of the found violations, exit non-zero. Otherwise, exit zero.

This PR also adds a GHA workflow that first fetches the latest version of Sorald from Maven Central, and then runs the buildbreaker script. Currently, a lot of stuff is hard-coded (such as the version of Sorald to use, and the source path to analyze). That will later be input for the GitHub Action.

* Example of running the workflow when there are repairable violations (i.e. **failure**): https://github.com/SpoonLabs/sorald/runs/2342788710#step:7:8
    - Note: I've since changed the "error" message to `Repairable violations found: <violations>`
* Example of running the workflow when there are no repairable violations (i.e. **success**): https://github.com/SpoonLabs/sorald/runs/2342832172#step:7:8

This is all very much a WIP, and things are bound to be tweaked, but I think it's a good first attempt that's ready to be merged into the repo. We'll iterate on this until satisfied, and then create a GitHub Action of it.

### DIFF
--- a/.github/workflows/sorald_buildbreaker.yml
+++ b/.github/workflows/sorald_buildbreaker.yml
@@ -1,0 +1,37 @@
+# Workflow for breaking the build if Sorald finds a repairable violation.
+#
+# Note that actions are specified by commit hash. This is to avoid the security
+# risk of someone injecting malicious code into a release and then simply
+# changing a tag.
+
+name: sorald-buildbreaker
+
+on:
+  pull_request:
+  push:
+    branches: master
+
+jobs:
+  check-for-repairable-violations:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
+      - uses: actions/setup-java@8764a52df183aa0ccea74521dfd9d506ffc7a19a # v2.0.0
+        with:
+          java-version: 11
+          distribution: 'adopt'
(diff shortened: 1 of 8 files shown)


slarse:
As I've already progressed so far with #486, perhaps we should not merge these scripts. They were a good starting point for my prototyping, but I don't think we'll need them now.

khaes-kth (approved):
Looks very good to me!

khaes-kth:
@slarse Do you think we should wait until #486 is ready and merge that one instead?

slarse:
@khaes-kth Oh sorry, did you miss my comment before reviewing? Yeah, I'm not sure of the value of these scripts given that the end-goal is to encompass them into the GitHub Action anyway, and I've progressed so far with the prototype.

khaes-kth:
Great. Then, we'll wait for it.

BTS, Why shouldn't we close this PR without merge?

slarse:
> BTS, Why shouldn't we close this PR without merge?

I just keep it open for visibility at the moment.

slarse:
So, this is now over and done with as #486 has come to the point where it fulfills the same role. We don't need these scripts anymore.

PR #484, opened 2021-04-14
chore: Only run deploy job in SpoonLabs/sorald

### LINKED ISSUE #477: Deploy workflow should not run in forks
It doesn't make sense for it to run in forks, and it will always fail there due to a lack of the credentials. It ran on my fork and caused a workflow failure, which is annoying.

slarse:
Fix #477 

This PR edits the deploy job to only run if the repository is called `SpoonLabs/sorald` (i.e. does not run on forks, where it will always fail!)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -15,6 +15,7 @@ env:
 
 jobs:
   deploy:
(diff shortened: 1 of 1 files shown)


fermadeiral:
LGTM.

PR #494, opened 2021-04-20
fix: Make support scripts/CI workflows find any version of the Sorald JAR

### LINKED ISSUE #483: Support scripts/CI workflows expect SNAPSHOT version for JAR file
The CI failed horribly on the release as all support scripts expect a SNAPSHOT version of the JAR file. See https://github.com/SpoonLabs/sorald/actions/runs/748493223

The deployment still went through, though, as it only requires the unit tests to pass. Obviously not ideal, I'll have to check if there's any way to wait for all actions to finish before the deployment runs.

slarse:
Fix #483 

This PR simply removes the expectation that the compiled JAR file is a SNAPSHOT version, which is important for release commits.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,7 +60,7 @@ jobs:
       - name: Sanity check jarfile
         shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


fermadeiral:
LGTM.

PR #495, opened 2021-04-20
doc: Update README instructions on building and using Sorald

slarse:
I noted that the instructions in the README on building and using Sorald are a bit out of date, and so this PR updates them to reflect the current state of the project.

Notably, it expands upon the requirements for building Sorald from source, notes that there are prebuilt JARs available under releases, and simplifies some related instructions.


fermadeiral:
Thank you, @slarse!

PR #498, opened 2021-04-22
fix: Fix memory leak from accumulating compilation units across writes

### LINKED ISSUE #437: Compilation unit collector must be cleared after each model write
If it isn't cleared, the already written models will remain in memory. This results in segment mode not working as it should in terms of memory.

It seems prudent at this point to introduce a test for the memory behavior. It could possibly be done by cloning some very large project and running Sorald on it from the command line, with a sufficiently small heap for the JVM to pull through in segment mode, but not in normal mode.

slarse:
Fix #437 

This fixes a memory leak in segment mode caused by the CompilationUnitCollector retaining references to already printed compilation units. Also replaces the JAR sanity check test with a highly memory-constrained test, using the packaged JAR in segment mode. [Here's an example of where it reproduces the memory leak](https://github.com/SpoonLabs/sorald/pull/498/checks?check_run_id=2410450540#step:11:30).

The memory constrained test takes as long to run as all of the unit tests combined, but I think that's fine considering that it adds a real-world test to the test suite. Note that it also asserts the amount of repairs (170) for the given rule, so it's the largest-scale test we have right now.

### DIFF
--- a/.github/mem-constrained-exec.sh
+++ b/.github/mem-constrained-exec.sh
@@ -16,7 +16,7 @@ if [[ ! (-f "$SORALD_JAR_PATH") ]]; then
 fi
 
 HEAP_SIZE="128m"
-MAX_FILES_PER_SEGMENT=50
+MAX_FILES_PER_SEGMENT=25
 
 echo "Running memory-constrained test with heap size=$HEAP_SIZE and segment size=$MAX_FILES_PER_SEGMENT"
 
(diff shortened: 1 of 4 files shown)


fermadeiral:
> Note that it also asserts the amount of repairs (170) for the given rule, so it's the largest-scale test we have right now.

Nice!

Thanks for the fix, @slarse!

PR #499, opened 2021-04-23
chore: Remove SNAPSHOT from path in memory-constrained test script

slarse:
Another one of those hard-coded SNAPSHOT paths had snuck its way into the CI. It works fine as long as the current version in the repository is a SNAPSHOT version, but blows up on releases.

This PR simply removes `SNAPSHOT-` from the file path.

### DIFF
--- a/.github/mem-constrained-exec.sh
+++ b/.github/mem-constrained-exec.sh
@@ -9,7 +9,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-SORALD_JAR_PATH=$(echo target/sorald-*-SNAPSHOT-jar-with-dependencies.jar)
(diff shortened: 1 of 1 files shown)


fermadeiral:
LGTM!

fermadeiral:
@slarse, do you prefer that I always delete your branch used in the PR just after merging it?

slarse:
@fermadeiral Yeah I think that's good. Actually, I'll just activate auto-deletion of merged branches in the repository :)

Not sure if that applies to forks, but right now I'm working within the Sorald repo for the SonarCloud analysis to work (PRs from forks can't get the secret token).

fermadeiral:
> Actually, I'll just activate auto-deletion of merged branches in the repository :)

Great, thanks!

Issue #482, opened 2021-04-14
Document release process


slarse:
Here's the entire release process that I performed, which appears to have worked:

### Command line steps

* `mvn -B clean release:prepare -DpushChanges=false`
    - This creates two _unsigned_ commits: one for the release, and then one for the next development iteration
    - Does not push to the repo (we want to make some edits first)
* `git rebase -x 'git commit -S --amend --no-edit' HEAD~2`
    - This rewrites the two commits in the previous step and signs them with whatever GPG key you have configured
    - We want all commits on the master branch to be signed!
    - You can use any GPG key that's uploaded to GitHub, this is unrelated to publishing to Maven Central
    - See the [GitHub docs on signing commits](https://docs.github.com/en/github/authenticating-to-github/signing-commits)
* `git tag -s -f sorald-<VERSION> HEAD~`
    - Replace `<VERSION>` with the release version
    - This signs the tag with your GPG key
* `git push git@github.com:spoonlabs/sorald.git HEAD~:master --tags`
    - This pushes the tag and the release commit
    - Wait a few seconds for the deploy GitHub Action to kick in before pushing the development iteration commit (GitHub may otherwise skip running the deploy action for the release)
* `git push git@github.com:spoonlabs/sorald.git`
    - This pushes the next development iteration commit

### GitHub steps

* Go to [releases](https://github.com/SpoonLabs/sorald/releases) and click `Edit` on the tag for the new version.
* Use the tag name as the release name
* Add the changelog to the body, and any other relevant information
* Add the `jar-with-dependencies` jar to the artifacts
    - You should have it locally in `target/sorald-<VERSION>-jar-with-dependencies.jar`
    - You can also get it from Maven Central at `https://oss.sonatype.org/service/local/repositories/releases/content/se/kth/castor/sorald/<VERSION>/sorald-<VERSION>-jar-with-dependencies.jar`
    - Take care to replace the `<VERSION>` placeholders with the release version
* Publish the release!

### Future improvements

* Eliminate the need to push the release and next iteration commits at separate times
    - Can for example run the deploy action on creation of a GitHub release
* Create a script that does everything

PR #500, opened 2021-04-23
chore: Only run non-SNAPSHOT publish step on publish of release

slarse:
Related to #482 

This adjusts the deploy workflow such that it only deploys non-snapshot versions once a GitHub release is published. This makes the release process a bit more predictable than when deploying from master, as GitHub Actions typically only builds the very latest commit, which when pushing a release is in fact the "next iteration" commit, and not the release commit.

I had an awkward workaround for this (push the release commit, then wait until the action starts, then push the next iteration commit). By deploying only on a published release, that workaround is no longer necessary, and it also allows us to release _before_ merging into master. That's good if there's trouble in the publish step.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -58,7 +58,7 @@ jobs:
           gpg-passphrase: SORALD_GPG_PASSPHRASE
 
       - name: Publish to Maven Central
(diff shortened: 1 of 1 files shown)


fermadeiral:
LGTM.

PR #501, opened 2021-04-23
chore: Disable PR comment

### LINKED ISSUE #497: Disable codecov report on PRs
The Codecov report is a huge comment on PRs and most often it's not necessary. If the checks pass we're fine, if they don't pass we can go find the report on codecov.io instead.

slarse:
Fix #497 

The Codecov PR comment is large and in the way, let's get rid of it. We're really only interested in the check, if we want the report we can go to codecov.io.

I think that the change only takes effect once merged into master.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,8 @@
 codecov:
   require_ci_to_pass: yes
 
+comment: false
+
 coverage:
   precision: 2
   round: down


fermadeiral:
> The Codecov PR comment is large and in the way

It also bothers me. Thanks for changing that!

Issue #426, opened 2021-03-17
Use workflow GitHub token in support script tests to raise rate limit


slarse:
Support script tests are failing due to GitHub rate limiting. We should use the workflow's token to avoid that problem.

_Originally posted by @slarse in https://github.com/SpoonLabs/sorald/issues/414#issuecomment-800986930_

PR #502, opened 2021-04-23
feat(1854): Avoid deleting non-static method invocations

### LINKED ISSUE #491: DeadStoreProcessor (rule 1854) may remove invocations that have side effects
For example, given a dead store like so:

```java
int value = methodCallWithSideEffect();
```

Then, Sorald will repair it like so:

```diff
-int value = methodCallWithSideEffect();
```

This changes the semantics of the program!

The ideal way to fix this problem would be to analyze any called methods for side effects. But, we won't always have all of the sources, and so that in general won't be possible. The _safe_ way to do this would be to never remove method calls, but to do the patches like so instead:

```diff
-int value = methodCallWithSideEffect();
+methodCallWithSideEffect();
```

But that has the drawback of leaving useless method calls:

```diff
-int value = methodCallWithSideEffect();
-int otherValue = methodCallWithoutSideEffect();
+methodCallWithSideEffect();
+methodCallWithoutSideEffect();
```

One could imagine we make this configurable by way of options (see #334). For example, I imagine an invocation like so:

```bash
$ sorald repair --source . --rule-key 1854 \
	--rule-options side-effect-free=methodCallWithoutSideEffect()
```

Leading to this diff instead:

```diff
-int value = methodCallWithSideEffect();
-int otherValue = methodCallWithoutSideEffect();
+methodCallWithSideEffect();
```

slarse:
Fix #491 

As non-static method invocations are likely to have side effects, we don't delete them. Static method invocations are however still deleted, as they are fairly unlikely to have side effects. Static invocations _could_ still have non-local side effects, such as sending a network request or the like. If we find that's a problem, we'll deal with it then.


slarse:
The support script tests have hit the rate limit again, need to see to #426 :)

slarse:
I think this is ready for review now. The patch diff hit isn't great because of defensive programming on fetching types with Spoon (which may not be available in noclasspath mode), and so some branches aren't taken. I don't feel like it's worth testing, see the [report at codecov](https://app.codecov.io/gh/SpoonLabs/sorald/compare/502/diff#diff-c3JjL21haW4vamF2YS9zb3JhbGQvcHJvY2Vzc29yL0RlYWRTdG9yZVByb2Nlc3Nvci5qYXZh)

fermadeiral:
I didn't check the actual patch, but I agree with the change in Sorald based on what you explained.

WDYT about writing this in [HANDLED_RULES.md](https://github.com/SpoonLabs/sorald/blob/master/docs/HANDLED_RULES.md)?

slarse:
> WDYT about writing this in [HANDLED_RULES.md](https://github.com/SpoonLabs/sorald/blob/master/docs/HANDLED_RULES.md?rgh-link-date=2021-04-26T11%3A41%3A57Z)?

Yup, that's a good idea. In documenting it, I came up with an example that wasn't handled correctly, so I also found a bug to fix :). I'll re-request a review once I'm done.

slarse:
@fermadeiral Ready again. Fixed a bug where invocations to unresolved methods would be assumed to be static, and thus were deleted (we want it the other way around, if they are unresolved we assume instance method and don't delete).

Also documented the new behavior in HANDLED_RULES docs

fermadeiral:
LGTM! Thank you.

Issue #479, opened 2021-04-14
Add rule key to summary output of miner


slarse:
I often use the miner to find what I can repair in a project, and it's always somewhat annoying to try to figure out which rule key corresponds to which rule. The summary output is just `CheckName=<num_violations>`, and to find the key I either have to go into the stats file or browse through the repair command's help section.

It'd be much nicer if the output was on the form `CheckName[<rule_key>]=<num_violations>`, or something like that, as you then wouldn't have to go hunting for the key.

slarse:
@khaes-kth Do you want to do this?

khaes-kth:
@slarse 
Yes, I do.

slarse:
Cool, I'll leave this to you then!

slarse:
Fixed in #505

PR #505, opened 2021-04-27
feat: Miner output also includes rule keys

khaes-kth:
This addresses #479.


slarse (approved):
Code looks good! Failing workflow is due to #506, so we can ignore it.

slarse:
Thanks @khaes-kth, this will make my work a lot less annoying!

PR #508, opened 2021-04-28
chore: Only run Sonar step on the main repository

### LINKED ISSUE #506: Skip Sonar on PRs from forks
Currently, we try to run SonarCloud on PRs from forks, which causes the builds to fail as the token isn't available. We should just skip this step on PRs from forks.

slarse:
Fix #506 

This prevents failing the workflow when a PR comes in from a fork.


fermadeiral:
LGTM. Thank you, @slarse.

PR #521, opened 2021-05-06
feat: Utilize classpath when repairing in Maven mode

### LINKED ISSUE #205: Add possibility to run Sorald in classpath mode (and specify the classpath)

slarse:
Fix #205 

Currently, this is just a test case that demonstrates the need to run in classpath mode.

### DIFF
--- a/src/test/java/sorald/MavenLauncherTest.java
+++ b/src/test/java/sorald/MavenLauncherTest.java
@@ -1,11 +1,17 @@
 package sorald;
 
+import static org.hamcrest.CoreMatchers.equalTo;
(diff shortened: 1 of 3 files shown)


fermadeiral:
Thank you for this contribution, @slarse!

Issue #485, opened 2021-04-15
Codecov security breach (we are kind of affected)


slarse:
There was a [security breach with Codecov's coverage uploader](https://codecov.io/disclosure). This repo was affected only by my SonarCloud token, which may have been stolen. I have revoked the compromised token and placed a new one.

Our other secrets were added after the security event, and are in any case sectioned off into an environment that Codecov's uploader does not have access to.

I'll leave this issue here for a few days and then I'll close it, just wanted to make everyone aware.

monperrus:
ack, thanks for the notice.

PR #525, opened 2021-05-11
chore: Use codecov-action v1.5.0

slarse:
This PR bumps codecov-action to v1.5.0. The big difference is that the bash uploader (which was the cause of the security breach mentioned in #485) is integrated into the action rather than being fetched from a remote source, making it much more secure.


fermadeiral:
Thanks @slarse.

PR #527, opened 2021-05-12
feat: Add classpath resolution for Maven projects to miner

### LINKED ISSUE #523: Implement Maven mode for miner
As shown in #521, it's possible to find more rule violations when the classpath is set to include all dependencies. We should extend the miner to also have a Maven mode that makes use of the classpath.

slarse:
Fix #523, related to #533 

This PR implements classpath resolution for Maven projects for the miner. I decided to go with adding a flag `--resolve-classpath-from <path-to-project-root>`, that currently only works for Maven projects (there must be a file `<path-to-project-route>/pom.xml`).

I chose to go this route instead of adding e.g. `--maven` or `--miner-mode MAVEN`, as we can quite easily add support for other build systems and just infer them by looking for build files. For example, to add support for Gradle, and then probably use the `gradle dependencies` command in some way. In other words, we can add support for more build systems without changing the CLI.

Initially I opted for just a `--resolve-classpath` argument that uses the `--source` path as the project route, but as I ran into issues with the repair version of Maven mode, I decided to go with separate source and classpath resolution paths. After this is merged, I will extend the feature to also work for repair mode, see #533.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,15 +203,20 @@ mine --help`).
       --git-repos-list=<reposList>
                              The path to the repos list.
   -h, --help                 Show this help message and exit.
+      --handled-rules        When this argument is used, Sorald only mines
+                               violations of the rules that can be fixed by
+                               Sorald.
       --miner-output-file=<minerOutputFile>
                              The path to the output file.
-      --source=<source>
-                             The path to the file or folder to be analyzed and
-                               possibly repaired.
+      --resolve-classpath    Resolve the classpath of a project for more
+                               accurate scans. Currently only works for Maven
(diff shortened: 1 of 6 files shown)


slarse:
@khaes-kth Hold off on reviewing this, I've run into problems with the `repair` version of Maven mode and I think I can slightly alter this implementation to get a unified solution.

slarse:
@khaes-kth OK it's all good now. Slightly revised to solve the problem I've got in repair mode.

khaes-kth (approved):
Looks good to me. Thanks @slarse

PR #528, opened 2021-05-12
feat(2142): Put curly brackets around inserted if-statement

### LINKED ISSUE #526: Rule 2142: InterruptedExceptionProcessor should put curly brackets around `if`
When a catcher has a union type, the processor produces a bracketless if-statement:

```diff
                        try {
                                job.get();
                        } catch (InterruptedException | ExecutionException e) {
+                               if (e instanceof InterruptedException)
+                                       Thread.currentThread().interrupt();
+
                                throw new SpoonException("failed to wait for parallel processor to finish", e);
                        }
                }
```

In my experience, it's much more common for code styles to require curly brackets than not to, so I think we should put brackets by default.

slarse:
Fix #526 

This PR amends the `InterruptedExceptionProcessor` to put curly brackets around inserted if-statements. Previously, it would look like so:

```diff
                        try {
                                job.get();
                        } catch (InterruptedException | ExecutionException e) {
+                               if (e instanceof InterruptedException)
+                                       Thread.currentThread().interrupt();
+
                                throw new SpoonException("failed to wait for parallel processor to finish", e);
                        }
                }
```

but with this patch it looks like so:

```diff
                        try {
                                job.get();
                        } catch (InterruptedException | ExecutionException e) {
+                               if (e instanceof InterruptedException) {
+                                       Thread.currentThread().interrupt();
+                               }
+
                                throw new SpoonException("failed to wait for parallel processor to finish", e);
                        }
                }

### DIFF
--- a/src/main/java/sorald/processor/InterruptedExceptionProcessor.java
+++ b/src/main/java/sorald/processor/InterruptedExceptionProcessor.java
@@ -82,7 +82,11 @@ public class InterruptedExceptionProcessor extends SoraldAbstractProcessor<CtCat
         ctIf.setCondition(
                 fact.createBinaryOperator(
                         varRead, interruptedExcAccess, BinaryOperatorKind.INSTANCEOF));
-        ctIf.setThenStatement(statementToWrap);
+
+        // We add a block instead of just the naked statement here to force printing of curly
+        // brackets
+        CtBlock<?> thenBranch = fact.createCtBlock(statementToWrap);
+        ctIf.setThenStatement(thenBranch);
 
         return ctIf;
     }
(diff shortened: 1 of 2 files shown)


fermadeiral:
LGTM. In fact I haven't noticed that before, otherwise I would have opened an issue about it. I am a curly brackets person :)

Thank you, @slarse!

slarse:
> I am a curly brackets person :)

So am I. You've probably not noticed this was a problem (just like I hadn't) as this particular part of this processor is new, and I don't think it's been used yet. It was when I first tried to use it in practice that I noticed it didn't put out curly brackets :)

fermadeiral:
@slarse this is why I will miss you in the Sorald project: you are a detailed person :)

PR #534, opened 2021-05-17
DONTMERGE: This is just a test PR for sorald-buildbreaker

slarse:
I'm just trying to implement PR commenting, see spoonlabs/sorald-buildbreaker#53

### DIFF
--- a/src/main/java/sorald/Main.java
+++ b/src/main/java/sorald/Main.java
@@ -8,6 +8,9 @@ public class Main {
             throw new InterruptedException();
         } catch (InterruptedException | IllegalArgumentException e) {
             System.out.println("oops!");
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
         }
 
         int exitStatus = Cli.createCli().execute(args);
--- a/src/main/java/sorald/Main.java
+++ b/src/main/java/sorald/Main.java
@@ -8,9 +8,6 @@ public class Main {
             throw new InterruptedException();
         } catch (InterruptedException | IllegalArgumentException e) {
             System.out.println("oops!");
-            if (e instanceof InterruptedException) {
-                Thread.currentThread().interrupt();
-            }
         }
 
         int exitStatus = Cli.createCli().execute(args);
--- a/src/main/java/sorald/Main.java
+++ b/src/main/java/sorald/Main.java
@@ -8,6 +8,9 @@ public class Main {
             throw new InterruptedException();
         } catch (InterruptedException | IllegalArgumentException e) {
             System.out.println("oops!");
+            if (e instanceof InterruptedException)
+                Thread.currentThread().interrupt();
+
         }
 
         int exitStatus = Cli.createCli().execute(args);
--- a/src/main/java/sorald/Main.java
+++ b/src/main/java/sorald/Main.java
@@ -12,6 +12,7 @@ public class Main {
 
         int exitStatus = Cli.createCli().execute(args);
         if (exitStatus != 0) {
+            int otherExitStatus = exitStatus;
             System.exit(exitStatus);
         }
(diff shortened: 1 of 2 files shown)


src/main/java/sorald/Main.java
```diff
@@ -4,6 +4,12 @@
 
 public class Main {
     public static void main(String[] args) {
+        try {
+            throw new InterruptedException();
+        } catch (InterruptedException | IllegalArgumentException e) {
+            System.out.println("oops!");
+        }
+
```
slarse, line 12:
```suggestion
this is better
```

slarse: HAH, it worked :D

slarse, line 12:
```suggestion
this is better
```

monperrus: oh super nice!

slarse, line 12:
```suggestion
this is better
```

src/main/java/sorald/Main.java
```diff
@@ -4,6 +4,12 @@
 
 public class Main {
     public static void main(String[] args) {
+        try {
+            throw new InterruptedException();
+        } catch (InterruptedException | IllegalArgumentException e) {
+            System.out.println("oops!");
```
slarse, line 10: To fix violation '2142:sorald/Main.java:9:17:9:66', Sorald suggests the following:
```suggestion
$(hunk.additions.join('
'))
```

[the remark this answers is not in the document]
slarse: I mean, that's not right, but we're getting somewhere :D

slarse:
For some flabbergasting reason, `git diff` produced the following diff on GHA:

```diff
 /usr/bin/git diff -U0
diff --git a/src/main/java/sorald/Main.java b/src/main/java/sorald/Main.java
index 0af4c5d..108992f 100644
--- a/src/main/java/sorald/Main.java
+++ b/src/main/java/sorald/Main.java
@@ -9 +9 @@ public class Main {
-        } catch (InterruptedException | IllegalArgumentException e) {
+        } catch (java.lang.InterruptedException | java.lang.IllegalArgumentException e) {
@@ -10,0 +11,3 @@ public class Main {
+            if (e instanceof InterruptedException)
+                Thread.currentThread().interrupt();
+
```

I have no idea of why it makes two hunks out of that, it doesn't happen for me locally.

slarse:
@monperrus AHA, great success! See the two suggestions posted above by the GHA bot. It's not a bug in buildbreaker that one fix got posted as two suggestions, because the patch actually looks like this with sorald-0.1.0 (which is still the default in buildbreaker):

```diff
             throw new InterruptedException();
-        } catch (InterruptedException | IllegalArgumentException e) {
+        } catch (java.lang.InterruptedException | java.lang.IllegalArgumentException e) {
             System.out.println("oops!");
+            if (e instanceof InterruptedException)
+                Thread.currentThread().interrupt();
+
         }
 
         int exitStatus = Cli.createCli().execute(args);
```

While sorald-0.2.0 performs this patch instead:

```diff
             throw new InterruptedException();
         } catch (InterruptedException | IllegalArgumentException e) {
             System.out.println("oops!");
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
         }
 
         int exitStatus = Cli.createCli().execute(args);
```

While we should of course use the latest version of Sorald in buildbreaker, this does highlight one thing we will want to do: merge diff hunks that are very close and relate to the same violation specifier. I.e. we should strive for one suggestion per violation specifier, even if that violation spawns multiple diff hunks.

algomaster99:
Closing this due to inactivity and I think its purpose has been served.

PR #535, opened 2021-05-18
feat: Add UnusedLocalVariableProcessor (SonarSource rule 1481)

slarse:
This PR adds a processor for the rule [S1481: Unused local variables should be removed](https://rules.sonarsource.com/java/RSPEC-1481).

Example transformation:

```diff
 public static void main(String[] args) {
-    for (int x = 0, y = 0, z = 10; x <= z; x++) { // Noncompliant, y is not used
+    for (int x = 0, z = 10; x <= z; x++) { // Noncompliant, y is not used
        System.out.println("Current: " + x);
        System.out.println("Goal: " + z);
    }
}
```

This processor is mostly to validate the new matching strategy for local variables introduced with #524. It's not actually that useful a processor as the dead store processor (rule 1854) covers most cases of unused local variables. Just not this one with the for loop header, which is a tricky one for the matching strategy. But it seems to work!


fermadeiral:
LGTM. Thank you, @slarse!

PR #536, opened 2021-05-19
feat: Add `--resolve-classpath-from` to repair command

### LINKED ISSUE #533: Add `--resolve-classpath-from` argument
A problem that I often run into with Maven mode is that I can't repair _only_ production code, or _only_ test code. It would be nice to be able to supply e.g. `--source src/main/java --resolve-classpath-from .` to resolve the classpath from the current directory (using Maven, or Gradle, or whatever build system) and then building a model from `src/main/java`.

This may or may not supplant the Maven mode entirely. There's still value to letting Maven figure out where the source files are for automatic use, but for manual use it's much more convenient to be able to specify them.

slarse:
Fix #533 

This PR adds the `--resolve-classpath-from <path-to-project-root>` option to the repair command. It works precisely the same way as for the mine command, and resolves the classpath from the provided directory path. Currently only works for Maven projects.

While this allows us to run Spoon in noclasspath mode in most cases, I chose to disable it for now. The reason is that segment mode won't work in noclasspath mode even with the classpath provided, as this classpath only includes dependencies, and then parts of the project itself would be missing in each segment. A solution would be to require the project to be compiled before-hand, and include the project's own bytecode on the classpath as well, but I feel that's too much to squeeze into this single PR.

### DIFF
--- a/src/main/java/sorald/Repair.java
+++ b/src/main/java/sorald/Repair.java
@@ -54,13 +54,18 @@ public class Repair {
 
     final List<SoraldEventHandler> eventHandlers;
     private final CompilationUnitCollector cuCollector;
+    private final List<String> classpath;
 
-    public Repair(SoraldConfig config, List<? extends SoraldEventHandler> eventHandlers) {
+    public Repair(
+            SoraldConfig config,
+            List<String> classpath,
+            List<? extends SoraldEventHandler> eventHandlers) {
         this.config = config;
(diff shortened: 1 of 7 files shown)


slarse:
Note: force push as I accidentally based this branch off #531 instead of master :)

fermadeiral:
LGTM. Thanks, @slarse.

PR #542, opened 2021-05-20
doc: Fix outdated release link

slarse:
The link in the README that points to the release is hard-coded to 0.1.0. This PR points it to the `latest` endpoint instead, which updates automatically with each new release.


fermadeiral:
Thanks @slarse.

Issue #522, opened 2021-05-07
BestFitScanner doesn't handle joint declarations well (affects DeadStoreProcessor (rule 1854))


slarse:
Maybe you already have small reproducible examples and I am not sure if this is even related but running `java -jar sorald-0.2.0-jar-with-dependencies.jar repair --source=Test.java --rule-key=1854` (Unused assignments should be removed) on

```java
public class Test {
    public static void main(String[] args) {
        String x = "hello", y = "world";
        System.out.println(x);
    }
}
```
yields:
```java
public class Test {
    public static void main(String[] args) {String y = "world";
        String x;
        System.out.println(x);
    }
}
```

_Originally posted by @lyxell in https://github.com/SpoonLabs/sorald/issues/337#issuecomment-834429422_

slarse:
There are two problems here:

There's a missing newline after the opening curly brace. I thought I fixed that in https://github.com/INRIA/spoon/pull/3856, but apparently I didn't, at least not for statement lists.

The joint declaration is strangely modeled in Spoon. It's modeled as two separate `CtVariable`, but with the exact same source position. As we match Sonar rule violations to Spoon elements by best positional match, and these will always tie, it will simply pick the first one. So we need to expand the matching rules in order to accommodate for joint declarations.

lyxell:
<blockquote>The joint declaration is strangely modeled in Spoon. It's modeled as two separate CtVariable, but with the exact same source position. As we match Sonar rule violations to Spoon elements by best positional match, and these will always tie, it will simply pick the first one. So we need to expand the matching rules in order to accommodate for joint declarations.</blockquote>

Would it be impossible to change this in Spoon? Also, how is the source position represented by Spoon? Is it by line/column number, token index or file offset?

What code is handling the matching of Sonar rule violations to Spoon elements?

lyxell:
Given a declaration like this, Sonar seems to report the following columns:
```
String x = "hello", y = "world";
         ^^^^^^^^^    ^^^^^^^^^
```

Is `getOriginalSourceFragment` given by `Spoon` really unstable?

Otherwise, maybe `CtVariables` could get special treatment in `BestFitScanner` and instead of relying on `elem.getSourceStart()` use something like `elem.getSourceStart() + elem.getOriginalSourceFragment().indexOf(elem.getSimpleName())`.

lyxell:
Another heuristic that might work is searching for `getSimpleName` from the ends of the fragment returned by Sonar and expanding it:
[...]

PR #545, opened 2021-05-20
fix(1854): Handle dead stores in expression position

### LINKED ISSUE #541: Processor for S1854 (DeadStore) misses variable references
The processor for S1854 does not recognize that a variable can be read and written to at the same time, leading to compile errors. Minimal example:
```java
int x = 10;
int y = x++;
x = 20;
System.out.println(y);
```

transformed into:

```java
int x = 10;
int y;
System.out.println(y);
```

slarse:
Fix #541 

This PR incorrect deletions when dead stores appear in expression position, rather than statement position. As far as I know, this applies only to suffix operators (e.g. `int y = x++`) and expression assignments (e.g. `int y = (x = 1)`).

@lyxell wdyt? Can you break it?


lyxell:
Have you tried running the DeadStoreCheck from `sonar-java`?

https://github.com/SonarSource/sonar-java/blob/master/java-checks-test-sources/src/main/java/checks/DeadStoreCheck.java

slarse:
> Have you tried running the DeadStoreCheck from sonar-java?

That's what's used to detect the dead store. The best fits scanner we dabbled with before is the component that matches rule violations (warnings) raised by `sonar-java` to elements in the Spoon AST.

> It would be nice to see a test for preincrement

Preincrement does not result in a dead store when used in expression position. I.e. the actual store from the preincrement itself isn't a dead store, as it returns the same value as is stored in the variable. The entire statement can of course still be a dead store, and there's an unhandled case there: the statement being a dead store, but the expression-store isn't dead. For example:

```diff
 int x = 10;
-int y = ++x; // dead store in y
 System.out.println(x);
```

The repair is incorrect because it changes the behavior of the program. So we need to also look out for live stores in expressions that are part of dead store statements. Complicated :)

> assignment operator like +=

It didn't even occur to me that you could put composite assignment operators in expression position... I will try that, they may cause shenanigans.

lyxell:
> That's what's used to detect the dead store. The best fits scanner we dabbled with before is the component that matches rule violations (warnings) raised by `sonar-java` to elements in the Spoon AST.

Check the link again! I mean the actual test, it has a few examples of noncompliant preincrements. But yes, it is pretty confusing that they named the test like that.

slarse:
> > That's what's used to detect the dead store. The best fits scanner we dabbled with before is the component that matches rule violations (warnings) raised by `sonar-java` to elements in the Spoon AST.
> 
> Check the link again! I mean the actual test, it has a few examples of noncompliant preincrements. But yes, it is pretty confusing that they named the test like that.

Oh, rofl. But yeah, none of the noncompliant preincrements are in expression position, they are all in statement position. As a side note, we actually don't handle them in statement position (nor do we handle postincrements in statement position, oopsie!), but that's due to a bug that's completely unrelated to stores in expressions. I'll open a separate issue about that.

lyxell:
> Oh, rofl. But yeah, none of the noncompliant preincrements are in expression position, they are all in statement position.

Oh, right. Then I guess the only problem is the other assignment operators, does this approach break for them?

slarse:
Okay, so it appears as if Sonar does not flag violations on composite assignment operators in expression position (e.g. `+=`, `*=` etc). That's probably a bug to file with `sonar-java`. I'll leave a sentinel in the test case (see below) so we can detect if this is fixed when we upgrade `sonar-java`.

```java
    void expressionPlusEquals() {
        int x = 1;
        int y = (x += 10); // Compliant; false negative by Sonar, left as sentinel if it becomes positive
        System.out.println(y);
    }
```

slarse:
Live store expressions within dead store statements exceeds the position matching precision. I managed to get this repair:

```diff
    void expressionPlusEquals() {
        int x = 10;
-       int y = (x += 1);
+      int y = 1;
        System.out.println(x);
    }
```

Although obviously, the dead store is in the `y` variable, the position reported by Sonar corresponds to `= (x += 1);`, which in terms of position overlap is a better match for `x += 1` than it is for `int y = (x += 1)`. So the position matching does what it's supposed to do correctly, it's just not the right thing to do here. We may need to reconsider what elements the `DeadStoreProcessor` considers, perhaps it shouldn't match on statements at all as Sonar always flags the _expression_ that's assigned as the dead store.

Anyway, that kind of blows the scope of this PR out of proportion, so I think we should stop here. @lyxell wdyt?

lyxell:
> Anyway, that kind of blows the scope of this PR out of proportion, so I think we should stop here. @lyxell wdyt?

Yes, probably. Although, going back to the earlier discussion, I am still feeling a bit uneasy about the assumptions made when CtRole is not STATEMENT. I am not really sure when they are assigned but there's a few in there that looks a bit scary given the current logic, I think.

Have you considered https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/path/CtRole.html#ELSE and https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/path/CtRole.html#FORUPDATE for example?

slarse:
> > Anyway, that kind of blows the scope of this PR out of proportion, so I think we should stop here. @lyxell wdyt?
> 
> Yes, probably. Although, going back to the earlier discussion, I am still feeling a bit uneasy about the assumptions made when CtRole is not STATEMENT. I am not really sure when the roles are assigned but when reading through them there's a few in there that looks a bit scary given the current logic, I think.
> 
> Have you considered https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/path/CtRole.html#ELSE and https://spoon.gforge.inria.fr/mvnsites/spoon-core/apidocs/spoon/reflect/path/CtRole.html#FORUPDATE for example?

```java
if (true) { } else x++;

if (...) x++;
```

These `x++` are in statement position (the branches of an `if` are always statements). As such, Sorald currently doesn't do anything due to the bug outlined in #546.

```java
for (int y; y < 10; x++)
```

That one _would_ be a problem, but Sonar does not consider dead stores in for loop headers (as discussed in https://github.com/SpoonLabs/sorald/issues/522#issuecomment-838116295). We should probably add a test sentinel for that if Sonar suddenly starts to detect them. Can you think of any other expression position that isn't actually read?

lyxell:
> That one _would_ be a problem, but Sonar does not consider dead stores in for loop headers (as discussed in [#522 (comment)](https://github.com/SpoonLabs/sorald/issues/522#issuecomment-838116295)).

Here's a minimal counterexample:

```java
class Test {
    public static void main(String args[]) {
        int y;
        for (;; y = 30) {}
        y = 40;
        return y;
    }
}
```

slarse:
> > That one _would_ be a problem, but Sonar does not consider dead stores in for loop headers (as discussed in [#522 (comment)](https://github.com/SpoonLabs/sorald/issues/522#issuecomment-838116295)).
> 
> Here's a minimal counterexample:
> 
> ```java
> class Test {
>     public static void main(String args[]) {
>         int y;
>         for (;; y = 30) {}
>         y = 40;
>         return y;
>     }
> }
> ```

Interesting, so they consider pure assignments but no other kinds of stores. That's very inconsistent.

The question then is: how to generically figure out that an expression is actually read? Or do we need to hard-code it? Or is it _only_ for loop updates where an expression can appear without being read?

lyxell:
> The question then is: how to generically figure out that an expression is actually read? Or do we need to hard-code it?

This is tricky I think, I guess reads propagate through the AST so a graph search is probably necessary.

> Or is it _only_ for loop updates where an expression can appear without being read?

I think at least normal statements, for loop updates, for loop inits, casts and instanceof expressions. Probably a few more places too.

slarse:
> This is tricky I think, I guess reads propagate through the AST so a graph search is probably necessary.

I mean, it's an AST (heavily augmented, but still an AST), so there's no data flow built-in. Although there is a [spoon-control-flow module](https://github.com/INRIA/spoon/tree/master/spoon-control-flow), I wouldn't trust it as it hasn't been maintained for years. And it's version `0.0.2`.

I don't think there's any other choice than to hard-code the cases where an unread expression is possible.
 
> > Or is it _only_ for loop updates where an expression can appear without being read?
> 
> I think at least normal statements, for loop updates, for loop inits, casts and instanceof expressions. Probably a few more places too.

So the for loops are given, but what do you mean about casts and instanceof? Are you talking about the new [pattern matching](https://docs.oracle.com/en/java/javase/14/language/pattern-matching-instanceof-operator.html)?

Anyway, I think we've thought this through enough. I'll fix the applicable for-loop headers problems (and if you can explain what you mean about casts and instanceof, those too). We don't aim for Sorald's processors to be theoretically perfect, and so if there's an extremely seldom used corner case we don't cover, that doesn't matter until it actually poses a problem.

We typically go with the "make it, break it, fix it" approach, where the "break it" part is done through large-scale experiments. Then we find crashes and failed repairs, add test cases for them, and fix them.

slarse:
Okay, so now that I've been playing around, Sonar appears to detect dead stores in loop headers no problem, both in init and update. Tried `y++`, `++y` and `y += 1`. Maybe I messed something up last time.

slarse:
Fun fact: Sorald actually does the correct repair for dead stores in loop headers, but that's because it crashes with a class cast exception when it tries to do the replacement and only manages to do the deletion.

slarse:
@lyxell I think I've come to a realization. A pure expression in Java _must_ be read, but a [_statement expression_](https://docs.oracle.com/javase/specs/jls/se14/html/jls-14.html#jls-StatementExpression) does not need to be read (see the link for the exact elements that are statement expressions). This means that the places that allow an expression not to be read, actually only allows statement expressions (which in Spoon are elements that implement both `CtExpression` and `CtStatement`).

If this is correct (I'm relatively certain), then when replacing the dead store (e.g. `x++`) with just the returned value (in this case, a variable read from `x`), it should be sufficient to simply check that a variable read is assignable to the position where we want to put it. For example, in a for init, a variable read is _not_ assignable because the for init accepts only statement expressions and local variable declarations.

I hope that makes sense, I'm currently putting it into code with a bunch of test cases for loops but there appears to be some complications with the printer (it's printing deleted values ...).

lyxell:
> If this is correct (I'm relatively certain), then when replacing the dead store (e.g. `x++`) with just the returned value (in this case, a variable read from `x`), it should be sufficient to simply check that a variable read is assignable to the position where we want to put it. For example, in a for init, a variable read is _not_ assignable because the for init accepts only [statement expressions and local variable declarations](https://docs.oracle.com/javase/specs/jls/se14/html/jls-14.html#jls-ForInit), and a variable read is neither of those. Same goes for e.g. `x += 2` in the for update, as that only accepts statement expressions, but a literal `2` is not a statement expression.
> 
> I hope that makes sense, I'm currently putting it into code with a bunch of test cases for loops but there appears to be some complications with the printer (it's printing deleted values ...).

Right. You mean that all of the invalid repairs like:
```diff
- for (int y; y < 10; x++)
+ for (int y; y < 10; x)
```
can be recognized by the fact that they will actually introduce an invalid AST node/syntax error. Reading the grammar again this makes sense. Nice find. Seems like the only places then actually is where a normal statement could occur as well as in for inits and for updates?

slarse:
> > If this is correct (I'm relatively certain), then when replacing the dead store (e.g. `x++`) with just the returned value (in this case, a variable read from `x`), it should be sufficient to simply check that a variable read is assignable to the position where we want to put it. For example, in a for init, a variable read is _not_ assignable because the for init accepts only [statement expressions and local variable declarations](https://docs.oracle.com/javase/specs/jls/se14/html/jls-14.html#jls-ForInit), and a variable read is neither of those. Same goes for e.g. `x += 2` in the for update, as that only accepts statement expressions, but a literal `2` is not a statement expression.
> > I hope that makes sense, I'm currently putting it into code with a bunch of test cases for loops but there appears to be some complications with the printer (it's printing deleted values ...).
> 
> Right. You mean that all of the invalid repairs like:
> 
> ```diff
> - for (int y = 0; y < 10; x++)
> + for (int y = 0; y < 10; x)
> ```
> 
> can be recognized by the fact that they would actually introduce an invalid AST node/syntax error.

Precisely. This is easy to check in Spoon, without any catering to specific cases.

> Seems like the only places then actually is where a normal statement could occur as well as in for inits and for updates?

Well, normal statements are separate and to be handled with #546. I'm also sure there are other places than for inits and updates where expression statements are allowed, but if my reasoning is correct and we can simply fall back on syntax, then it doesn't matter.

It will make more sense once I commit the code, but unfortunately there's a bug in the Sniper printer (it doesn't detect deletions in for loop headers), so I need to fix that first and then get back to this PR sometime next week.

lyxell:
> I'm also sure there are other places than for inits and updates where expression statements are allowed

Looking at the grammar, the production:

```bnf
StatementExpression ::=
    Assignment
    PreIncrementExpression
    PreDecrementExpression
    PostIncrementExpression
    PostDecrementExpression
    MethodInvocation
    ClassInstanceCreationExpression
```

can only be reached from three locations, the normal statement position (as an expression statement), from *for init* and from *for updates*.

> normal statements are separate and to be handled with #546

Then there is actually only *for updates* and *for inits* left where this can occur.

slarse:
> > I'm also sure there are other places than for inits and updates where expression statements are allowed
> 
> Note that in the grammar expression statements and statement expressions are different things.
> Looking at the grammar, the relevant production:

Yes, sloppy terminology on my part. Everywhere I've used the words "statement" and "expression" in succession (in any order), I've meant statement expression.

> can only be reached from **three** locations, the normal statement position (as part of an _expression statement_), from _for inits_ and from _for updates_.

How did you figure this out? I think the Java grammar as presented in the JLS is too fragmented to get a good overview. Did you find a page somewhere with only the grammar? To verify what you wrote I did a search through the PDF of the JLS, and found that `StatementExpression` only does appear in those places you mentioned. But searching like that only works for very distinct terms.

lyxell:
> How did you figure this out? I think the Java grammar as presented in the JLS is too fragmented to get a good overview. Did you find a page somewhere with only the grammar? To verify what you wrote I did a search through the PDF of the JLS, and found that `StatementExpression` only does appear in those places you mentioned. But searching like that only works for very distinct terms.

They usually publish the whole grammar in a separate section in the online version of the spec, e.g. https://docs.oracle.com/javase/specs/jls/se15/html/jls-19.html

slarse:
> They usually publish the whole grammar in a separate section in the online version of the spec, e.g. https://docs.oracle.com/javase/specs/jls/se15/html/jls-19.html

Ah, that's great. Thanks!

src/test/resources/processor_test_files/1854_DeadStore/SonarFalseNegatives.java
```diff
@@ -0,0 +1,25 @@
...
+            y++;
+        }
+    }
+
+    void expressionPlusEquals() {
+        int x = 1;
+        int y = (x += 10); // Compliant; dead store in x, false negative by sonar
+        System.out.println(y);
+    }
+}
```
slarse, line 25: Note these false negatives. We keep them in this test case to notice if Sonar starts to notice them in future updates.

slarse:
@lyxell Spoon patch is in and this works now. Have a look if you'd like, and then we'll merge this.

lyxell:
Cool! Looks good to me.

slarse:
@fermadeiral ping for merge :)

fermadeiral:
Thank you, @slarse and @lyxell!

Issue #540, opened 2021-05-19
Processor for S2095 (UnclosedResources) introduces new resource leak


lyxell:
In the following code there is a resource leak for `bw2`:

```java
import java.io.BufferedWriter;
import java.io.File;
import java.io.FileWriter;
import java.io.IOException;
import java.util.*;

class Test {
    public static void saveTo(File file1, File file2) {
        BufferedWriter bw1 = null;
        BufferedWriter bw2 = null;
        try {
            bw1 = new BufferedWriter(new FileWriter(file1));
            bw1.write("Write some stuff to file 1");

            bw2 = new BufferedWriter(new FileWriter(file2));
            bw2.write("Write some stuff to file 2");

        } catch (IOException e) {
            e.printStackTrace();
        } finally {
            System.out.println("done");
            try {
                if (bw1 != null) {
                    bw1.close();
                }
            } catch (IOException e) {
                e.printStackTrace();
            }
        }
    }
}
```

The processor UnclosedResourcesProcessor for S2095 seems to just remove the whole finally-block when creating the try-with-resources statement. While this behavior fixes the resource leak for `bw2` it introduces a new resource leak for `bw1` and also changes the output of the program. Here's the result:

```java
import java.io.BufferedWriter;
import java.io.File;
import java.io.FileWriter;
import java.io.IOException;
import java.util.*;

class Test {
    public static void saveTo(File file1, File file2) {
        BufferedWriter bw1 = null;
        try (BufferedWriter bw2 = new BufferedWriter(new FileWriter(file))) {
            bw1 = new BufferedWriter(new FileWriter(file));
            bw1.write("Write some stuff to file 1");
            bw2.write("Write some stuff to file 2");
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}
```

This is related to but different from #386 I think.

Edit: Running it one more time on the output from the previous run introduces a compile error:

```java
import java.io.BufferedWriter;
import java.io.File;
import java.io.FileWriter;
import java.io.IOException;
import java.util.*;

class Test {
    public static void saveTo(File file1, File file2) {
        try (BufferedWriter bw1 = new BufferedWriter(new FileWriter(file1))) {
            bw1.write("Write some stuff to file 1");
            bw2.write("Write some stuff to file 2");
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}
```

slarse:
[...]

PR #550, opened 2021-05-28
feat(2095): Handle multiple unclosed closeables in same try

### LINKED ISSUE #386: UnclosedResourcesProcessor: handle multiple closeables
Problematic patch: https://github.com/khaes-kth/Sorald-CI/blob/fixed_LianjiaTech/retrofit-spring-boot-starter_a6cfd41658170f108be0391fd26c353650d90d75_2095_1613019018315/fixed_repo/SoraldGitPatches/soraldpatch_0.patch

I'll elaborate on this later ...

slarse:
Fix #386 

This PR is a minor extension to the `UnclosedResourcesProcessor` that simply makes it respect an existing try-with-resources, and add resources to it rather than delete it and create a new one (which would cause resources to be lost; only the very last resource processed would ever appear in the resource block of the try). By extension, this allows the processor to deal with multiple unclosed resources in the same try-block. Here's how that works:

1. There's one violation for each unclosed resource
2. The first violation is fixed first (due to iteration order of the AST), and creates a try-with-resources block, with any remaining violating elements being placed within it
3. Any further violations are fixed simply by deleting the violating element and adding it the the (now existing) try-with-resources
    - This is the code that corresponds to the patch in this PR.

There are a gazillion corner cases which this does not cover. I will keep improving this processor with tiny PRs every once in a while. Next up is #540.

### DIFF
--- a/src/main/java/sorald/processor/UnclosedResourcesProcessor.java
+++ b/src/main/java/sorald/processor/UnclosedResourcesProcessor.java
@@ -49,7 +49,11 @@ public class UnclosedResourcesProcessor extends SoraldAbstractProcessor<CtConstr
 
         CtBlock parentCtBlock = parent.getParent(CtBlock.class);
         boolean isInTry = parentCtBlock.getParent() instanceof CtTry;
-        if (isInTry) {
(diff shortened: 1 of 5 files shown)


fermadeiral:
@slarse thank you!

PR #555, opened 2021-06-02
fix(3984): Retain simply qualified name of throwable type

### LINKED ISSUE #554: UnusedThrowableProcessor creates fully qualified reference
The repair actually looks like this:

```diff
-new IllegalArgumentException();
+throw new java.lang.IllegalArgumentException();
```

For once, this is a bug in Sorald, and not in Spoon.

slarse:
Fix #554 

This PR fixes a problem where the `UnusedThrowableProcessor` always creates a fully qualified name for the throwable, even when a simply qualified name is used initially. I.e. the repair looks like so:

```java
-new IllegalArgumentException();
+throw new java.lang.IllegalArgumentException();
```

With this fix, the repair looks as intended:

```java
-new IllegalArgumentException();
+throw new IllegalArgumentException();
```

In doing the exact match test to check that the printed reference is indeed simply qualified, I also discovered that comments on the thrown expression must be moved to the `CtThrow`, otherwise they are printed like this before the semicolon:

```java
throw new IllegalArgumentException()// here's a comment
;
```

And that just looks very strange.

### DIFF
--- a/src/main/java/sorald/processor/UnusedThrowableProcessor.java
+++ b/src/main/java/sorald/processor/UnusedThrowableProcessor.java
@@ -1,17 +1,24 @@
 package sorald.processor;
 
 import sorald.annotations.ProcessorAnnotation;
+import spoon.reflect.code.CtComment;
 import spoon.reflect.code.CtConstructorCall;
-import spoon.reflect.code.CtThrow;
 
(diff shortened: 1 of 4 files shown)


fermadeiral:
Nice fix, thank you @slarse.

PR #556, opened 2021-06-02
fix(1854): Fix handling of unary operators in statement position

### LINKED ISSUE #546: DeadStoreProcessor (rule 1854) does not delete prefix/suffix operators in statement position
A dead store can take the form of a prefix/suffix operator in statement position (e.g. a standalone `x++` or `--y`), but the processor just skips over those. It should delete the entire statement.

slarse:
Fix #546 

This PR is a very minor extension to the `DeadStoreProcessor` that makes it just delete unary operators in statement position, as their return values can't possibly be read (because they're in statement position!).

When I opened #546 such unary operators were just ignored, but now the actually cause a crash. So this PR is doubly needed!


fermadeiral:
LGTM.

PR #559, opened 2021-06-07
fix(2095): Retain finally-block from try when converting to try-with-resources

### LINKED ISSUE #540: Processor for S2095 (UnclosedResources) introduces new resource leak

slarse:
Fix #540 (this is actually supposed to fix #540, not a mistake this time!)

This PR changes the `UnclosedResourcesProcessor` to retain the finally block when converting a try to a try-with-resources. Previously, it would just drop the finally block.

### DIFF
--- a/src/main/java/sorald/processor/UnclosedResourcesProcessor.java
+++ b/src/main/java/sorald/processor/UnclosedResourcesProcessor.java
@@ -34,17 +34,18 @@ public class UnclosedResourcesProcessor extends SoraldAbstractProcessor<CtConstr
         CtTryWithResource tryWithResource = getFactory().createTryWithResource();
         tryWithResource.addResource(variable);
 
-        CtBlock<?> parentCtBlock = parent.getParent(CtBlock.class);
-        boolean isInTry = parentCtBlock.getParent() instanceof CtTry;
-        boolean isInTryWithResource = parentCtBlock.getParent() instanceof CtTryWithResource;
-        if (isInTryWithResource) {
-            ((CtTryWithResource) parentCtBlock.getParent()).addResource(variable);
+        CtBlock<?> enclosingBlock = parent.getParent(CtBlock.class);
+        CtElement enclosingBlockParent = enclosingBlock.getParent();
+        if (enclosingBlockParent instanceof CtTryWithResource) {
+            ((CtTryWithResource) enclosingBlockParent).addResource(variable);
             parent.delete();
-        } else if (isInTry) {
+        } else if (enclosingBlockParent instanceof CtTry) {
+            CtTry enclosingTry = (CtTry) enclosingBlockParent;
+            tryWithResource.setCatchers(enclosingTry.getCatchers());
+            tryWithResource.setBody(enclosingBlock);
+            tryWithResource.setFinalizer(enclosingTry.getFinalizer());
+            enclosingTry.replace(tryWithResource);
             parent.delete();
-            tryWithResource.setCatchers(((CtTry) parentCtBlock.getParent()).getCatchers());
-            parentCtBlock.getParent().replace(tryWithResource);
-            tryWithResource.setBody(parentCtBlock);
         } else {
             encloseResourceInTryBlock((CtStatement) parent, tryWithResource);
         }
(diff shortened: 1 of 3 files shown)


slarse:
This PR introduces another problem, namely that the finally block may now contain a close of an auto-managed resource (i.e. the resource appears in the resource list _and_ it is closed in the finally block). For example, coding like this is pretty common:

```java
    public static void saveTo(File file1, File file2) {
        BufferedWriter bw1 = null;
        BufferedWriter bw2 = null;
        try {
            bw1 = new BufferedWriter(new FileWriter(file1));
            bw2 = new BufferedWriter(new FileWriter(file2));
        } catch (IOException e) {
            e.printStackTrace();
        } finally {
            try {
                if (bw1 != null) {
                    bw1.close();
                }
                if (bw2 != null) {
                    bw2.close(); // Noncompliant; bw1.close() could throw and so we might not get here
                }
            } catch (IOException e) {
                e.printStackTrace();
            }
        }
    }
```

`bw1.close()` could throw, and so `bw2.close()` may never be reached, causing SonarJava to flag it as unclosed. The repair then becomes this (whith this PR's version of Sorald):

```java
import java.io.BufferedWriter;
import java.io.File;
import java.io.FileWriter;
import java.io.IOException;
import java.util.*;

class Main {
        saveTo(new File("hello.txt"), new File("byebye.txt"));
    }

    public static void saveTo(File file1, File file2) {
        BufferedWriter bw1 = null;
        try (BufferedWriter bw2 = new BufferedWriter(new FileWriter(file2))) {
            bw1 = new BufferedWriter(new FileWriter(file1));
        } catch (IOException e) {
            e.printStackTrace();
        } finally {
            try {
                if (bw1 != null) {
                    bw1.close();
                }
                if (bw2 != null) {
                    bw2.close();// Noncompliant; bw1.close() could throw and so we might not get here

                }
            } catch (IOException e) {
                e.printStackTrace();
            }
        }
    }
}
```

Now, `bw2` is an auto-managed resource, but it's _also_ closed in the finally block (as the finally block is no longer removed). Calling `close()` twice generally isn't a problem, the second call is just a NOOP, but it doesn't look nice. I think we should "clean" the finally block appropriately, but IMO that should be handled in a separate PR. I just wanted to make note of the problem.

fermadeiral:
> I think we should "clean" the finally block appropriately, but IMO that should be handled in a separate PR.

I agree with you.

For this PR, it looks good to me. Thank you, @slarse.

PR #564, opened 2021-06-09
fix: Clean up out-of-scope references to auto-managed resource

### LINKED ISSUE #561: UnclosedResourcesProcessor (rule 2095): Retaining finally block can cause "double close" of resources
This PR introduces another problem, namely that the finally block may now contain a close of an auto-managed resource (i.e. the resource appears in the resource list _and_ it is closed in the finally block). For example, coding like this is pretty common:

```java
    public static void saveTo(File file1, File file2) {
        BufferedWriter bw1 = null;
        BufferedWriter bw2 = null;
        try {
            bw1 = new BufferedWriter(new FileWriter(file1));
            bw2 = new BufferedWriter(new FileWriter(file2));
        } catch (IOException e) {
            e.printStackTrace();
        } finally {
            try {
                if (bw1 != null) {
                    bw1.close();
                }
                if (bw2 != null) {
                    bw2.close(); // Noncompliant; bw1.close() could throw and so we might not get here
                }
            } catch (IOException e) {
                e.printStackTrace();
            }
        }
    }
```

`bw1.close()` could throw, and so `bw2.close()` may never be reached, causing SonarJava to flag it as unclosed. The repair then becomes this (whith this PR's version of Sorald):

```java
import java.io.BufferedWriter;
import java.io.File;
import java.io.FileWriter;
import java.io.IOException;
import java.util.*;

class Main {
        saveTo(new File("hello.txt"), new File("byebye.txt"));
    }

    public static void saveTo(File file1, File file2) {
        BufferedWriter bw1 = null;
        try (BufferedWriter bw2 = new BufferedWriter(new FileWriter(file2))) {
            bw1 = new BufferedWriter(new FileWriter(file1));
        } catch (IOException e) {
            e.printStackTrace();
        } finally {
            try {
                if (bw1 != null) {
                    bw1.close();
                }
                if (bw2 != null) {
                    bw2.close();// Noncompliant; bw1.close() could throw and so we might not get here
                }
            } catch (IOException e) {
                e.printStackTrace();
            }
        }
    }
}
```

Now, `bw2` is an auto-managed resource, but it's _also_ closed in the finally block (as the finally block is no longer removed). Calling `close()` twice generally isn't a problem, the second call is just a NOOP, but it doesn't look nice. I think we should "clean" the finally block appropriately, but IMO that should be handled in a separate PR. I just wanted to make note of the problem.

_Originally posted by @slarse in https://github.com/SpoonLabs/sorald/issues/559#issuecomment-856015810_

slarse:
Fix #561 

This PR makes the UnclosedResourcesProcessor clean up references to the variable that was inlined into the resource block. Previously, such left-over references were not cleaned up, causing compile errors.

Here's an example of the new-and-improved repair, note how it removes the `if` in the catcher:

```diff
 public class ReferenceInCatcherAndFinalizerToFieldWithSameNameAsResource {
     private int bw = 2;

     public void saveTo(File file) throws IOException {
-        BufferedWriter bw = null;
-        try {
-            bw = new BufferedWriter(new FileWriter(file)); // Noncompliant
+        try (BufferedWriter bw = new BufferedWriter(new FileWriter(file))) {
             bw.write("Write some stuff to file");
         } catch (IOException e) {
             e.printStackTrace();
             System.out.println(this.bw); // reference to field; should not be cleaned!
-            if (bw != null) { // reference to local variable; should be cleaned!
-                bw.close(); // reference to local variable; should be cleaned!
-            }
         } finally {
             System.out.println(this.bw); // reference to field; should not be cleaned!
             System.out.println("done");

         }
     }
 }
```

It works by looking for references to the inlined variable inside the catchers and finalizer, and removing any statement containing a reference. It's possible that this could "over clean" for example by removing an `if` statement where the condition has a reference to the resource, but there's also unrelated stuff inside of the  then or else blocks. But I think we'll deal with that only if we actually find it happening, as I've said before, this processor will never be perfect.

### DIFF
--- a/src/main/java/sorald/processor/UnclosedResourcesProcessor.java
+++ b/src/main/java/sorald/processor/UnclosedResourcesProcessor.java
@@ -12,6 +12,7 @@ import spoon.reflect.code.CtTry;
 import spoon.reflect.code.CtTryWithResource;
 import spoon.reflect.code.CtVariableWrite;
 import spoon.reflect.declaration.CtElement;
+import spoon.reflect.reference.CtLocalVariableReference;
 import spoon.reflect.reference.CtVariableReference;
 
 @ProcessorAnnotation(key = 2095, description = "Resources should be closed")
@@ -30,14 +31,14 @@ public class UnclosedResourcesProcessor extends SoraldAbstractProcessor<CtConstr
         }
     }
 
-    private void createCtTryWithResource(CtElement parent, CtLocalVariable variable) {
+    private void createCtTryWithResource(CtElement parent, CtLocalVariable<?> variable) {
         CtTryWithResource tryWithResource = getFactory().createTryWithResource();
-        tryWithResource.addResource(variable);
+        tryWithResource.addResource(variable.clone());
 
         CtBlock<?> enclosingBlock = parent.getParent(CtBlock.class);
         CtElement enclosingBlockParent = enclosingBlock.getParent();
         if (enclosingBlockParent instanceof CtTryWithResource) {
-            ((CtTryWithResource) enclosingBlockParent).addResource(variable);
+            ((CtTryWithResource) enclosingBlockParent).addResource(variable.clone());
             parent.delete();
         } else if (enclosingBlockParent instanceof CtTry) {
             CtTry enclosingTry = (CtTry) enclosingBlockParent;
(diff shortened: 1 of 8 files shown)


fermadeiral:
LGTM.

PR #566, opened 2021-06-11
test: Verify that processor test resources compile

### LINKED ISSUE #565: Separately test that all resources that should compile do compile
The test resources in `processor_test_files` are implicitly tested to compile when running the Maven test, but the error output isn't great.

We should add two separate sanity checks: one that checks that all input files compile, and one that checks that all of the `.expected` files compile.

slarse:
Fix #565 

This PR adds a test that runs through all Java input files and `.expected` Java files in `src/test/resources/processor_test_files` that are not prefixed and with `NOCOMPILE` and tries to compile them. If compilation fails, the test reports output like so:

```java
org.opentest4j.AssertionFailedError: /tmp/junit5557181061285140448/InsecureTransformerFactoryUsage.java:11: error: cannot find symbol
        TransformerFactory transformerFactory = createTrnsformerFactory();
                                                ^
  symbol:   method createTrnsformerFactory()
  location: class InsecureTransformerFactoryUsage ==> 
Expected :true
Actual   :false
```

Previously, the only verification we had that test resources compiled like they should was one of the Maven mode tests, and that would fail with a pretty non-obvious error message. It also did not check the `.expected` files, only the input `.java` files.

### DIFF
--- a/src/test/java/sorald/processor/ProcessorTestFilesCompileTest.java
+++ b/src/test/java/sorald/processor/ProcessorTestFilesCompileTest.java
@@ -1,0 +1,68 @@
+package sorald.processor;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import javax.tools.Diagnostic;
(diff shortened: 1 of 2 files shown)


fermadeiral:
LGTM.

PR #574, opened 2021-06-16
feat: Add CollectionsEmptyConstantsProcessor (SonarSource rule 1596)

fermadeiral:
This PR adds a processor for the rule ["Collections.EMPTY_LIST", "EMPTY_MAP", and "EMPTY_SET" should not be used](https://rules.sonarsource.com/java/RSPEC-1596).

Example repair:

```diff
- List<String> collection1 = Collections.EMPTY_LIST;  // Noncompliant
- Map<String, String> collection2 = Collections.EMPTY_MAP;  // Noncompliant
- Set<String> collection3 = Collections.EMPTY_SET;  // Noncompliant
+ List<String> collection1 = Collections.emptyList();
+ Map<String, String> collection2 = Collections.emptyMap();
+ Set<String> collection3 = Collections.emptySet();
```


src/main/java/sorald/processor/CollectionsEmptyConstantsProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+        description =
+                "\"Collections.EMPTY_LIST\", \"EMPTY_MAP\", and \"EMPTY_SET\" should not be used")
+public class CollectionsEmptyConstantsProcessor extends SoraldAbstractProcessor<CtFieldAccess<?>> {
+
+    @Override
+    protected void repairInternal(CtFieldAccess<?> element) {
+        CtType<?> collectionsType = getFactory().Class().get(Collections.class);
+
+        CtMethod<?> methodToBeCalled;
+        if (element.getVariable().getSimpleName().equals("EMPTY_LIST")) {
```
slarse, line 20: On a side note, this violates the rule you implemented a processor for in #562  :)

(sorald-buildbreaker does not catch it because it uses Sorald 0.3.0, which doesn't have that processor)

slarse (changes requested):
One comment on the code that I think could be simpler/less hard-coded.

Apart from my comment about simplifying the processor a bit, there's one other thing that crossed my mind. Right now, the target of the method invocation is computed by directly getting a reference to `Collections`. But what if `Collections` isn't actually imported, and it's used with a fully qualified name? That would lead to this repair, which could cause a compile error:

```java
// original
List<Integer> list = java.util.Collections.EMPTY_LIST;
// repair
List<Integer> list = Collections.emptyList();
```

However, the check in Sonar is a bit lazily implemented and only checks if the full name of the target is `Collections`, thus it doesn't flag qualified accesses to `java.util.Collections`. But perhaps it will in the future, so you could perhaps add a line like so in the .java and .expected files.

```java
List<Integer> list = java.util.Collections.EMPTY_LIST; // Compliant due to false positive
```

That way we catch it if they fix the problem (which we should potentially report).

src/main/java/sorald/processor/CollectionsEmptyConstantsProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+            methodToBeCalled = collectionsType.getMethodsByName("emptyList").get(0);
+        } else if (element.getVariable().getSimpleName().equals("EMPTY_MAP")) {
+            methodToBeCalled = collectionsType.getMethodsByName("emptyMap").get(0);
+        } else {
+            methodToBeCalled = collectionsType.getMethodsByName("emptySet").get(0);
+        }
```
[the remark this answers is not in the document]
fermadeiral: That's a great suggestion! Applied!

src/main/java/sorald/processor/CollectionsEmptyConstantsProcessor.java
fermadeiral: :X

fermadeiral:
@slarse, suggestions applied.

slarse (approved):
Nice, short and simple processor :)

PR #575, opened 2021-06-16
fix: Fix test errors caused by compile output in test resources

### LINKED ISSUE #573: Deployments are failing
The deployments are failing as, for some reason, the tests aren't passing in the deploy environment.

slarse:
Fix #573 

The cause of #573 was that the compile test introduced in #566 produced .class files in the test resources directory. These class files would then cause the surefire plugin to crash for some reason. I don't know why it crashes because of that, but the class files shouldn't be there either way.

This also explains why only the deploy job failed: it runs the tests twice, once when running `package` and once when running `deploy` (we should probably remove the `package` target). All of the other jobs run the tests only once, and thus don't encounter the problem.

The reason this happened was that there are two overloads of the `ProcessorTestHelper.getTestCaseStream()` method. One of them returns the test cases for the resources in a directory (this was the one that was used), and the other one copies all test resources to a temporary directory and then returns test cases based on that (this is the one that should have been used).

This PR changes the tests to use the method that copies the resources to a temporary directory first, and also renames that method and updates its documentation to avoid this confusion in the future.

### DIFF
--- a/src/test/java/sorald/GatherStatsTest.java
+++ b/src/test/java/sorald/GatherStatsTest.java
@@ -31,7 +31,7 @@ public class GatherStatsTest {
     public void statisticsFile_containsExpectedStats(
             RepairStrategy repairStrategy, @TempDir File tempDir) throws Exception {
         ProcessorTestHelper.ProcessorTestCase<?> testCase =
-                ProcessorTestHelper.getTestCaseStream()
+                ProcessorTestHelper.getTestCasesInTemporaryDirectory()
                         .filter(tc -> tc.ruleKey.equals("2755"))
                         .findFirst()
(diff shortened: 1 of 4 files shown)


slarse:
@fermadeiral @khaes-kth Urgent to merge to fix deployment failures, first to the party please review and merge :)

If you've run any tests locally since yesterday, you probably have class files lying around your test resources. You'll want to delete those to avoid test strangeness.

You can clean the resources directory like so:

```bash
git clean -xf src/test/resources
```

Note that this removes _any_ untracked file, so make sure to commit everything you want to commit before doing it.

fermadeiral:
LGTM.

PR #576, opened 2021-06-17
refactor: Extract not method into Transformations utility class

### LINKED ISSUE #572: Create a SpoonUtils utility class with reusable Spoon transformations
The method `not(CtExpression<T> expr)` is a copy from another processor. We should have something like `SpoonUtils` to keep these utility methods together and accessible to all processors. I believe this should be handled in a different PR though.

_Originally posted by @fermadeiral in https://github.com/SpoonLabs/sorald/issues/571#issuecomment-862243110_

To summarize that comment, some Spoon transformations are reused across different processors. This is a concrete example, but I'm sure we can find other transformations that are duplicated (either as parts of methods or methods of their own).

slarse:
Fix #572 

This PR adds a new `Transformations` utility class, which is meant to house reusable high-level transformation methods for use across the entire project. If we look through the project, I think we'll find many more recurring transformations that can be extracted into this class.

Note that this class is specifically for transformations. I think that if we want reusable Spoon-"something else", we should create another class with an appropriate name for that.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -44,6 +44,11 @@ When merging a PR, maintainers should take care to:
 
 For adding a new processor in Sorald, please, follow the instructions described below.
 
+0) Familiarize yourself with the [`sorald.util` utility package](#reusable-utility-classes-and-methods)
+
+This is to ensure that you do not recreate functionality that we already have a
+solution for.
+
 1) Find the name for the new processor
 
 The first step is to find the name for the new processor.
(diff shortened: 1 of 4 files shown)


slarse:
Ping @fermadeiral

fermadeiral:
@slarse, I'm sorry for the delay, I saw your request before but then I forgot to do the review.

My review: the content of the PR is nice. What about adding something in the contributing instructions regarding this? It could be just like one sentence, saying that the person adding a new processor should look at the `Transformations` class while coding the new processor. I mean, a new contributor might code things inside the new processor that already exist in the utility class, which is just a waste of time.

slarse:
@fermadeiral No worries, this isn't exactly a crucial PR.

I've updated `CONTRIBUTING.md` with clear references to the `sorald.util` package, and the `Transformations` class in particular.

fermadeiral:
LGTM.

PR #577, opened 2021-06-17
refactor: Replace use of ImmutableList with List.copyOf

slarse:
#331 

The package containing `ImmutableList` is a transitive dependency of Sorald through SonarJava. It's not there anymore in SonarJava 7+, and so this PR removes the two uses of it. The same functionality is available as unmodifiable lists in the standard library, so I don't think it's a good idea to pull in a dependency just for this.


khaes-kth (approved):
LGTM!

PR #578, opened 2021-06-17
refactor: Update to sonar java 7

### LINKED ISSUE #331: Update to sonar-java-7.0.0.26422

slarse:
Fix #331 

I'm currently at the point where it runs, but it fails tons of tests as many warnings that were previously raised are not anymore, and that the post filter functionality has change so we don't respect `SuppressWarnings` at the moment.


slarse:
Per my musings in #331, this isn't what we want to do. We should first migrate to using the SonarLint APIs, and then upgrade the version of SonarJava.

PR #581, opened 2021-06-21
refactor: Use full string key for Sonar rules

### LINKED ISSUE #580: Use full SonarJava key in all internal parts and as input
Required for #579 

Currently, for some reason the key used externally is an int (without the `S` prefix), while internally in places we must prepend the S for things to work out. Additionally, some parts use the key as an `int` while others use a `String`.

In order to proceed with #579, all key usages must be with the full string identifier.

slarse:
Fix #580 

This PR fixes the inconsistent use of rule keys internally in Sorald, where in some places an integer was used, and in others a string. Additionally, the full key is now used (e.g. `"S1854"` instead of `"1854"`). This is due to needing the full key when interacting with Sonar, and translating back and forth is both a hassle and error prone. It also doesn't make much sense not to use the full key, it's just an element of potential confusion for users.

The implementation is backwards compatible for now. That is to say, if someone were to type `--rule-key 1854` or provide a violation specifier with the "shortened" rule key on the command line, that would still work.


slarse:
@khaes-kth Any chance you can have a look at this today? It's a blocker for #582

khaes-kth:
@slarse 
Sure. I will do it now.

khaes-kth (approved):
LGTM.

slarse:
:+1:

Issue #579, opened 2021-06-21
Refactor internals to use SonarLint core API instead of internal SonarJava API


slarse:
As discussed in #331, the best way forward is to make use of the public API in SonarLint core. This will allow us to upgrade the SonarJava implementation without breakage with every single upgrade. I'll make this refactoring in two phases.

### Encapsulate all Sonar APIs into the `sorald.sonar` subpackage
There's currently widespread use of certain Sonar parts, most notable being the `JavaFileScanner` and `JavaCheck` classes. These should all be encapsulated into the `sorald.sonar` subpackage, and we should instead use our own facades in other parts of Sorald.

### Migrate to use the SonarLint core API
Once `sorald.sonar` nicely encapsulates all of the Sonar stuff, the migration should be relatively straightforward. I've tested in a separate project that it's possible to use the SonarLint API for our purposes, and in fact it's a lot more straightforward. There's significantly less tweaking to do to make it work than what we currently have to do with the SonarJava internals.

slarse:
@monperrus @fermadeiral @khaes-kth At several points, we've discussed the fact that we don't actually need to use SonarJava to detect rule violations. In theory, we could use any static analyzer. However, SonarJava is currently wired into Sorald quite tightly. In migrating to the SonarLint API, I need to poke around large parts of the project, and therefore if we want to have the option of using other static analyzers in the future now would be a good time to decide how to do that.

There are two viable options.

1. Encapsulate the rule definitions and violations into generalizable interfaces, and then wire in SonarLint as one implementation of that
    - Downside: Sorald becomes more complex.
    - Upside: Sorald becomes more self-contained and relies less on Sonar (at least in theory). Also does not exclude option 2 (see below).
2. Explicitly depend on SonarLint in larger parts of the codebase, and wire in additional static analyzers as SonarLint plugins
    - Downside: A maintainer of Sorald must also understand SonarLint plugins in order to add a new static analyzer. This is especially bad if we want to _implement_ static analysis with Spoon.
    - Upside: Sorald itself becomes less complex, and delegates the concern of static analysis to somewhere else

I'm leaning toward option 1. simply because it's good practice to encapsulate external dependencies as much as possible, and the fact that it doesn't actually exclude option 2.

monperrus:
Ack. LGTM, encapsulation with proper, documented interfaces is good.

slarse:
[...]

PR #582, opened 2021-06-22
refactor: Encapsulate SonarJava in sorald.sonar subpackage

slarse:
#579 

This PR completely encapsulates all Sonar components into the `sorald.sonar` package, with the corresponding generalizations being placed in the `sorald.rule` package.

With this design, I managed to replace the current use of SonarJava with SonarLint core, without changing anythign outside of the `sorald.sonar` package. So, the isolation is pretty good. There are still some leaky abstractions, but further work to support other static analyzers than Sonar should be relatively small. More importantly, we can now customize how we use Sonar in particular without having to rewire the entire application.

I'm still not sure _how_ to proceed with SonarJava, but after the problems I found in using SonarLint (see #583), I'm leaning towards sticking to SonarJava but using some higher-level APIs. Regardless, this PR facilitates any approach we might take with Sonar in the future.

This PR is pretty big but it's rather hard to split as it's a fundamental redesign. Give reviewing it a go and let me know if something is unclear.

@fermadeiral @khaes-kth whoever has the time to review this go ahead :)


khaes-kth (approved):
It certainly LGTM!

fermadeiral:
LGTM too.

slarse:
Let's merge then :)

PR #583, opened 2021-06-28
refactor: Use SonarLint core in Sorald

slarse:
Playing around with using SonarLint instead of directly using SonarJava. It's... slower.


slarse:
There seems to be two strong reasons not to use SonarLint:

* Startup of the core components before running analysis is somewhat costly, around 500-800 ms
    - Can be somewhat alleviated through clever caching, but the startup cost must be incurred at least once during a repair
* SonarJava must be provided as a separate jarfile. This makes it rather inconvenient to create a single, standalone jar for Sorald. One solution would be to download the SonarJava plugin if it can't be found on disk, but that's somewhat odd as we still need SonarJava as a dependency of the project to access some information.

algomaster99:
@slarse I needed help with https://github.com/SpoonLabs/sorald/issues/659 and your work on this PR is related. I am assuming this is the second task of https://github.com/SpoonLabs/sorald/issues/579 where you try to use SonarLint APIs and then configure Sonar Java on top of it [because it is the intended way to do it](https://github.com/SpoonLabs/sorald/issues/331#issuecomment-864901712) and won't break anything from `6.9` to `7`. But you have also outlined the problems above of configuring SonarLint APIs. How do you think should we proceed then?

> SonarJava must be provided as a separate jarfile.

If a MAVEN dependency is not accepted by SonarLint APIs, we have to resort to this. I think it is okay to expect a user to manually provide the jar file of the static analyzer. Suppose we have 2 more static analyzers in the future, we will ask the user to give a path to the jar file of the supported static analyzers. We can, however, recommend which version should be used.

> Startup of the core components before running analysis is somewhat costly, around 500-800 ms
> Can be somewhat alleviated through clever caching, but the startup cost must be incurred at least once during a repair

If we directly update from Sonar Java 6 -> 7, there are A LOT of changes in the Sonar Java API and that is why you suggested [stepping up to SonarLint in the first place](https://github.com/SpoonLabs/sorald/issues/331#issuecomment-864879178). Reading from investigations, I think it is necessary to have SonarLint so that so many breaking changes do not occur when upgrading SonarJava which I think would take place very often for Sorald.

slarse:
> If a MAVEN dependency is not accepted by SonarLint APIs

It is not.

> Reading from investigations, I think it is necessary to have SonarLint so that so many breaking changes do not occur when upgrading SonarJava which I think would take place very often for Sorald.

Historically, every single update has introduced breaking changes, because we use things that are not meant to be used for anything but internals. So from that perspective, using SonarLint instead is dead simple.

But I really don't have much to add to this discussion. The two major downsides remain, namely the hit to running time and the fact that the JAR needs to be provided. This PR is essentially complete as a POC, except that it doesn't include the JAR file and now obviously has some conflicts to resolve.

algomaster99:
If you ever tried upgrading SonarJava to `7`, did you document what issues you were facing? @khaes-kth and I can then decide what should be the best way to go about it.

slarse:
> If you ever tried upgrading SonarJava to `7`, did you document what issues you were facing? @khaes-kth and I can then decide what should be the best way to go about it.

You mean when using SonarJava directly as a library? I did try it, see #578

algomaster99:
> You mean when using SonarJava directly as a library?

Yes. I tried it on my own as well. I was definitely facing a lot of problems because a lot of types have been changed. A lot of tests also failed.

I will take this up with @khaes-kth, @monperrus, and @fermadeiral .

algomaster99:
Reopened at #696.

PR #584, opened 2021-07-06
feat: Add processor for deleting unused private fields (S1068)

algomaster99:
This PR adds processor for [Unused "private" fields should be removed](https://rules.sonarsource.com/java/RSPEC-1068).

Example repair is written in the HANDLED_RULES.md and pushed as a test case in this PR.


src/test/resources/processor_test_files/S1068_UnusedPrivateField/UnusedPrivateField.java
```diff
@@ -0,0 +1,12 @@
+import java.io.Serializable;
+
+public class UnusedPrivateField implements Serializable {
+    private static final long serialVersionUID = 42L; // Compliant as this field is used in deserialization
```
algomaster99, line 4: This field was not deleted even though I **did not** implement the class using `Serializable` initially. I think this could be a bug in the sonar scanner for violations. What do you think, @monperrus @slarse ?

slarse: You could have a look inside of the `UnusedPrivateFieldCheck` and see how they select fields. Possibly, they just ignore any field with that name.

slarse (changes requested):
Processor LGTM, but could you also add a test case to test joint declarations along these lines?

```java
public class UnusedPrivateFieldInJointDeclaration {
    int used, unused;

    public void printUsed() {
        System.out.println(used);
    }
}
```

src/test/resources/processor_test_files/S1068_UnusedPrivateField/UnusedPrivateField.java
algomaster99: [It](https://github.com/SonarSource/sonar-java/blob/a5947abe43d79a638c61413613b21f417fa9f400/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateFieldCheck.java#L144-L147) checks for modifiers apart from checking for name. I could propose if they want check for class implementation too but they can probably dismiss this proposal as trivial because `serialVersionUID` is a semantically reserved word in Java.

algomaster99:
@slarse Hold the review.

algomaster99:
@slarse The tests failed for joint declarations. I could see that the violation for `unused` private field was being returned by sonar [here](https://github.com/SpoonLabs/sorald/blob/master/src/test/java/sorald/Assertions.java#L70) while debugging so I am suspecting that the `BestFitScanner` is not able to identify the corresponding `CtElement`.

slarse:
> @slarse The tests failed for joint declarations. I could see that the violation for `unused` private field was being returned by sonar [here](https://github.com/SpoonLabs/sorald/blob/master/src/test/java/sorald/Assertions.java#L70) while debugging so I am suspecting that the `BestFitScanner` is not able to identify the corresponding `CtElement`.

I suspected that would happen. Let's have a look at this together during the meeting Thursday, we've had this problem before with local variables.

slarse:
From meeting (starting point to improve best fits scanner):

```
diff --git a/src/main/java/sorald/sonar/BestFitScanner.java b/src/main/java/sorald/sonar/BestFitScanner.java
index c09f0341..44953b8c 100644
--- a/src/main/java/sorald/sonar/BestFitScanner.java
+++ b/src/main/java/sorald/sonar/BestFitScanner.java
@@ -181 +181 @@ public class BestFitScanner<E extends CtElement> extends CtScanner {
-            String ident =
+            String precedingIdentifier =
@@ -183 +183,6 @@ public class BestFitScanner<E extends CtElement> extends CtScanner {
-            return ident.equals(((CtVariable<?>) element).getSimpleName());
+            String identifierAtViolationPosition =
+                    getIdentifierFromViolationPosition(
+                            violation, element.getPosition().getCompilationUnit());
+
+            String simpleName = ((CtVariable<?>) element).getSimpleName();
+            return Set.of(precedingIdentifier, identifierAtViolationPosition).contains(simpleName);
@@ -188,0 +194,17 @@ public class BestFitScanner<E extends CtElement> extends CtScanner {
+
+    private static String getIdentifierFromViolationPosition(
+            RuleViolation violation, CtCompilationUnit cu) {
+        int sourceStart =
+                calculateSourcePos(
+                        violation.getStartLine(),
+                        violation.getStartCol(),
+                        cu.getLineSeparatorPositions());
+        int sourceEnd =
+                calculateSourcePos(
+                        violation.getEndLine(),
+                        violation.getEndCol(),
+                        cu.getLineSeparatorPositions());
+        String cuSource = cu.getOriginalSourceCode();
+        return cuSource.substring(sourceStart, sourceEnd + 1).strip();
+    }
+
```

algomaster99:
@slarse I added the test cases which we discussed in the meeting and fortunately, the patch you came up with worked for all of the cases. :)

slarse (approved):
Lucky break, I suppose!

Well, LGTM :)

algomaster99:
I just realised that I forgot to add the rule [here](https://github.com/SpoonLabs/sorald/blob/master/README.md#usage). Shall do it in the next PR.

slarse:
We forget that with just about every other processor. IMO either we should auto-generate that listing, or we should remove the listing of processors there. We have the `HANDLED_RULES.md` document after all.

So, I mean... I wouldn't bother.

algomaster99:
> we should remove the listing of processors there

I agree with this. Clients would prefer to see the listing on the terminal only. Would be better to just describe what `--rule-key=<ruleKey>` flag is for and explain its syntax.

Issue #367, opened 2021-02-05
Automatically generate `HANDLED_RULES.md`


slarse:
This isn't urgent, but I think we should generate the `HANDLED_RULES.md` file. The descriptions of the processors can be written in the processor source file, and everything else can simply be generated.

This will ensure that documentation stays up-to-date, make it easier to add new processors, and also provide a reference for the operation of a processor right in the source file.

algomaster99:
> The descriptions of the processors can be written in the processor source file, and everything else can simply be generated.

I was thinking we could add test cases for each rule using the resources inside `processor_test_files` but that might make the length of the description excessively long given the number of test cases for some of the sonar rules. I just thought to put forth this idea anyway.

Also, how shall we classify each rule into Sonar rule types (`BUG`, `CODE_SMELL`, etc.)? Since there is no segregation in `sorald/processors`, should we consider adding another attribute in `ProcessorAnnotation` for marking the violation type?

algomaster99:
I am thinking the python script for generating `HANDLED_RULES.md` should go inside `experimentation/tools` so that it can be executed as `python3 -m sorald.handled_rules`. However, writing a Python script would require string parsing of docstrings and the attributes inside `@ProcessorAnnotation`. We could make it easier by adding keywords in comments so that the required text is easier to find in the file but still it could be tough to keep the format consistent across all processors.

Another solution could to write a Java script (no pun intended) so that we are able to use `spoon`. This would make it really easy to access the comments over types and the attributes of `@ProcessorAnnotation`. But there are cons with this approach too.
1. We use jinja for rendering templates in python so we would have to look for corresponding library in Java. In the worst case, we would have to manually generate the entire markdown file.
2. This script is essentially a chore and all chores are packaged as a python package under `experimentation/tools` so it seems odd to ostracize the script in question from the package.

slarse:
> Also, how shall we classify each rule into Sonar rule types (BUG, CODE_SMELL, etc.)?

We can get that from the [rule metadata](https://github.com/SonarSource/sonar-java/tree/master/java-checks/src/main/resources/org/sonar/l10n/java/rules/java). We fetch the rule metadata for another thing, you can [extract that functionality into the `_helpers` package](https://github.com/SpoonLabs/sorald/blob/6452f48125050f7f94e9e76d9cb1390fa25845f9/experimentation/tools/sorald/prmessage.py#L66-L69) and reuse it.

[...]

PR #585, opened 2021-07-16
fix: modify hyperlink in HANDLED_RULES.md to correctly link list item to its description

algomaster99:
I accidentally put the wrong hyperlink in #584 . This PR corrects that.


slarse:
Thanks @algomaster99, I didn't catch that. We definitely need to generate that file, it's just way too error prone to edit it manually :)

algomaster99:
I agree too. I  have some ideas/suggestions about it. I will articulate them tomorrow on issue - https://github.com/SpoonLabs/sorald/issues/367.

slarse:
Cool! My only thought thus far was, as mentioned in #367, to write the descriptions in the files. As some sort of comment, that is to say. But I really haven't thought too much about it.

PR #588, opened 2021-08-10
chore: Add script for generating HANDLED_RULES.md

### LINKED ISSUE #367: Automatically generate `HANDLED_RULES.md`

algomaster99:
Fixes #367 

Since this PR is quite huge, I will explain the flow of the script I have written.

#### Usage

```sh
python3 -m sorald.handled_rules -o <output_file>
```

This script invokes the main function of `sorald/experimentation/tools/sorald/handled_rules.py` which in turn executes `sorald/experimentation/tools/scripts/GetKeyAndDescription.java`.

#### Purpose of `GetKeyAndDescription.java`

This script creates a Spoon model for the entire `processor` package and returns a JSON object of rule key and description.
```json
[
{"rule_key": 1068, "repair_description": "A long elaborate description of repair as outlined in HANDLED_RULES.md"},
...
]
```

We could also have used python only to extract the docstrings (description) and rule key using RegEx maybe. However, I felt that might be prone to error since some processors had either one or two annotations (`ProcessorAnnotation` and `IncompleteProcessor`). Moreover, we would have to carefully parse the docstrings to avoid "*". Such inconsistencies and technicalities made me want to use Spoon to extract this AST information.

#### `handled_rules.py`

Takes in the JSON object and renders the `HANDLED_RULES,md`. The `HANDLED_RULES,md` in this PR is automatically generated. :)

#### Enhancements and ToDos
- [ ] Write tests for this script
- [ ] Render headings of `HANDLED_RULES,md` in a for-loop itself. In other words, `Bug` heading shouldn't be rendered if no sorald doesn't have any repairs for any kind of bugs. This is mainly important if in future, we ever implement a repair for one of the [Security Hotspots](https://rules.sonarsource.com/java/type/Security%20Hotspot).
- [ ] Update README for `tools` package.


experimentation/tools/scripts/GetKeyAndDescription.java
```diff
@@ -0,0 +1,51 @@
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.util.Set;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
```
algomaster99, line 7: I am not sure about using these packages. I googled how to return a JSON object and got this as the answer. Luckily, these packages were in the classpath.

experimentation/tools/scripts/GetKeyAndDescription.java
```diff
@@ -0,0 +1,51 @@
...
+        allProcessors.stream().forEach(processor -> {
+            // Skips SoraldAbstractProcessor
+            if (processor.isAbstract()) return;
+
+            JsonObject rule = new JsonObject();
+
+            // Remove 'S' prefix from rule key
+            CtLiteral<String> ruleKey = processor.getAnnotations().stream()
+                    .filter(annotation -> annotation.getType().getSimpleName().equals("ProcessorAnnotation"))
+                    .findFirst().get().getValue("key");
```
algomaster99, line 32: A shorter way to find annotation would have been `getAnnotation(ProcessorAnnotation.class)` but this requires us to add sorald jar in the classpath too. I didn't want to edit the currently generated `cp.txt` so I went with this approach.

experimentation/tools/sorald/handled_rules.py
```diff
@@ -0,0 +1,203 @@
...
+ENCODING = "utf8"
+
+BUG = []
+BUG_DETAIL = []
+
+CODE_SMELL = []
+CODE_SMELL_DETAIL = []
+
+VULNERABILITY = []
+VULNERABILITY_DETAIL = []
```
algomaster99, line 25: This can be combined to form two lists - `List` and `Detail`. Both would render for all repairs for all violation types. But that can be an enhancement for another PR.

experimentation/tools/sorald/handled_rules.py
```diff
@@ -0,0 +1,203 @@
...
+{% endfor %}
+"""
+
+OUTPUT_ARG = "--output"
+
+PATH_TO_CLASSPATH = pathlib.Path(__file__).absolute().parent.parent.parent.parent / "cp.txt"
+with open(PATH_TO_CLASSPATH, 'r') as cp:
+    CLASSPATH = cp.read()
+
+PATH_TO_JAVA_SCRIPT = join(dirname(dirname(abspath(__file__))), "scripts", "GetKeyAndDescription.java")
```
algomaster99, line 75: Using `pathlib` to resolve the path was throwing errors while executing the `JAVA` script. Somehow, the java CLI was interpreting `PosixPath`'s string representation a class and not a source file path. In other words, it was interpreting it as - `sorald.experimentation.tools.scripts.GetKeyAndDescription.java`

experimentation/tools/sorald/handled_rules.py
```diff
@@ -0,0 +1,203 @@
...
+
+    else:
+        raise Exception(f"New rule violation type, {violation_type}, encountered.")
+
+
+def get_link_to_detail(title: str, rule_key: int) -> str:
+    link = title.lower()
+    link = re.sub(r"\s", "-", link)
+    link = re.sub(r"[^A-Za-z0-9_-]", "", link)
+    return f"#{link}-sonar-rule-{rule_key}"
```
algomaster99, line 195: This function is to create in-page links for the repairs. I wrote this based on general observation. I am not sure about the exact conventions of how HTML links headers in pages.

algomaster99:
I think we would also need to add an exemption for docstrings in spotless. Spotless is adding HTML tags that escape the markdown syntax.

slarse:
You probably need to put the markdown stuff into `<code>` tags to avoid Javadoc freaking out about bad HTML.

slarse:
Alternatively, an easier solution to avoid problems with Spotless and Javadoc might be to just create a markdown file for each processor to contain the documentation. For example, `DeadStoreProcessor.java` would be accompanied by `DeadStoreProcessor.md`.

algomaster99:
> That also eliminates the need to use Spoon in the script, which simplifies the CI.

Right. We can then get the rule key corresponding to each processor using simple RegEx.

experimentation/tools/sorald/handled_rules.py
slarse: The links are generated by GitHub markdown. It's typically just the content of the heading in lowercase, with spaces replaced by dashes. For example `# Hello there` can be linked to as `[Go to hello there](#hello-there)`. Some special symbols are removed, but I don't know off the top of my head which ones those are.

Anyway, the gist of it is that a link to a given section heading in the Markdown file should be generated from the text in the heading.

experimentation/tools/scripts/GetKeyAndDescription.java
slarse: Gson is used in Sorald to produce the stats files.

experimentation/tools/sorald/handled_rules.py
slarse: In general, mutable globals is a one-way ticket to bug town. For example, what happens if you import this script and run it twice? It also suggests that there's a function somewhere with a side effect of mutating these globals, which is also a bad smell.

I'd make these a return value of some function instead and use that where appropriate.

experimentation/tools/sorald/handled_rules.py
algomaster99: > Some special symbols are removed, but I don't know off the top of my head which ones those are.

Apart from `_` and `-`, I have removed everything else. If a special case arises, I will handle it then.

experimentation/tools/sorald/handled_rules.py
algomaster99: Yeah, I was thinking about this too. I initially thought I would write `handled_rules.py` as a class so that I can contain such private variables and then return when classification is over.

experimentation/tools/sorald/handled_rules.py
slarse: This must be due to some other error on your end. If the Java CLI interpreted a string as dot-separated, that's because the string you passed to it was dot-separated.

algomaster99: I will try to deal with this again and if I still face a problem, we can discuss this in the meeting tomorrow.

Anyway, we would no longer need this because we are removing the spoon script.

experimentation/tools/sorald/handled_rules.py
algomaster99: Okay. So you are suggesting that I pass the entire heading text instead of passing the title and rule key separately?

Also, I didn't know about `remove_special_chars`. I will use that instead in this function.

slarse: > Okay. So you are suggesting that I pass the entire heading text instead of passing the title and rule key separately?

Yes, because that's what's used to generate the link. You should be able to pass an arbitrary heading to it. It's also easier to understand as there are no magic words like in the current function.

`remove_special_chars` isn't a defined function, I was just to lazy to copy-paste your regex.

slarse: Edited, that ought to work.

algomaster99:
@slarse Hold the review. There are some things I need to refactor in the script because I have hard-coded a few things in there.

algomaster99:
@slarse I have made the required changes. I have sorted the output based on the python default string comparison. However, I lowecased the title before comparison.

I wanted some advice regarding tests. Should we create a dummy processor package in `tests/resources` and then test the entire thing at once?

algomaster99:
@slarse Which python formatter has been used to maintain the `tools` package? I think I would need to use it given the amount of code I have written.

src/test/java/sorald/processor/ProcessorDescriptionTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class ProcessorDescriptionTest {
+    private static final Path PROCESSOR_PACKAGE = Paths.get("src/main/java/sorald/processor");
+
+    @Test
+    public void test_eachProcessorIsAccompaniedByDescription() {
+        List<File> processors =
+                Arrays.stream(PROCESSOR_PACKAGE.toFile().listFiles())
+                        .filter(file -> file.getName().endsWith(".java"))
+                        .collect(Collectors.toList());
```
slarse, line 22: This is a little bit shaky, we can do better as Sorald knows all of the processors.

Add the following method to `Processors.java`:

```java
    /**
     * @return A list of all processors sorted by name.
     */
    public static List<Class<? extends SoraldAbstractProcessor<?>>> getAllProcessors() {
        return RULE_KEY_TO_PROCESSOR.values().stream()
                .sorted(Comparator.comparing(Class::getName))
                .collect(Collectors.toList());
    }
```

The sorting by name is just for consistency across invocations, makes testing more predictable.

Then rewrite this collection like so:

```suggestion
        List<File> processors = Processors.getAllProcessors().stream()
                .map(Class::getSimpleName)
                .map(procName -> PROCESSOR_PACKAGE.resolve(procName + ".java"))
                .map(Path::toFile);
```

slarse (changes requested):
Mostly looks great! I have a few notes on simplifications and improvements, but not much at all.

In addition to my comments on the code, the following would also be good:

1. Add a comment (i.e. a quote or something, with `>`) at the top of `HANDLED_RULES.md` saying that it is a generated file that should not be edited manually, and link to the script.
2. Update `CONTRIBUTING.md` with the fact that one should add a `BlaBla.md` file when creating a `BlaBla.java` processor.

As we don't want Java contributors to have to deal with Python, we'll simply update `HANDLED_RULES.md` in CI like we update `ACHIEVEMENTS.md`, see `.github/workflows/support.yml`. But we'll add that in a separate PR.

src/main/java/sorald/processor/EqualsArgumentTypeProcessor.md
```diff
@@ -0,0 +1,13 @@
...
+Example:
+
+```diff
+
+@Override public boolean equals(Object obj) {
++       if (obj == null || getClass() != obj.getClass()) {
++               return false;
++       }
+return toFile().equals(((SpoonResource) obj).toFile());
+}
```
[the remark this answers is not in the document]
algomaster99: This is weird because the style check test pass even though I agree the formatting is off.

experimentation/tools/sorald/handled_rules.py
```diff
@@ -0,0 +1,222 @@
...
+
+@dataclasses.dataclass
+class ViolationDetail:
+    title: str
+    sonar_url: str
+    sonar_url_text: str
+    repair_description: str
+
+    def __lt__(self, other):
+        return self.title.lower() < other.title.lower()
```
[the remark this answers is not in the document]
algomaster99: > I don't see a reason for separating them, just create one dataclass with all of the information.

I separated them because the list with headings didn't have `repair_description` and the list of `repair_description` didn't require the link to itself so I decided to segregate them. If we keep them together, we can use the same list to render both the lists so that's an advantage for sure.

slarse:
> @slarse Which python formatter has been used to maintain the `tools` package? I think I would need to use it given the amount of code I have written.

Missed this question. I use [black](https://github.com/psf/black) for everything.

```bash
pip install black
black experimentation/tools
```

src/main/java/sorald/processor/EqualsArgumentTypeProcessor.md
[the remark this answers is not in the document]
algomaster99: Oops. I thought this was a Java file. :sweat_smile: I will fix it.

src/test/java/sorald/processor/ProcessorDescriptionTest.java
```diff
@@ -0,0 +1,39 @@
...
+                        .map(Path::toFile)
+                        .collect(Collectors.toList());
+        processors.forEach(
+                processor ->
+                        assertTrue(
+                                getDescription(processor).isFile(),
+                                "Description corresponding to "
+                                        + processor.getName()
+                                        + " does not exist."));
+    }
```
slarse, line 32: As a final improvement here, I'd suggest making this a parameterized test instead of a single test with a loop. IMO, it's almost always preferable to do this when feasible as there's zero risk of for example there being no matching files (a parameterized test with empty parameterization throws an exception). In this case, I also think it increases the readability of the test.

A simple way to do it would be like so:

```suggestion
    @ParameterizedTest
    @MethodSource("processorFileProvider")
    public void test_eachProcessorIsAccompaniedByDescription(File processor) {
        assertTrue(
                getDescription(processor).isFile(),
                "Description corresponding to " + processor.getName() + " does not exist.");
    }

    private static Stream<Arguments> processorFileProvider() {
        return Processors.getAllProcessors().stream()
                .map(Class::getSimpleName)
                .map(procName -> PROCESSOR_PACKAGE.resolve(procName + ".java"))
                .map(Path::toFile)
                .map(Arguments::of);
    }
```

slarse (changes requested):
I suggest one clarification in the docs and a final improvement to the test (as much to show you another possibility as for making the test cleaner), then it's ready to merge.

docs/CONTRIBUTING.md
```diff
@@ -104,8 +104,9 @@ run `mvn spotless:apply`, and then commit any changes the formatter makes.
 5) Update documentation
 
 Your processor is done, it's passing the minimum tests, so now you can update the documentation.
-In the `HANDLED_RULES.md` file, you should add the new handled rule in the table of contents, and then a summary of its processor with an example.
-You should do it following the same way as the existing handled rules are documented (note that the table of contents and summaries are alphabetically ordered).
+Write a summary of the processor supporting it with an example in <PROCESSOR_NAME>.md file and place it inside the
+`sorald.processor` package. An example name of such file could be `CastArithmeticOperandProcessor.md` if your
+processor's name is `CastArithmeticOperandProcessor`.
```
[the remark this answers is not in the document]
algomaster99: Since we are enforcing the requirement, we should use 'must' instead of 'should'. The rest of the wording sounds good to me.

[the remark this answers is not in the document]
slarse: Sure.

src/test/java/sorald/processor/ProcessorDescriptionTest.java
algomaster99: > for example there being no matching files (a parameterized test with empty parameterization throws an exception)

Just to clarify: do you mean that if there are no processors, the test with method source would throw an error?

slarse: Yes.

slarse: When asserting in a loop (or stream) inside a test, you always run the risk of a vacuously passing test as the loop/stream is empty. That's why, when asserting in a loop, you should always first assert that collection you're iterating over is non-empty.

algomaster99:
@slarse Any idea how we could customise the display name of the parameterized test? Currently, it shows the full path of the processor for project root (string representation of `File` object of a processor). We can just show the name of the processor because that would be clearer. However, the `name` parameter of `@ParameterizedTest` doesn't let us invoke functions.

slarse:
@algomaster99 Easiest way is to do something like this: https://stackoverflow.com/a/57894201

IMO it's overkill for this test and the fully qualified name to the file is fine.

algomaster99:
@slarse I think we can proceed for merging if there are no more suggestions.

The SO post would introduce a clearer display name but at the expense of clarity of the test function since we won't use the _nicer representation_ inside the test.

slarse (approved):
LGTM, thanks @algomaster99

Issue #589, opened 2021-08-12
Existing Sorald code is not generating remediated code


PrakashNayak:
Steps I followed
*******************************
1. git clone https://github.com/SpoonLabs/sorald.git
2. cd sorald
3. mvn package -DskipTests
4 cp target/sorald-*-jar-with-dependencies.jar sorald.jar

5. Opend the command prompt
Ran below command 
repair --source C:/test/solard_code/sonarqube-repair/source/act/UnusedPrivateField .java --rule-key  1068

6. This is Original file
 public class UnusedPrivateField {
    private String a = "Hello world!";

     public int compute(int number) {
         return number * 42;
     }
 }

Output in console
************************
INFO  JavaClasspath initialization
INFO  JavaClasspath initialization (done) | time=81ms
INFO  JavaTestClasspath initialization
INFO  JavaTestClasspath initialization (done) | time=0ms
INFO  Java Main Files AST scan
INFO  1 source files to be analyzed
INFO  1/1 source files have been analyzed
INFO  Java Main Files AST scan (done) | time=1273ms
INFO  Java Test Files AST scan
INFO  0 source files to be analyzed
INFO  Java Test Files AST scan (done) | time=1ms
INFO  0/0 source files have been analyzed
INFO  Java Generated Files AST scan
INFO  0 source files to be analyzed
INFO  Java Generated Files AST scan (done) | time=1ms
INFO  0/0 source files have been analyzed
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
-----Number of fixes------
UnusedPrivateFieldProcessor: 0
-----End of report------

I want answer of below questions 
**************************************
1 . Why its not removing the unused private variable?
2. Where file will get generated ?
3. Am I doing any mistake in running the command 
[Testing.zip](https://github.com/SpoonLabs/sorald/files/6974839/Testing.zip)

I want to two things here first it should remove the unused variable and remove the unusedprivate fields.

Please help me and let me know @prakash.bit.nayak@gmail.com

slarse:
Hi @PrakashNayak,

> 1 . Why its not removing the unused private variable?

[...]

Issue #335, opened 2021-01-19
Hi I am facing this issue while using repair command 


DnvSaiteja:
java -jar sorald/target/sorald-1.1-SNAPSHOT-jar-with-dependencies.jar repair --original-files-path sorald/src/test/resources/processor_test_files/2095_UnclosedResources/ZipFolder.java --rule-keys 2095 --workspace outputdir

 After executng above command, I am getting the below error

]
INFO  Configured Java source version (sonar.java.source): 14
INFO  JavaClasspath initialization
INFO  JavaClasspath initialization (done) | time=36ms
INFO  JavaTestClasspath initialization
INFO  JavaTestClasspath initialization (done) | time=1ms
INFO  Java Main Files AST scan
INFO  1 source files to be analyzed
INFO  1/1 source files have been analyzed
INFO  Java Main Files AST scan (done) | time=2490ms
INFO  Java Test Files AST scan
INFO  0 source files to be analyzed
INFO  Java Test Files AST scan (done) | time=1ms
INFO  Java Generated Files AST scan
INFO  0/0 source files have been analyzed
INFO  0 source files to be analyzed
INFO  Java Generated Files AST scan (done) | time=0ms
INFO  0/0 source files have been analyzed
java.lang.IllegalArgumentException: 'other' is different type of Path
        at java.base/sun.nio.fs.WindowsPath.relativize(WindowsPath.java:400)
        at java.base/sun.nio.fs.WindowsPath.relativize(WindowsPath.java:42)
        at sorald.event.models.WarningLocation.<init>(WarningLocation.java:16)
        at sorald.event.models.miner.MinedViolationEvent.<init>(MinedViolationEvent.java:24)
        at sorald.Repair.lambda$mineViolations$2(Repair.java:138)
        at java.base/java.lang.Iterable.forEach(Iterable.java:75)
        at sorald.Repair.mineViolations(Repair.java:136)
        at sorald.Repair.getRuleViolations(Repair.java:105)
        at sorald.Repair.repair(Repair.java:88)
        at sorald.cli.RepairCommand.call(RepairCommand.java:121)
        at sorald.cli.RepairCommand.call(RepairCommand.java:27)
        at picocli.CommandLine.executeUserObject(CommandLine.java:1933)
        at picocli.CommandLine.access$1200(CommandLine.java:145)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
[...]

Issue #273, opened 2020-12-07
Add support for Windows


khaes-kth:
There are some tests that do not pass on Windows because of the difference between file paths in windows and linux. For example, `sorald.processor.ProcessorTest`.

* [ ] fix the tests on Windows
* [ ] add Windows CI

slarse:
I wasn't aware that we support Windows. If that's the case, we must also run CI for Windows. It's not just about filepaths, Windows uses a file system that differs quite a lot from those found in Unix-like systems. Permissions are especially a pain, and I've had massive trouble with temporary files and directories before (which Sorald uses a lot both in testing and production).

khaes-kth:
> I wasn't aware that we support Windows

Actually, I'm not sure if we want to support Windows for users either. But it would be nice to, at least, make the tests pass so that developers who user Windows (like me :D) can ensure tests pass locally before pushing it on Github.

slarse:
I think we should discuss this at the meeting. Due to how dependent Sorald is on file system interaction, supporting Windows might be an added cost that we don't want to have.

As for local development, have you considered Windows Subsystem for Linux (2)? You'd definitely be able to run the tests in there.

slarse:
On a side note, [Windows is the most popular dev platform according to the SO survey](https://insights.stackoverflow.com/survey/2020#technology-developers-primary-operating-systems), so if we want to have a lot of end users of Sorald (as opposed to running it with e.g. Repairnator), we should probably at least attempt to support Windows.

khaes-kth:
> have you considered Windows Subsystem for Linux (2)?

Not really! I will give it a try. Thanks for the suggestion.

slarse:
Just for the record, we've decided to add Windows support in the coming months. It's just not quite at the top of the priority list at the moment :)

slarse:
It would probably be helpful to fix #166 first, as using filepaths as strings is a massive complication for supporting different file systems.

slarse:
We have partial Windows support as of #591, but there are still a few test failures. These are ignored with `DisabledOnOs(OS.WINDOWS)` markers followed by a FIXME comment. We'll want to fix these before stating that we have full Windows support.

PR #590, opened 2021-08-12
doc: Add badge and text on platform support

slarse:
As noted in #273, we currently don't support Windows, but there's really no mention of that anywhere other than the issue tracker. In the past year, two Windows users have reported problems, see #335 and #589.

This PR adds a badge at the top of the README with the supported platforms, as well as some text in the `Prerequisites` section.


algomaster99 (approved):

slarse:
Ping @fermadeiral @monperrus @khaes-kth for merge

PR #593, opened 2021-08-16
fix: Fix linking of `handled_rules.py`

algomaster99:
A small typographical error in the link to `handled_rules.py` has been fixed in this PR. Unrelated to this change, many whitespaces have also been removed even though I didn't edit that particular part of the template.


slarse (approved):
LGTM

Issue #602, opened 2021-08-30
Import wrongly removed


khaes-kth:
When 2184 fixes are applied on [this](https://github.com/CloudburstMC/Nukkit.git) (commit: 47d54ad26109b5d89936e0d9123c591a99a2cf60), the following change is made:

```
@@ -23,9 +23,6 @@
  */
 package co.aikar.timings;

-import static co.aikar.timings.TimingIdentifier.DEFAULT_GROUP;
-import static co.aikar.timings.TimingsManager.*;
-
 public class FullServerTickTiming extends Timing {
     private static final TimingIdentifier IDENTIFIER = new TimingIdentifier(DEFAULT_GROUP.name, "Full Server Tick", null);
     final TimingData minuteData;


```

algomaster99:
@khaes-kth I can have a look at this issue.

algomaster99:
@khaes-kth I inspected this issue and I think it is because of the pretty printer options we use. For example, check this line.

https://github.com/SpoonLabs/sorald/blob/673137680e5d7aae663230d627aaf18316e3cf68/src/main/java/sorald/Repair.java#L295

It adds an option for cleaning imports which can either add or remove the import statements. Do we need this functionality in sorald? Sorald is supposed to sniper print the file with the repairs and it shouldn't concern the import statements if it is not explicitly causing a violation (we will know this because static analyzer will tell us). Please correct me if I am wrong.

I tried removing this option, however, many other tests fail, but I will take a look at it later.

khaes-kth:
> Sorald is supposed to sniper print the file with the repairs and it shouldn't concern the import statements if it is not explicitly causing a violation (we will know this because static analyzer will tell us).

I agree. However, it is possible that a fix requires new imports. Isn't it?

algomaster99:
> However, it is possible that a fix requires new imports.

That's right but we should let the processors deal with it not the pretty printer.

khaes-kth:
> That's right but we should let the processors deal with it not the sniper printer.

Yeah. That's an option to be considered. I'm not sure that it is clearly preferable. Actually, if we have a proper "import-adder" in Spoon that does this automatically, it makes sense to leave this to the pretty-printer instead of having specific import-adders for processors.

slarse:
[...]

PR #611, opened 2021-11-05
refactor: store excluded references in a set which wraps `IdentityHashMap`

algomaster99:
Reference #602 

I thought I would refactor the code first as suggested by @slarse. 

> Just glancing at that code I would begin with wrapping that map in a set (with Collections.newSetFromMap), because the values aren't used :)

### DIFF
--- a/src/main/java/sorald/SelectiveForceImport.java
+++ b/src/main/java/sorald/SelectiveForceImport.java
@@ -1,7 +1,9 @@
 package sorald;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.IdentityHashMap;
+import java.util.Set;
 import spoon.reflect.path.CtRole;
 import spoon.reflect.reference.CtTypeReference;
 import spoon.reflect.visitor.ForceImportProcessor;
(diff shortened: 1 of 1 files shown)


algomaster99:
I am not sure what went wrong here, because tests passed on my machine. @khaes-kth could you please re-run the workflow? I think the failure could be because of the flaky tests.

slarse:
@algomaster99 The scheduled builds on master have been failing for a few days, it's probably not related to this PR.

khaes-kth:
>  @khaes-kth could you please re-run the workflow? I think the failure could be because of the flaky tests.

I did it, but it seems the problem isn't fix with a rerun. 

@fermadeiral @monperrus : can you add @algomaster99  as a collaborator so that he can see the logs? It seems I do not have access to do it.

monperrus:
Aman and you now have role "maintainer".

algomaster99:
> I did it, but it seems the problem isn't fix with a rerun.

The scheduled builds are failing as pointed out by Simon. I have created [a PR](https://github.com/INRIA/spoon/pull/4272) to fix it, but it seems like that isn't the best way.

algomaster99:
The tests should pass here after #617 is merged.

algomaster99:
@khaes-kth once all tests are green, you may proceed to merge. This PR just refactors the code and it doesn't fix anything.

algomaster99:
@khaes-kth ping for merge.

khaes-kth (approved):
LGTM!

PR #616, opened 2021-11-12
DONTMERGE: Debug Windows path

slarse:


algomaster99:
Thank you for a step-by-step debugging session. This documents it quite well. I was trying to run same steps on my local machine itself because the behaviour of the subprocess was similar there too.

algomaster99:
Debugged and results are incorporated [here](https://github.com/INRIA/spoon/commit/5842bfda67a855b6a8191c4d1616fcd2dbe9527e).

PR #670, opened 2021-12-02
docs: update tests files and description for S2111

algomaster99:
The processor for `S2111` already handled the case when `new BigDecimal` has more than two arguments, but it was not documented in the processor test file. The description of S2111 was also slightly misleading. Starting the description with "Any constructor ..." and then talking about a case where it has more than one argument defeats the purpose of generalising in the first statement.

P.S. This is the second time we will test the script `handled_rules.py`. I am really looking forward to it :)


fermadeiral:
LGTM. Thank you, @algomaster99!

PR #675, opened 2021-12-06
chore: remove `picocli-codegen`

algomaster99:
I am unsure why this dependency is present in `pom.xml`. Without it, all tests are green so I thought to remove it. This was first pointed out by running `depclean` analysis.

Reference: https://github.com/castor-software/depclean/issues/108#issuecomment-983593761.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,12 +48,6 @@
             <artifactId>picocli</artifactId>
             <version>${picocli.version}</version>
         </dependency>
-        <dependency>
-            <groupId>info.picocli</groupId>
-            <artifactId>picocli-codegen</artifactId>
-            <version>${picocli.version}</version>
-            <scope>provided</scope>
-        </dependency>
         <!-- https://mvnrepository.com/artifact/org.json/json -->
         <dependency>
             <groupId>org.json</groupId>


khaes-kth:
I don't remember exactly, but as far as I can recall, we did generate some code somewhere! Maybe this `picocli-codegen` is required for that. I think @slarse knows better.

slarse:
`picocli-codegen` provides a bunch of compile-time checks for `picocli`, which otherwise appear at runtime. For example, try going here:

https://github.com/SpoonLabs/sorald/blob/f71357444c19232f6c9240faabbc44bccf1af948/src/main/java/sorald/cli/Cli.java#L18

And change `RepairCommand.class` to something nonsensical like `String.class`. Then run `mvn clean compile`, and the processor will get angry:

```
Subcommand is missing @Command annotation with a name attribute
```

However, if you now remove `picocli-codegen` from `pom.xml`, and THEN run `mvn clean compile`, it will compile just fine. There's nothing in Java's type system that can do the kinds of checks required here.

It can do a bunch of other stuff too, but that's why I included it.

https://picocli.info/#_annotation_processor

slarse:
To be clear, of course the project will compile without `picocli-codegen` if it compiles with `picocli-codegen`, but the reverse is not true.

khaes-kth:
Considering what Simon said, I guess it's better to keep it. WDYT @algomaster99 ?

algomaster99:
@khaes-kth I agree with @slarse . Let's keep this dependency now that we know its use.

Issue #603, opened 2021-08-30
Variable definition distorted


khaes-kth:
When 2142 fixes are applied on [this](https://github.com/salesforce/mirus.git) (commit: 9eab3f69a4a9125b9bf9c9fd77949903f94a67f9), the following change is made:

```
-  private final SourcePartitionValidator.MatchingStrategy validationStrategy;
+  private finalcom.salesforce.mirus.SourcePartitionValidator.MatchingStrategy validationStrategy;
```

khaes-kth:
A similar issue when 2272 fixes applied on [this](https://github.com/movsim/movsim.git) (commit: 24d4f4b83a7d2e14f16419fbf7ce05cb019c8066)
```
@@ -106,7 +112,7 @@ public class RoadSegment extends DefaultWeightedEdge implements Iterable<Vehicle
 
     private final int laneCount;
 
-    private final LaneSegment laneSegments[];
+    private final LaneSegment laneSegments;
 
     // TODO extend Node idea to keep information of connecting roadSegments
     private int sizeSourceRoadSegments = -1;
```

algomaster99:
@khaes-kth This may be fixed by https://github.com/INRIA/spoon/pull/4279.

khaes-kth:
Great! Is it possible to add a test in Sorald that checks if it is fixed or not?

algomaster99:
Yes, that is possible. Let the linked PR get merged and then I will submit a PR with the test for it. Meanwhile, I can test with my local build of spoon.

khaes-kth:
Awesome. Looking forward to it.

algomaster99:
Apparently, this is not so straightforward :disappointed:. This won't be fixed by https://github.com/INRIA/spoon/pull/4279.

The fragment `SourcePartitionValidator.MatchingStrategy|2817, 2858|SourcePartitionValidator.MatchingStrategy|` is an instance of `TokenSourceFragment` instead of `ElementSourceFragment`. Another anomaly is that its role is `IDENTIFIER` instead of `TYPE` which is wrong. `com.salesforce.mirus.SourcePartitionValidator.MatchingStrategy` is clearly the `type of the field and not an identifier.

[...]

Issue #600, opened 2021-08-30
Duplicate variable definition


khaes-kth:
When rule 2142 fixes are applied on [this repo](https://github.com/mitre/HTTP-Proxy-Servlet.git) (commit: 34edf588a), a variable definition is duplicated:

```
     HttpGet url = new HttpGet(String.format("http://localhost:%d/chatProxied/test", serverPort));
 
     try (CloseableHttpClient chc = HttpClientBuilder.create().build();
-            CloseableHttpResponse chr = chc.execute(url)) {
+            CloseableHttpResponse chr = chc.execute(url);CloseableHttpResponse chr = chc.execute(url)) {
       try (InputStream is = chr.getEntity().getContent()) {
         byte[] readData = readBlock(is);
         assertTrue("No data received (message1)", readData.length > 0);
```

monperrus:
Will be closed when https://github.com/INRIA/spoon/pull/4309 is merged

monperrus:
https://github.com/INRIA/spoon/pull/4309 merged

PR #676, opened 2021-12-09
chore: upgrade Spoon to `10.0.1-beta-2`

algomaster99:
We are upgrading the version of Spoon to this beta release because it has fixes to our high priority bugs and a workaround. I will push tests for all of them in this PR becasue they all are contingent upon the version upgrade. Moreover, it won't make the PR huge and difficult to review.

### Bugs

| Issue  | Spoon PR  | Commit in this PR  |
|---|---|---|
| https://github.com/SpoonLabs/sorald/issues/603  | https://github.com/INRIA/spoon/pull/4296  | https://github.com/SpoonLabs/sorald/pull/676/commits/e326cee8c72eea2899b8606862c8eb5074ba6adf  |
| https://github.com/SpoonLabs/sorald/issues/600  | https://github.com/INRIA/spoon/pull/4309  | https://github.com/SpoonLabs/sorald/pull/676/commits/9992719c1ea4aca6105c83d99f0c7fd369bb9994  |
| https://github.com/SpoonLabs/sorald/issues/602  | https://github.com/INRIA/spoon/pull/4331  | https://github.com/SpoonLabs/sorald/pull/676/commits/76c1e5b516f9184a07417d130cf070bf5e5de7b2  |

### Workaround

| Introduced In | Spoon PR  | Commit in this PR  |
|---|---|---|
| https://github.com/SpoonLabs/sorald/pull/617 | https://github.com/INRIA/spoon/pull/4298  | https://github.com/SpoonLabs/sorald/commit/f3b0280cccb39e744acc2315a2e261e2e39b0117  |


pom.xml
```diff
@@ -355,18 +355,6 @@
...
-            <id>spoonSnapshot</id>
-            <repositories>
-                <repository>
-                    <id>ow2.org-snapshot</id>
-                    <name>Maven Repository for Spoon Snapshots</name>
-                    <url>https://repository.ow2.org/nexus/content/repositories/snapshots/</url>
-                    <snapshots><enabled>true</enabled></snapshots>
-                </repository>
-            </repositories>
-        </profile>
```
algomaster99, line 369: I removed this XML config because it was not working for fetching SNAPSHOTs of Spoon when I was trying to debug Sorald using it. For using SNAPSHOT version, we can refer how it is done in [`diffmin`](https://github.com/SpoonLabs/diffmin/pull/90/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L55-L61) because I know it works there.

algomaster99:
@slarse I needed help with writing test cases for issues listed in the table above. All of those issues were related to the sniper printer, but they have been fixed in the `beta-2` release of Spoon. We are thinking of also adding a test case here in sorald which would check the printing behaviour. For example, consider #600. I tried making a test resource and the `.exact` file corresponding to it. However, it threw a lot of compilation errors in the test suite which I think stems from Sorald's test suite trying to compile its output after repairing. 

Is there a way in sorald to specify that the output shouldn't be compiled, but it should also check for the exact match? I found a keyword `NOCOMPILE_` in some test resources so I changed `RepairAndTryWithResource.java` to `NOCOMPILE_RepairAndTryWithResource.java.exact` in https://github.com/SpoonLabs/sorald/pull/676/commits/38e17db39d5f92cb028656a28081b0726a55d6a6, but I did not see a test in the runner which checked for exact file (I purposely put a wrong substring to see if the tests fail, but none did.).

algomaster99:
@khaes-kth Let's wait for tests to turn green and then I think we can merge this.

pom.xml
slarse: You don't want to do it like it is in diffmin, because that adds a repository to anything that depends on diffmin. In general, one should avoid adding repositories in pom files, [here's a pretty good article about it](https://blog.sonatype.com/2009/02/why-putting-repositories-in-your-poms-is-a-bad-idea/).

Putting the repository in a profile means that it won't be included by default in dependent libraries/applications, which is desirable. As a concrete example of why it's bad to put repositories in pom files (without putting them in custom profiles): the old snapshot repository we used for Spoon was in its pom file, and it was pretty unreliable and went down every now and then. And every time it did, it wasn't possible to build Spoon _or anything depending on Spoon_, as that repository couldn't be reached. Even though nothing was fetched from it.

This works perfectly if you activate `spoonSnapshot` profile, e.g. `mvn -P spoonSnapshot compile`, but by default the repository isn't included (which is what we want).

algomaster99: I did not know this. Thanks!

I was about to argue that it is still safe to enclose SNAPSHOT URLs in `repositories` however, sorald is also used as a dependency in some other projects, eg. repairnator. So I get your point. I will revert this change whenever we need the SNAPSHOT version next time.

khaes-kth: > I will revert this change whenever we need the SNAPSHOT version next time.

@algomaster99 Do you think we should open an issue not to forget this? Or should we do it right now?

algomaster99: I don't think so opening an issue is necessary because it is not needed and whenever we want to work with the SNAPSHOT version again, we will know what we have to do. :)

khaes-kth: Fine :)

PR #680, opened 2021-12-14
fix(SynchronizationOnStringOrBoxedProcessor): enables Sorald to generate multiple similar fields in separate classes for synchronization

khaes-kth:
Sorald does not generate multiple variables with the same hashcode in different classes, as it considers the hashcode to be the unique ID that represents a variable. As an example, see [this](https://github.com/khaes-kth/sorald/blob/fix/SyncronizationOnSimilarVariables/src/test/resources/processor_test_files/S1860_SynchronizationOnStringOrBoxed/SynchronizationOnSimilarVariables.java#L14).

This PR fixes this by adding the variable's qualified name to its unique ID.

### DIFF
--- a/src/main/java/sorald/processor/SynchronizationOnStringOrBoxedProcessor.java
+++ b/src/main/java/sorald/processor/SynchronizationOnStringOrBoxedProcessor.java
@@ -85,11 +85,11 @@ public class SynchronizationOnStringOrBoxedProcessor
 
             c.addFieldAtTop(newField);
             old2NewFields.put(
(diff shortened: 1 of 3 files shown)


algomaster99 (changes requested):
I think `getQualifiedName` would be sufficient to make it unique.

Moreover, should we also change the type argument of `old2NewMethods` as we did for `old2NewFields`? We can probably introduce a failing test case for that in a separate PR.

algomaster99:
@khaes-kth could you run spotless?

PR #683, opened 2021-12-16
docs: update usage of `repair` command

algomaster99:
We should really have a python script to generate such things. @slarse you opened an issue for that, as far as I remember. Could you link it?


khaes-kth (approved):
LGTM. Thanks Aman.

slarse:
@algomaster99 I don't think I did. If I did it's one of the 46 open issues, just scroll through them :)

algomaster99:
You might have said it during one of our meetings and maybe that's why I was recalling it, otherwise, I can't find it in the open issues.

Issue #659, opened 2021-11-29
Upgrade sonar-java to 7.15.0.30507


algomaster99:
~- [x] org.sonarsource.java:java-checks-testkit~
~- [x] org.sonarsource.java:java-frontend~
~- [x] org.sonarsource.java:java-checks~

~Upgrading `org.sonarsource.java:java-frontend` induce a lot of breaking behaviour in the code. It is better to create an issue and deal with it.~

Now we do not need to upgrade each package independently. We can simply upgrade the JAR. However, it would break S2755 so that still needs to be investigated.

algomaster99:
We need to find proper release notes before jumping into this issue. They have split a lot of APIs into different packages. For example, we would also need `org.sonarsource.sonarqube:sonar-plugin-api-impl`.

algomaster99:
@khaes-kth @monperrus I think we can focus on this issue before adding any further rules.

monperrus:
yes, 100% agree.

monperrus:
>  All later versions of the plugin break S2755.

For this, since we want to depend on the latest sonar-java, two options:

* we drop the rule
* we re-write the detector ourselves (we have Spoon right ;-)

I'd be in favor of the latter

algomaster99:
@monperrus I will report this to SonarJava and then we can surely confirm if the issue is at our end or theirs. Then we can take the action.

monperrus:
great

algomaster99:
I forgot to update it here. I checked this myself and found out that the source position is correctly detected when the static analysis is executed by SonarQube itself.

algomaster99:
New goal next week - resolve this issue. Work already started at #900. 5 test cases are failing. It should not be hard to resolve them unless it breaks one of our processors.

PR #696, opened 2022-01-24
refactor: Use SonarLint core in Sorald

algomaster99:
Continuation of https://github.com/SpoonLabs/sorald/pull/583.

These changes create a Sonar Lint engine and then run Sonar Java as its plugin. The changes are inspired by the #583 (thanks @slarse ) and https://github.com/SonarSource/sonarlint-core/ implementations.

Like @slarse said, the test suites are WAY slower. To give context, each test in `ProcessorTest` takes around 2 seconds compared to 100ms previously. It is probably because we initialize Sonar Lint engine **and** analysis configuration every time we start mine or repair. I think we can bring down the time if we can somehow skip initialising Sonar Lint engine because that remains the same for all our executions.

- [x] Bring down test time
- [x] Refactor SonarLint engine (there are a lot of methods we do not need)


algomaster99:
We can approach the problem in two ways:

1. Initialise the SonarLintEngine before all tests of a test suite and stop it after all the tests have been executed. This is how [SonarSource runs their test](https://github.com/SonarSource/sonarlint-core/blob/0850afd36c001fe91c6f2810e08163394e672f1e/its/tests/src/test/java/its/StandaloneTest.java#L64-L80).
2. Cache the SonarLintEngine in JVM and then whenever we need to analyse something, we attach _a new instance of AnaylsisEngine_ because the previous one would have stopped. I have a feeling it would save us time since we do not need to initialise `analysisGlobalConfig` and `pluginInstancesRepository` repeatedly.

Approach `1` requires us to refactor the entire test cases because we need to build the engine within the test suite and pass it to `projectScanner`. The latter would entail modifying a lot of source methods. Moreover, it may make the test suite dependent upon SonarJava which we do not want in foresight as we may integrate other static analysers in future.

I am learning towards approach `2` for now. The idea seems simple. I am looking at how I can cache an object.

monperrus:
great, what's the total execution time of the test suite with this?

slarse:
@algomaster99 You can cache the sonarlint engine to  make consecutive test runs execute faster.

algomaster99:
> great, what's the total execution time of the test suite with this?

@monperrus it's around 30 minutes for Sorald's `build-and-test` workflow. You can have a look at it [here](https://github.com/SpoonLabs/sorald/runs/4920967158?check_suite_focus=true). On the bright side, it passed.

@slarse 
How do I cache it? Do you have any suggestions for it? I tried to serialise sonar lint object into a file, however, that did not work because some fields of the `SonarLint` engine were throwing `NotSerializableException`.

monperrus:
30 minutes is not a blocker IMHO we can proceed with this design

slarse:
> > great, what's the total execution time of the test suite with this?
> 
> @monperrus it's around 30 minutes for Sorald's `build-and-test` workflow. You can have a look at it [here](https://github.com/SpoonLabs/sorald/runs/4920967158?check_suite_focus=true). On the bright side, it passed.
> 
> @slarse How do I cache it? Do you have any suggestions for it? I tried to serialise sonar lint object into a file, however, that did not work because some fields of the `SonarLint` engine were throwing `NotSerializableException`.

Serialization is useful when you need to "load" object from one execution of a JVM to another, and complex objects often cannot be serialized. As long as you're within a single JVM process, which is the case for the unit tests, you can most easily just cache in a static field.

Note that there are some caveats to making such caching thread safe. For thread safety, it's best to use thread-local storage (built-in to Java as `ThreadLocal`). I don't know if sonarlint is thread safe to begin with, though, and it's definitely a little bit of "programming for the future". You decide what to do with that knowledge :)

algomaster99:
> 30 minutes is not a blocker IMHO we can proceed with this design

@monperrus  It does slow down the development process. I think it is worth the effort to bring it down. :)

> As long as you're within a single JVM process, which is the case for the unit tests, you can most easily just cache in a static field. That is to say, serialization is not what you want to use here.

I did try to initialise `sonarLintEngine` in a static field. Its implementation looked something like this:

```diff
- private final SonarLintEngine sonarLint = getOrCreateEngine();
+ private static final SonarLintEngine sonarLint = getOrCreateEngine();
```
Depending upon if it was `null`, I created the `sonarLint` engine again. I shall document soon why it was not working. I am trying out a few things.

src/main/java/sorald/sonar/SonarStaticAnalyzer.java
```diff
@@ -20,21 +20,23 @@
...
+                    Paths.get("target/classes").resolve("sonar-java-plugin-6.12.0.24852.jar");
+            StandaloneGlobalConfiguration globalConfig =
+                    StandaloneGlobalConfiguration.builder()
+                            .addPlugin(sonarJavaPath)
+                            .addEnabledLanguage(Language.JAVA)
+                            .build();
+            sonarLint = new SonarLintEngine(globalConfig);
+        } else {
+            sonarLint.recreateAnalysisEngine();
+        }
```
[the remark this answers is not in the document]
slarse: I don't remember exactly what component I cached, but I do remember that I did get the unit tests to run in something like 5-6 minutes.

algomaster99:
@slarse @monperrus I finally found the exact component that took a lot of time to start up! In the latest commit, I prevent it from closing. Now it takes 5 minutes to run the `build-and-test` workflow. :partying_face: 

This problem was eating me up since Friday and now I can finally sleep :)

monperrus:
excellent!

algomaster99:
@monperrus @slarse @khaes-kth 

This PR is ready to review.

### Summary of the changes
1. Implement `SonarLintEngine`.
1. Implement `SonarStaticAnalyzer` which runs as a plugin of the above engine. Note that we use `sonar-java-plugin` version `6.12.0.24852` and not the latest. All later versions of the plugin break [S2755](https://rules.sonarsource.com/java/RSPEC-2755). This has been noted by @slarse [here](https://github.com/SpoonLabs/sorald/issues/331#issue-786806685) and I also checked for the latest versions.
1. Delete tests `ScannedViolationTest.java` and `RuleVerifierTest.java` as they were tightly coupled with Sonar Java internals.
2. Refactor `RepairTest` to pass the rules to the scanner at once.
3. One of the rules, S4829, has been deprecated so I have removed it from the expected violations.
4. Fix resolution of sonar java plugin version in python tests.

There are minor things left, however, it can be reviewed at this stage.
- [ ] Improve coverage
- [ ] Fix Sonar violations. (If we don't fix it, how can we expect others to do so? :laughing: )

### Some future work
1. Look why `Checks` is needed. We compute all SonarRuleDefinition during runtime in `SonarLintEngine` - `allRulesDefinitionsByKey`.
2. Deliberate if we want to upgrade `sonar-java-plugin`'s version given the above regression.

monperrus:
>  All later versions of the plugin break S2755.

See my answer in https://github.com/SpoonLabs/sorald/issues/659

monperrus:
LGTM! Thanks a lot.

algomaster99:
Wait for merge. I shall also write test cases where [`textRange` is null](https://github.com/SpoonLabs/sorald/pull/696/files#diff-f0cbed36f32aab9e1fd108af2800b9dc14ab7f0fc9f34fcf112616bb47507923R14-R20). One of the rules where it is null - https://rules.sonarsource.com/java/RSPEC-1996.

khaes-kth:
> Wait for merge. I shall also write test cases where [`textRange` is null](https://github.com/SpoonLabs/sorald/pull/696/files#diff-f0cbed36f32aab9e1fd108af2800b9dc14ab7f0fc9f34fcf112616bb47507923R14-R20). One of the rules where it is null - https://rules.sonarsource.com/java/RSPEC-1996.

Thanks Aman. Ping me when you are done with the new tests.

algomaster99:
@khaes-kth 

I stand corrected. The sonar-java plugin was not loaded when we executed the Sorald as standalone JAR. When we executed from the Sorald root (or the CI), it used to resolve the path from the `target` folder. I have added a workaround [here](https://github.com/SpoonLabs/sorald/pull/696/commits/7988bbe5a1558628200440039b514a0e0b49414e). We can have a look at it today.

We can also discuss the exclusion of some rules made in https://github.com/SpoonLabs/sorald/pull/696/commits/b24bd480f2977ec3fdfb73cf46fad2e1a14c60d3. These rules do not have `textRange` attached to them.

khaes-kth:
> When we executed from the Sorald root (or the CI), it used to resolve the path from the target folder.

Aha! That's what also happened last time we were testing it. Let's discuss it in person :+1:

algomaster99:
@khaes-kth This is ready to review. Sonar still complains about how we created the temporary directory.

khaes-kth:
> Sonar still complains about how we created the temporary directory.

Having a look at Sonar documents, I think something like `Files.createTempFile("prefix", "suffix").toFile()` may work. Note that it is using `java.nio.file.Files.createTempFile`, not `java.io.File.createTempFile` and it is providing both `prefix` and `suffix`. Can you give it a try?

slarse:
I would advice against bundling the sonar-java JAR-file with Sorald, for several reasons.

1. It bloats the size of the Git repository. Due to how Git stores file revisions, each time you update to a new version of the sonar-java plugin, the repository will roughly get as much larger as that binary blob is large.
    - This can technically be mitigated by downloading the plugin JAR when building the project, but see the next two points
2. It bloats the size of your own redistributable JAR.
3. It strongly ties Sorald to a specific version of sonar-java, when in fact Sorald will probably work with many different versions (e.g. all 7.x versions).
4. Redistributing the original JAR like this is troubling from a license perspective as Sorald is not LGPL-licensed. Arguably, Sorald is already in murky territory when it comes to licensing.

For the purposes of just getting this underway, I would personally implement it such that Sorald looks for the sonar-java plugin based on an environment variable (e.g. `SORALD_SONAR_JAVA_PATH`), and defaults to looking in the current working directory if that environment variable isn't defined. You can make this more sophisticated down the line, for example by downloading an "optimal" JAR on demand if it can't be found locally.

But yeah, I really wouldn't put the sonar-java plugin JAR inside of Sorald's Git repository, nor bundle it with Sorald. I think it's an all-round bad idea.

algomaster99:
@khaes-kth 

I think we cannot ignore the violations which have `null` `textRange` [because all the violations are put into a `HashSet`](https://github.com/SpoonLabs/sorald/blob/b5039d5b7cf1fcdbf526da4238de941eef56c23a/src/main/java/sorald/sonar/ProjectScanner.java#L68) which would internally call [`hashCode`](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/rule/RuleViolation.java#L79) resulting in a `NullPointerException`.

@slarse I shall revert to your comment tomorrow. :)

algomaster99:
@slarse I agree with your points. Our goal was to somehow provide a standalone JAR to be able to run `repair` or mine. If we don't package the Sonar JAR with Sorald JAR, we would be expecting our users to download it themselves for now. Are we okay with this, @monperrus @khaes-kth ? The only better solution would have been that the SonarLintEngine accepted a maven dependency instead of a path to the JAR.

khaes-kth:
> @slarse I agree with your points. Our goal was to somehow provide a standalone JAR to be able to run `repair` or mine. If we don't package the Sonar JAR with Sorald JAR, we would be expecting our users to download it themselves for now. Are we okay with this, @monperrus @khaes-kth ? The only better solution would have been that the SonarLintEngine accepted a maven dependency instead of a path to the JAR.

I guess @slarse is right. We may need to sacrifice some execution simplicity to all the negative points @slarse mentioned.

khaes-kth:
> I think we cannot ignore the violations which have `null` `textRange` [because all the violations are put into a `HashSet`](https://github.com/SpoonLabs/sorald/blob/b5039d5b7cf1fcdbf526da4238de941eef56c23a/src/main/java/sorald/sonar/ProjectScanner.java#L68) which would internally call [`hashCode`](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/rule/RuleViolation.java#L79) resulting in a `NullPointerException`.

Per discussion, we need to know what was exactly happening before this PR, then we'll make a decision.

slarse:
> Our goal was to somehow provide a standalone JAR to be able to run repair or mine.

A workaround to this would be to simply download the sonar-java plugin on-demand. Check if it's locally available, preferably in a cache directory, e.g. `~/.cache/sorald/...` or equivalent (there's probably  a library to find the local OS cache directory, like the `appdirs` package for Python). When you run Sorald, you first check if sonar-java is available in the working directory or the designated cache, and otherwise download it.

algomaster99:
@slarse @khaes-kth The tests fail because we are not able to create a cache directory - `/home/user/.cache/sorald`. I explicitly [wrote a test](https://github.com/SpoonLabs/sorald/pull/696/files#diff-68a24b6b42d240d14d9c7daa493417dccb9bebfe1235045d352fd09ceb1752fcR86) for it and it is failing. I think this might have to something with file permissions in actions.

khaes-kth:
> I think this might have to do something with file permissions in actions

That makes sense.

algomaster99:
@khaes-kth 
> >I think we cannot ignore the violations which have null textRange because all the violations are put into a HashSet which would internally call hashCode resulting in a NullPointerException.

> Per discussion, we need to know what was exactly happening before this PR, then we'll make a decision.

I checked this. We were ignoring violations without position before as well.

src/main/java/sorald/sonar/SonarLintEngine.java
```diff
@@ -0,0 +1,232 @@
...
+                .setModulesProvider(globalConfig.getModulesProvider())
+                .build();
+    }
+
+    /** Get or creates the one and only instance of this class. */
+    public static SonarLintEngine getInstance() {
+        if (theOnlyInstance == null) {
+            theOnlyInstance = new SonarLintEngine();
+        } else {
+            theOnlyInstance.recreateAnalysisEngine();
```
khaes-kth, line 133: Per our discussion, we do not need to `recreateAnalysisEngine()` every time we call `getInstance()`.

src/main/java/sorald/sonar/SonarLintEngine.java
```diff
@@ -0,0 +1,232 @@
...
+                new AnalysisEngine(analysisGlobalConfig, pluginInstancesRepository, null);
+    }
+
+    private static Path getOrDownloadSonarJavaPlugin() {
+        File cacheDirectory = FileUtils.getCacheDir();
+        String sonarJavaPluginFileName =
+                cacheDirectory + System.getProperty("file.separator") + SONAR_JAVA_PLUGIN_JAR_NAME;
+        File sonarJavaPlugin = new File(sonarJavaPluginFileName);
+        if (sonarJavaPlugin.exists()) {
+            return sonarJavaPlugin.toPath();
```
khaes-kth, line 82: We may need to use reflection to call this method twice and make sure it does not download the Jar the second time.

khaes-kth:
LGTM! I'll merge this PR.

monperrus:
strong milestone!

slarse:
Congrats @algomaster99, very nice work, thanks for concluding what I started :)

PR #703, opened 2022-01-26
fix: repair of S2142 should not induce S1193 violation

### LINKED ISSUE #697: Repair of S2142 can induce S1193
For the following snippet,
```java
catch (InterruptedException | IOException e) {
}
```
Sorald proposes the repair as below.
```java
catch (InterruptedException | IOException e) { 
    if (e instanceof InterruptedException) {
        Thread.currentThread().interrupt();
    }
}
```
This is a violation of [S1193: Exception types should not be tested using "instanceof" in catch blocks](https://rules.sonarsource.com/java/RSPEC-1193).

References
1. https://github.com/SpoonLabs/sorald/pull/696/commits/7a85a1c1adc5ed80145714e5e2bbb4ed097cef47
2. https://sonarcloud.io/project/issues?id=SpoonLabs_sorald&open=AX6SYB_whjp6fLbLYBVN&pullRequest=696&resolved=false&types=CODE_SMELL

algomaster99:
Fixes #697 

Before this gets merged, there is one thing we need to discuss. Consider the following case.
```java
catch (IOException | InterruptedException e) {
     logger.error("Raising exception to connect runtime", e);
     context.raiseError(e);
}
```
When we create another catch block for `InterruptedException`, should we or should we not copy the statements inside the original one? In this case, do we also need to copy `logger.error` and `context.raiseError` inside the new catch block we create? In my opinion, we should copy them as those statements, would be related to both the exception classes.

ToDo:
- [x] Include statements in the newly created catch block
- [x] Modify the processor description

### DIFF
--- a/src/main/java/sorald/processor/InterruptedExceptionProcessor.java
+++ b/src/main/java/sorald/processor/InterruptedExceptionProcessor.java
@@ -1,21 +1,20 @@
 package sorald.processor;
 
 import sorald.annotations.ProcessorAnnotation;
-import spoon.reflect.code.BinaryOperatorKind;
 import spoon.reflect.code.CtBlock;
 import spoon.reflect.code.CtCatch;
-import spoon.reflect.code.CtIf;
+import spoon.reflect.code.CtCatchVariable;
 import spoon.reflect.code.CtInvocation;
 import spoon.reflect.code.CtLabelledFlowBreak;
 import spoon.reflect.code.CtReturn;
 import spoon.reflect.code.CtStatement;
 import spoon.reflect.code.CtThrow;
+import spoon.reflect.code.CtTry;
 import spoon.reflect.code.CtTypeAccess;
-import spoon.reflect.code.CtVariableAccess;
 import spoon.reflect.declaration.CtClass;
 import spoon.reflect.declaration.CtMethod;
-import spoon.reflect.declaration.CtVariable;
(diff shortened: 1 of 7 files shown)


src/main/java/sorald/processor/InterruptedExceptionProcessor.java
```diff
@@ -32,13 +31,42 @@ protected void repairInternal(CtCatch element) {
...
+
+        // Remove InterruptedException from the catch
+        CtCatchVariable<?> catchVariable = violatedCatch.getParameter();
+        catchVariable.removeMultiType(refToInterruptedException);
+
+        // Add statement into a cloned catch block
+        // Currently, we clone the body of the violated catch which is theoretically wrong because
+        // there could be statements concerning only one of the exception types.
+        // Example - `if (e instanceof IOException) { ... }`. However, it is safe // NOSONAR:S125
+        // to ignore this case because it is rare.
```
algomaster99, line 55: I have added the comment here we noted yesterday.

Sonar would not let me comment `if (e instanceof IOException) { ... }` inside a meaningful comment and that's why I added a comment on line 54.

algomaster99 (commented):
@khaes-kth this is ready for review. Have a look at my comments too.

khaes-kth (approved):
LGTM!

PR #704, opened 2022-01-27
review: fix CI on windows

### LINKED ISSUE #273: Add support for Windows

MartinWitt:
Classical first PR trying to fix the CI on Windows.  The problems seems that the filename is simply the absolute path and not the relative path. How on earth this works on Unix is unknown to me. 
Fixes #273

### DIFF
--- a/src/test/java/sorald/GatherStatsTest.java
+++ b/src/test/java/sorald/GatherStatsTest.java
@@ -12,8 +12,6 @@ import java.util.ArrayList;
 import java.util.Set;
 import org.json.JSONArray;
 import org.json.JSONObject;
-import org.junit.jupiter.api.condition.DisabledOnOs;
-import org.junit.jupiter.api.condition.OS;
 import org.junit.jupiter.api.io.TempDir;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
@@ -25,7 +23,6 @@ import sorald.rule.RuleViolation;
 import sorald.sonar.ProjectScanner;
 import sorald.sonar.SonarRule;
 
-@DisabledOnOs(OS.WINDOWS) // FIXME make this test class pass on Windows
 public class GatherStatsTest {
 
     @ParameterizedTest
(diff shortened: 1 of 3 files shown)


algomaster99:
We are replacing `RuleVerifier` with [`SonarLintEngine`](https://github.com/SpoonLabs/sorald/pull/696/files#diff-7bae8e494b540a76e6d87048c0089e595e1ae162fb8faf8751acab79224095d3) in #696 . Once that refactoring PR is done, I think we can work on fixing the CI on Windows.

Moreover, the general way to fix this is to use `Path` instead of `String`. See https://github.com/SpoonLabs/sorald/issues/166. I believe this way we shall be able to support Sorald independent of the file system.

algomaster99:
@MartinWitt now you may continue to work on this. We have refactored the infrastructure of the project. See #696.

MartinWitt:
Works 👍 
https://github.com/SpoonLabs/sorald/blob/c9db9ffac490b0d72ee13abe6ab1ae1794341fe2/src/main/java/sorald/FileUtils.java#L132 You may want to remove this last artifact of windows, the File API can handle Unix style paths on windows.

algomaster99:
> You may want to remove this last artefact of windows, the File API can handle Unix style paths on windows.

@MartinWitt It is just not handling path format, but also storing the jar plugin in the conventionally accepted cache directory of Windows which now I realise is not correct. It should be something like this - `C:\Users\HP-JVF78\AppData\Local\aman\sorald\Cache\`.

I guess we can merge this PR regardless your suggested change?

algomaster99:
I am also surprised that the tests pass as is. I wonder why it was disabled before.

MartinWitt:
> I guess we can merge this PR regardless of your suggested change?

Yes absolutely the path is a simple style issue. And as it seems there are more problems, we should split this in a separate PR.

> I am also surprised that the tests pass as is. I wonder why it was disabled before.

I was surprised too.

algomaster99 (approved):
Thanks @MartinWitt !

algomaster99:
@MartinWitt If we have official support for Windows now, should we also update the README badge?

MartinWitt:
Yes, the badge and the README text about platform support.

algomaster99:
Could you submit a PR for that?

MartinWitt:
Yes I will create one.

PR #705, opened 2022-01-27
chore: add factorypath to gitignore

MartinWitt:
This was missing inside the `.gitignore`. The gitignore seems kinda small, have you thought about simply adding something like this https://www.toptal.com/developers/gitignore/api/java,visualstudiocode,eclipse,intellij,maven,gradle ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ script.sh
 sorald-workspace/
 FileOutputStrategyTest/
 sonar-java/
+.factorypath


algomaster99:
Hi @MartinWitt !

Thanks for your PR. I assume `.factorypath` is for development on Eclipse?

> gitignore seems kinda small, have you thought about simply adding something like this https://www.toptal.com/developers/gitignore/api/java,visualstudiocode,eclipse,intellij,maven,gradle

I always add globs to `.gitignore` when I know I want to ignore them. For example, templates also ignore the jar files but in #696 we need the sonar-java.jar to execute as a sonar-lint plugin. Of course, this is one exceptional case, but I feel the generic `.gitignore`s are more verbose than they need to be.

MartinWitt:
> I assume .factorypath is for development on Eclipse?

From my understanding, the eclipse compiler generates these because vscode has no more resource files inside the project folder.  In short it appears during my development.

PR #715, opened 2022-02-02
chore: Remove dependence of SonarJava jar plugin

algomaster99:
We accidentally merged the jar of SonarJava plugin. It is not used in the code as-is and instead, we download it from maven so it does not need to be bundled with our package.

### DIFF
--- a/experimentation/tools/tests/test_prmessage.py
+++ b/experimentation/tools/tests/test_prmessage.py
@@ -1,12 +1,10 @@
 import shlex
 import re
-import pathlib
 
 import pytest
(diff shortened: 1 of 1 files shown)


algomaster99:
@khaes-kth Let's wait for the tests to turn gree for sanity.

algomaster99:
Python tests use the jar file. Let me fix the python tests as well.

algomaster99:
@khaes-kth It is ready for merge.

khaes-kth (approved):
LGTM!

PR #726, opened 2022-02-04
style: remove never updated field

MartinWitt:
The smell was found by qodana and way too hard to automatic fix it. But the field was only queried but never updated. Queries against an empty hash map are hopefully not needed.


algomaster99 (approved):
Looks good to me.

I think check was here because of presumed symmetry with `fields`. If two methods in a class use the same lock, we only need to construct one field for creating a legal lock. But it seems like, we don't need it for methods.

Issue #713, opened 2022-02-02
`sorald.mine` does not report all violations because of lack of source position


algomaster99:
Reference: https://github.com/SpoonLabs/sorald/pull/696#issuecomment-1026544291

Some violations mined by SonarJava were filtered out because they did not have any source position associated with them. See https://github.com/SpoonLabs/sorald/pull/315. In my opinion, we should still report them even though we know we can't repair them with the current infrastructure of the project. The best and the most foolproof way to list down all such rules would be to refer to [sonar-java](https://github.com/SonarSource/sonar-java/)'s source code. For example, I observed that instances of [S1996](https://rules.sonarsource.com/java/RSPEC-1996) did not report any position with them. I referred to its source code and found out that they [do not report position for _any instance_](https://github.com/SonarSource/sonar-java/blob/cbf1e883e99c0c54b499b8663543fc6449aa856e/java-checks/src/main/java/org/sonar/java/checks/OneClassInterfacePerFileCheck.java#L44).

Known unreported violations:
- [S1996](https://rules.sonarsource.com/java/RSPEC-1996)
- [S1228](https://rules.sonarsource.com/java/RSPEC-1228)
- [S4032](https://rules.sonarsource.com/java/RSPEC-4032)

algomaster99:
It is very tough to determine this information statically, so I have asked this question on [their forum](https://community.sonarsource.com/t/list-all-rule-keys-whose-warnings-have-no-line-number-associated-with-it/58721). Last resort is to execute the code and figure this out from the dynamic information we get.

algomaster99:
I found the disparity in the rules we report and the rules that exist (SonarJava `v6.12.0.24852`). I have updated them in a gist: https://gist.github.com/algomaster99/755539f4e1ade1743d8a5b2c1bf4e238.

algomaster99:
[@NiklasWicklund](https://github.com/NiklasWicklund) reported some violations that are not detected by `sorald.mine`.
- [S1176](https://rules.sonarsource.com/java/RSPEC-1176)
- [S100](https://rules.sonarsource.com/java/RSPEC-100)
- [S101](https://rules.sonarsource.com/java/RSPEC-101)

I am pretty sure each of these rules would have a source position associated with them (based on my intuition) so I do not see why Sorald would not support them.

algomaster99:
Fixed in https://github.com/SpoonLabs/sorald/pull/751.

PR #746, opened 2022-03-11
feat: add checks which were assumed to be deprecated

algomaster99:
There were [some rules](https://github.com/SpoonLabs/sorald/issues/713#issuecomment-1064957430) that were not reported by Sorald because their `Check` was not an entry in `sorald.sonar.Checks`. This was done because we assumed that `DeprecatedRuleKey` would mean that the rule itself has been deprecated. However, it means that the _key_ corresponding to the rule has been deprecated. In other words, the key of a said rule has been updated.

The rules reported by @NiklasWicklund [here](https://github.com/SpoonLabs/sorald/issues/713#issuecomment-1064957430) were not being reported because we had disabled them.


algomaster99:
There are approximately 45 more rules which are not configured in `Checks.java`. Maybe we can add them all. @slarse @khaes-kth is there a script to generate [all the maps](https://github.com/SpoonLabs/sorald/blob/86363a1f745c69339c3cc3d7d0efbacb9f6ba37e/src/main/java/sorald/sonar/Checks.java#L131) defined in that file?

algomaster99:
I added tests to confirm that those violations are reported now.

khaes-kth:
Thanks for documenting taking care of this Aman. I am not aware of the script you are looking for.

I just wanted to know if this significantly affects our results in the paper. Can you assess how many new violations are detected after adding these checks? Currently, we report that "The total number of violations of all 97 fixable and partially fixable bug rules in our dataset is 4,928." This number is extracted from the mining `stats.json` file.

algomaster99:
@khaes-kth these rules are code smells. I don't think they should affect the rules. Moreover, if we do a manual analysis of code smell rules as well, these 3 rules would be unfixable.

algomaster99:
> I am not aware of the script you are looking for.

I am sorry I forgot to respond to this. I mean the maps [here](https://github.com/SpoonLabs/sorald/blob/86363a1f745c69339c3cc3d7d0efbacb9f6ba37e/src/main/java/sorald/sonar/Checks.java#L131-L672). I am pretty sure they are not maintained manually.

algomaster99:
@NiklasWicklund, you can now build the latest version on `master` to run your experiments.

PR #748, opened 2022-03-14
refactor: generalise app to accept any static analyzer

algomaster99:
My motivation behind this PR is to remove the [Checks](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/sonar/Checks.java) class because it is still dependent upon one internal class of `sonar-java`.

Upon contemplating on how to remove the class, I realised it needs a lot of refactoring prior to it and it eventually led up to generalising the whole app for a static analyzer. This PR is **not to be merged** anytime soon because I will make multiple pull requests to this branch adding each set of refactoring at a time.


MartinWitt:
Have you thought about serviceloading analyzers in the future(https://www.baeldung.com/java-spi)? This would decouple sorald from sonar completely and enables easier integration of new rules/analyzers. As you are already creating general interfaces, this could be possible.

algomaster99:
> Have you thought about serviceloading analyzers in the future(https://www.baeldung.com/java-spi)?

@MartinWitt Thank you for your suggestion! I did not know about this concept and it seems a solution for the problem I was facing. I was actually about to write a long post about the issue with generalising. The gist of that is in the following paragraph.

I was planning to add a flag `--static-analyzer` to sorald subcommands that would accept `SONAR` or other static analyzers. So first step was to define an enum like so: `public enum StaticAnalyzer { SONAR }` and then I use this enum to transform all classes in Sorald to accept generic type parameters.

However, the problem was that this does not make sense because `--static-analyzer` is passed during runtime so type erasure will remove all type parameters anyway.

Using Java SPI sounds like a good idea instead of using enums to store all the static analyzer we have integrated.

algomaster99:
Closing this PR to solely focus on removing Checks for now.

PR #750, opened 2022-03-15
refactor: remove dependence on `sonar-java-checks`

### LINKED ISSUE #722: `Checks.java` might no longer be needed
In #696, we integrated SonarJava as a plugin of [SonarLintEngine](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/sonar/SonarLintEngine.java#L57). The engine computes all the sonar rule definitions at runtime and stores in [here](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/sonar/SonarLintEngine.java#L57). We can use this collection itself when we want to `mine` instead of passing all the rule keys from `Checks`. For `repair`, we can validate two things:
1. We have a processor for it. I am not sure where it is done, but I am confident we do it.
2. The rule key is present in the collection mentioned above.

algomaster99:
Fixes #722 

We finally have started the journey to the end of the decoupling of Sorald and sonar java internal libraries! Now we don't need to manually provide the rule keys for mining or repairing. All the rules available for a SonarJava version will be computed at runtime.

There are some todos before we decide to merge this PR:
- [ ] Report all violations at runtime. 
- [ ] What should be the new rule name? Now we don't have short hand names like `UnusedPrivateFieldCheck` because those came from checks and we are not dependent upon them. For example the name `MagicNumberCheck` [here](https://github.com/SpoonLabs/sorald/blob/5ab1e14b4fcfb30cd91f30ad6069b33dc0cb39d2/src/test/resources/warning_miner/test_results.txt#L3) came from the check defined. 
- [x] Add tests in SonarLintEngine to ensure correct set of rules is fetched.
- [ ] Look if this does not alter the claims in the research paper.


algomaster99:
Continued it #751 .

PR #751, opened 2022-03-15
refactor: remove dependence on sonar-java-checks

### LINKED ISSUE #722: `Checks.java` might no longer be needed

algomaster99:
Unnecessarily spent so much time in #750, but fortunately, I found an easier way in this PR. The todos are still pending though.

## Duplicated from #750

Fixes #722 

We finally have started the journey to the end of the decoupling of Sorald and sonar java internal libraries! Now we don't need to manually provide the rule keys for mining or repairing. All the rules available for a SonarJava version will be computed at runtime.

There are some todos before we decide to merge this PR:
- [x] Report all violations at runtime. 
- [x] What should be the new rule name? Now we don't have short hand names like `UnusedPrivateFieldCheck` because those came from checks and we are not dependent upon them. For example, the name `MagicNumberCheck` [here](https://github.com/SpoonLabs/sorald/blob/5ab1e14b4fcfb30cd91f30ad6069b33dc0cb39d2/src/test/resources/warning_miner/test_results.txt#L3) came from the check defined [here](https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/test/files/checks/MagicNumberCheck.java).
- [x] Add tests to check if the correct subset of rules is "inferred".
- [x] Look if this does not alter the claims in the research paper.


algomaster99:
@khaes-kth two subtasks are remaining and in one of them, I need your help.

1. > What should be the new rule name? Now we don't have shorthand names like UnusedPrivateFieldCheck because those came from checks and we are not dependent upon them. For example, the name MagicNumberCheck [here](https://github.com/SpoonLabs/sorald/blob/5ab1e14b4fcfb30cd91f30ad6069b33dc0cb39d2/src/test/resources/warning_miner/test_results.txt#L3) came from the check defined [here](https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/test/files/checks/MagicNumberCheck.java).

	Any opinions on this? All the rules are now computed by SonarLintEngine [here](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/sonar/SonarLintEngine.java#L56). Each rule is an instance of [`SonarLintRuleDefinition`](https://github.com/SonarSource/sonarlint-core/blob/master/rule-extractor/src/main/java/org/sonarsource/sonarlint/core/rule/extractor/SonarLintRuleDefinition.java#L48) and it does not have anything like a _short description. It has a name attribute though, but I think it is too long. For example, for `S1068`, the name is "Unused "private" fields should be removed".

2. > Look if this does not alter the claims in the research paper.

	Based on intuition, I don' think it should alter any claims in the paper. We claim that there is a certain number of detected violations for the 10 rules there. It should not change because we have just added more rules and their violations would be ignored anyway.

algomaster99:
1. We are only keeping the rule keys now in the STDOUT of `mine`.
2. This PR should not affect the data in the paper because we have mentioned the date of the experiment.

`sonar-java-checks` is the last dependency in Sorald which bind to the internal codebase of SonarSource. Once this PR is merged, upgrading `sonar-java` should require the replacement of `sonar-java.jar` only assuming they haven't deprecated any rule for which we have a processor.

algomaster99:
@khaes-kth we can review this together for merge now.

monperrus:
excellent work!

Issue #698, opened 2022-01-26
add support for repairing Qodana warnings


monperrus:
It'd be great to add support for repairing Qodana bug warnings

Qodana: https://www.jetbrains.com/qodana/

Some processors already exist in https://github.com/MartinWitt/laughing-train by @martinwitt

monperrus:
This would come under the abstraction https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/rule/StaticAnalyzer.java

MartinWitt:
So after some thoughts I believe there are 2 ways to integrate the laughing-train code or qodana in sorald. The fast way is to integrate the current refactorings with their own detection. 
The more advanced way is using some kind of sarif parser using the qodana sarif output https://www.jetbrains.com/help/qodana/qodana-sarif-output.html. Qodana could be run with https://github.com/JetBrains/qodana-cli only requiring docker installed or requiring the sarif file provided by the user. 

In the sarif file we could match the rules by their ID and fix the bad smell with their position. This would allow sorald to fix a lot more bad smells like value always null or other very advanced IntelliJ stuff. (A sarif parser could be https://github.com/Contrast-Security-OSS/java-sarif)

The first approach yields faster results the second needs more time and should fix more bad smells(assuming there is no legal stuff hindering us).  Which approach do you prefer?

MartinWitt:
So after some more experimenting:

## How to integrate Qodana

More or less simple if we assume a docker host is running. Pull the image if it isn't available local and run it with the given parameters. After the Qodana run is finished, simply parse the SARIF file(there exist some good libs for it). Then convert the sarif rules to Sorald rules. A variant without docker does not exist to my knowledge. If creating an own container wrapping the Qodana container(docker in docker) or simple invocation  is more or less a style question. 

## The Interface Problem
The Sorald interface for an analyzer is a list of files, some rules and classpath.
`Collection<RuleViolation> findViolations(List<File> files, List<Rule> rule, List<String> classpath):`
The Qodana interface is more or less
`SarifFile analyze(String projectRoot, String pathToQodanaRulesFile, String pathToAnalyze)`
As you see, Qodana is project based, Sorald is file based. Qodana needs the whole project and not only some files.
So Qodana requires a new interface, or we simply abuse the contract and say the first file is the project root. 

## The Qodana Runtime

[...]

PR #757, opened 2022-03-23
refactor: convert constructor args to method parameter and store `SonarLintEngine` as a local variable

MartinWitt:
In #698 we had the discussion to add the `projectroot` as parameter for static analyzers because often they need it. This PR changes the interfaces and adds parameter to the interface. Furthermore, I would refactor the sonarlint-engine field to a local variable instead of a field, which is only used in a single method.

### DIFF
--- a/src/main/java/sorald/sonar/SonarStaticAnalyzer.java
+++ b/src/main/java/sorald/sonar/SonarStaticAnalyzer.java
@@ -16,15 +16,14 @@ import sorald.rule.StaticAnalyzer;
 
 public class SonarStaticAnalyzer implements StaticAnalyzer {
 
-
     @Override
-    public Collection<RuleViolation> findViolations(File projectRoot,
-            List<File> files, List<Rule> rules, List<String> classpath) {
+    public Collection<RuleViolation> findViolations(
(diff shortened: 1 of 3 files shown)


src/main/java/sorald/sonar/SonarStaticAnalyzer.java
```diff
@@ -15,22 +15,15 @@
 import sorald.rule.StaticAnalyzer;
 
 public class SonarStaticAnalyzer implements StaticAnalyzer {
-    private final File projectRoot;
-    private final SonarLintEngine sonarLint;
-
-    public SonarStaticAnalyzer(File projectRoot) {
-        this.projectRoot = projectRoot;
-        this.sonarLint = SonarLintEngine.getInstance();
```
algomaster99, line 23: I did this because I wanted to "instantiate" the instance of `SonarLintEngine` only and then I store it in the field here. This was done for faster execution of test cases. I now realise I have already done that by making `theOnlyInstance` static [here](https://github.com/SpoonLabs/sorald/blob/master/src/main/java/sorald/sonar/SonarLintEngine.java#L62). Thanks for the refactoring.

monperrus:
great intermediate step towards our grand goal! thanks @MartinWitt

PR #758, opened 2022-03-24
feat: use SPI for analyzer loading

MartinWitt:
This PR loads all classes implementing `StaticAnalyzer` and invokes them. Fixes more or less the to-do inside the code. We could change `StaticAnalyzer` from interface to abstract class to enforce a non-arg constructor or hope the programmer writes his code correct. Only `StaticAnalyzer` with a non-arg constructor are usable for SPI.

### DIFF
--- a/src/main/java/sorald/sonar/ProjectScanner.java
+++ b/src/main/java/sorald/sonar/ProjectScanner.java
@@ -5,11 +5,13 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.ServiceLoader;
 import java.util.Set;
 import sorald.Constants;
 import sorald.FileUtils;
 import sorald.rule.Rule;
 import sorald.rule.RuleViolation;
+import sorald.rule.StaticAnalyzer;
 
 /** Helper class that uses Sonar to scan projects for rule violations. */
 public class ProjectScanner {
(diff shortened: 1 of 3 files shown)


algomaster99:
@MartinWitt how about we create a branch called `spi` and we merge your PRs over there until it is entirely ready for merge into `master`?

MartinWitt:
Sounds like a reasonable idea because the PR itself is somewhat worthless without the rest. I believe you need to create a branch and I have to create a new PR for the branch.

algomaster99:
Pushed a new branch [`spi`](https://github.com/SpoonLabs/sorald/tree/spi). You can submit all your PRs here.

MartinWitt:
I changed the target branch to SPI.

src/main/java/sorald/sonar/ProjectScanner.java
```diff
@@ -61,10 +63,11 @@ private ProjectScanner() {}
...
             }
         }
-
-        // TODO generalize to not directly use the SonarStaticAnalyzer
-        var violations =
-                new SonarStaticAnalyzer().findViolations(baseDir, filesToScan, rules, classpath);
+        ServiceLoader<StaticAnalyzer> analyzers = ServiceLoader.load(StaticAnalyzer.class);
+        Set<RuleViolation> violations = new HashSet<>();
+        for (StaticAnalyzer analyzer : analyzers) {
+            violations.addAll(analyzer.findViolations(baseDir, filesToScan, rules, classpath));
```
algomaster99, line 69: Just one question. Do you plan on explicitly providing the name of the SPI (SonarStaticAnalyzer in this case) or keep this loop so that we report all violations from all static analyzers?

algomaster99 (approved):

src/main/java/sorald/sonar/ProjectScanner.java
MartinWitt: We need another method in the future `public static Set<RuleViolation> scanProject(
            File target, File baseDir, List<Rule> rules, List<String> classpath, String analyzerID)` with `analyzerID` as parameter for supporting the explicit invocation of a single analyzer.

algomaster99: That sounds great. `analyzerID` will somehow be passed through CLI and then we can parse it and pass it to service loaders.

PR #759, opened 2022-03-25
feat: abstract ruletype enum with IRuleType

MartinWitt:
Currently, the RuleType enum is used in alot of places. But there could exist more categories of bad smells than the 3 sonar ones. Using an interface in the code allows adding new categories easier, e.g., from other analyzers, without constantly changing the enum.

### DIFF
--- a/src/main/java/sorald/rule/IRuleType.java
+++ b/src/main/java/sorald/rule/IRuleType.java
@@ -1,0 +1,10 @@
+package sorald.rule;
+
+public interface IRuleType {
+
+    /**
+     * Returns the name of the rule type. A rule type is a category of rules, such as "Bug",
+     * "Vulnerability".
+     */
+    String getName();
+}
--- a/src/main/java/sorald/rule/RuleType.java
+++ b/src/main/java/sorald/rule/RuleType.java
@@ -4,8 +4,8 @@ package sorald.rule;
 public enum RuleType implements IRuleType {
     BUG("Bug"),
     VULNERABILITY("Vulnerability"),
-    CODE_SMELL("Code Smell"),
-    SECURITY_HOTSPOT("Security Hotspot");
+    CODE_SMELL("Code_Smell"),
+    SECURITY_HOTSPOT("Security_Hotspot");
 
     private final String name;
 
--- a/src/main/java/sorald/rule/RuleType.java
+++ b/src/main/java/sorald/rule/RuleType.java
@@ -1,9 +1,20 @@
 package sorald.rule;
 
 /** Enumeration of Sonar rule types */
-public enum RuleType {
-    BUG,
-    VULNERABILITY,
-    CODE_SMELL,
-    SECURITY_HOTSPOT;
+public enum RuleType implements IRuleType {
+    BUG("Bug"),
+    VULNERABILITY("Vulnerability"),
+    CODE_SMELL("Code Smell"),
+    SECURITY_HOTSPOT("Security Hotspot");
+
+    private final String name;
+
+    RuleType(String name) {
+        this.name = name;
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
 }
--- a/src/main/java/sorald/sonar/SonarRule.java
+++ b/src/main/java/sorald/sonar/SonarRule.java
@@ -2,13 +2,14 @@ package sorald.sonar;
 
 import java.util.Objects;
 import org.sonarsource.sonarlint.core.rule.extractor.SonarLintRuleDefinition;
+import sorald.rule.IRuleType;
 import sorald.rule.Rule;
 import sorald.rule.RuleType;
 
 public class SonarRule implements Rule {
     private final String key;
     private final String name;
-    private final RuleType type;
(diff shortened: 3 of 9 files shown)


src/main/java/sorald/rule/RuleType.java
```diff
@@ -1,9 +1,20 @@
 package sorald.rule;
 
 /** Enumeration of Sonar rule types */
-public enum RuleType {
-    BUG,
-    VULNERABILITY,
-    CODE_SMELL,
-    SECURITY_HOTSPOT;
+public enum RuleType implements IRuleType {
```
algomaster99, line 4:
```suggestion
public enum SonarRuleType implements IRuleType {
```
We can be more explicit here, I think.

src/main/java/sorald/sonar/SonarRule.java
```diff
@@ -2,13 +2,14 @@
...
 import org.sonarsource.sonarlint.core.rule.extractor.SonarLintRuleDefinition;
+import sorald.rule.IRuleType;
 import sorald.rule.Rule;
 import sorald.rule.RuleType;
 
 public class SonarRule implements Rule {
     private final String key;
     private final String name;
-    private final RuleType type;
+    private final IRuleType type;
```
algomaster99, line 12: Inside package `sonar`, we can safely use `SonarRuleType`.

src/main/java/sorald/rule/RuleType.java
```diff
@@ -1,9 +1,20 @@
...
+    BUG("Bug"),
+    VULNERABILITY("Vulnerability"),
+    CODE_SMELL("Code_Smell"),
+    SECURITY_HOTSPOT("Security_Hotspot");
```
algomaster99, line 8: Any specific reason to pass an argument to the constructor? We can use `name` method of `Enum` to get the string representation which will default to `BUG` and so on.

algomaster99 (changes requested):
Having enums inherit from an interface is a good idea to segregate rule types from different analyzers. We can proceed with this. Look at my comments.

src/main/java/sorald/rule/RuleType.java
algomaster99: We can also move `SonarRuleType` inside `sonar` package. Eventually, we can make `sonar` package a dependent module on the SPI we are working on.

MartinWitt: > Eventually, we can make sonar package a dependent module on the SPI we are working on.

Refactoring the sonar package to a subproject is the long-term goal.

src/main/java/sorald/sonar/SonarRule.java
MartinWitt: I would keep programming against the interface instead of the concrete type. This stops a programmer to add special behavior to `SonarRuleType` instead of the `IRuleType` interface.

algomaster99: > This stops a programmer to add special behavior to SonarRuleType instead of the IRuleType interface.

Right. We would want to have uniform behaviour across all enums and that is why we are subtyping them with an interface.

src/main/java/sorald/rule/RuleType.java
MartinWitt: The name method always returns the name in uppercase and is not really changeable. Using a String would allow us to change the name(the String) in the future. Furthermore, printing these is way easier because you don't need a lot of String manipulations. For example, you would print `Security_Hotspot` more likely than `SECURITY_HOTSPOT`. Converting `SECURITY_HOTSPOT` to a user-friendly, readable version requires unnecessary complex logic.

src/main/java/sorald/sonar/SonarRule.java
MartinWitt: There is also no real benefit to remove the abstraction and use the concrete class, or? I have more or less learned to program against interfaces, and never really thought about why we do this tbh.

algomaster99: > There is also no real benefit to remove the abstraction and use the concrete class, or?

It was readability for me. I could be sure that `type` would be one of the four enums defined in `SonarRuleType`.

src/main/java/sorald/rule/RuleType.java
algomaster99: > The name method always returns the name in uppercase and is not really changeable.

Sounds good. 

> Furthermore, printing these is way easier because you don't need a lot of String manipulations

If it is supposed to be an output for user, it is better to name it `toString` than `getName`.

src/main/java/sorald/rule/IRuleType.java
```diff
@@ -0,0 +1,10 @@
+package sorald.rule;
+
+public interface IRuleType {
+
+    /**
+     * Returns the name of the rule type. A rule type is a category of rules, such as "Bug",
+     * "Vulnerability".
+     */
```
algomaster99, line 8: This is sufficient, I believe.

algomaster99 (approved):

PR #763, opened 2022-03-29
feat: add rule repository

MartinWitt:


src/main/java/sorald/api/RuleRepository.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public interface RuleRepository {
+
+    /**
+     * Get all rules available to Sorald for analysis. Note that not all rules have a defined
+     * repair.
+     *
+     * @return a collection of all rules. Never null
+     */
+    public Collection<Rule> getAllRules();
```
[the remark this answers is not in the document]
MartinWitt: Good point, I added them.

src/main/java/sorald/annotations/ProcessorsClassGenerator.java
```diff
@@ -22,7 +22,7 @@
     public void process(ProcessorAnnotation annotation, CtClass<T> element) {
         if (cu == null) {
             cu = getFactory().createCompilationUnit();
-            processorsClass = getFactory().Type().get("sorald.Processors");
+            processorsClass = getFactory().Type().get("sorald.sonar.SonarProcessorRepository");
```
algomaster99, line 25: We should load `processorsClass` using `ServiceLoader`.

src/main/java/sorald/cli/RepairCommand.java
```diff
@@ -55,7 +56,7 @@
                 names = {Constants.ARG_RULE_KEY},
                 description =
                         "Choose one of the following rule keys:\n"
-                                + Processors.RULE_DESCRIPTIONS
+                                + SonarProcessorRepository.RULE_DESCRIPTIONS
```
algomaster99, line 59: We would also need to generalise `RepairCommand` using the service loader. However, it is not in the scope of the current PR. Putting this here to keep a mental note.

src/main/java/sorald/api/RuleRepository.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Collection;
+import javax.annotation.Nonnull;
+import sorald.rule.IRuleType;
+import sorald.rule.Rule;
+
+/**
+ * This class defines the interface for a repository of rules. It is used by Sorald to find all
+ * rules to use. All implementing classes need a no-arg constructor.
+ */
+public interface RuleRepository {
```
algomaster99, line 12: Is there some difference between `RuleRepository` and `Rules`? I think we can get rid of the latter as `RuleRepository` defined the behaviour and `SonarRules` implements it.

algomaster99 (changes requested):

src/main/java/sorald/annotations/ProcessorsClassGenerator.java
MartinWitt: Oh, you want to load each processor with SPI?

src/main/java/sorald/api/RuleRepository.java
MartinWitt: Rules is the aggregation class providing an API to query all repositories. We have multiple `RuleRepository` in the future.

algomaster99: Sounds good. Can we keep a better name so that it is clearer? I can't think of any.

src/main/java/sorald/annotations/ProcessorsClassGenerator.java
algomaster99: The code generator class would be required for all static analyzers, right? So I thought we could generalise this as well. What do you think?

MartinWitt: No, it's not needed for all analyzers. How other analyzers implement the ProcessorRepository API is an internal implementation detail. We could change the internals of our `ProcessorRepository` to use SPI instead of codegen.

src/main/java/sorald/api/RuleRepository.java
MartinWitt: `RuleProvider` would fit the naming scheme imo. Then we have multiple repository containing data and a single provider to query against.

src/main/java/sorald/annotations/ProcessorsClassGenerator.java
algomaster99: > How other analyzers implement the ProcessorRepository API is an internal implementation detail

That is a strong point. We use code generation it _implement_ SonarProcessor repository and we might need to for other static analyzers. 

Should we also put this class inside `sonar` package? Eventually, it will become a maven dependency so it will be clear which files we need to package.

MartinWitt: Yes, this class needs to be moved or replaced by SPI. It's more or less your choice.

algomaster99:
There are some conflicts. I don't know why I cannot view them here.

MartinWitt:
> There are some conflicts. I don't know why I cannot view them here.

I moved a file in branch A and changed it in Branch B. It's fixed now.

algomaster99:
I shall merge after CI is green for sanity.

MartinWitt:
Perfect, the next step is moving the API and sonar away from the core to 2 subprojects.

algomaster99:
We keep them inside Sorald as individual modules, right?

MartinWitt:
Yes

PR #783, opened 2022-04-05
feat: create sorald-api

MartinWitt:
This PR finally creates the sorald API module. I have only included the necessary classes in the API.


sorald-api/src/main/java/sorald/rule/Rule.java
```diff
@@ -13,14 +11,4 @@
...
-
-    /**
-     * Create a rule based on the key.
-     *
-     * @param key A key for which to create a rule.
-     * @return A rule based on the key.
-     */
-    static Rule of(String key) {
-        return new SonarRule(key);
-    }
```
algomaster99, line 25: It is okay we removed this from the interface. We can instead always use `new SonarRule` in `sorald-sonar`.

algomaster99 (approved):
Looks good to me! The next step is to rename `sorald/sorald` to `sorald/sorald-sonar`, I guess?

MartinWitt:
The next step is to move sorald-sonar classes to a sorald-sonar submodule, yes.

PR #786, opened 2022-04-07
chore: release sorald 0.4.0

algomaster99:
Used the `tools/sorald/release.py` to perform a release. I generated the changelog automatically using GitHub's option for now. It is not very easy to understand given that we are releasing after too long. We should probably perform releases more frequently and for that we need to document how frequently and upon what changes we release.

I went with a minor version upgrade because now we report all sonar violations instead of what was defined by `Checks`. There are many more feature introductions, of course.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
     <groupId>se.kth.castor</groupId>
     <artifactId>sorald</artifactId>
-    <version>0.4.0</version>
+    <version>0.4.1-SNAPSHOT</version>
 
     <name>Sorald</name>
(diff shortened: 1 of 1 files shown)


algomaster99:
@monperrus we can soon have a major version (`Sorald v1.0.0`) release as well. Do we have to do something before we can perform that release?

monperrus:
That's great, let's proceed?

algomaster99:
We can merge this PR. However, we must not squash them as documented [her](https://github.com/SpoonLabs/sorald/tree/master/experimentation/tools#soraldrelease).

PR #789, opened 2022-04-13
WIP: feat: add Qodana Analyzer

MartinWitt:
very early wip version.


algomaster99:
@MartinWitt Thanks for the work. I will have a look at it on Monday :)

algomaster99:
@MartinWitt How about we merge the SPI work version and then we start continuing with this integration of Qodana? Do we need to change the usage of Sorald because of SPI integration? Right now, we could just use `sorald-jar-with-dependencies` to run Sorald.

PR #794, opened 2022-04-21
fix: fix `test_prrecorder::test_make_full_record`

algomaster99:
The `test-support-scripts` workflow in our CI has been failing for a few weeks. See [this](https://github.com/SpoonLabs/sorald/runs/6104922154?check_suite_focus=true) for example. My first thought was that our implementation of Sorald has changed but seeing the logs of workflow runs did not reflect so.
Screenshot from 2022-04-21 13-24-07

The problem lied in the [difference of diff](https://github.com/SpoonLabs/sorald/runs/5975449271?check_suite_focus=true#step:8:312) which is weird because the actual diff is fetched from GitHub API. One possible cause of the failure can be that GitHub API has updated the way it wants to show the context of the diff.


khaes-kth (approved):
Seems reasonable

PR #801, opened 2022-04-26
chore: fix `test_prrecorder::test_make_full_record`

algomaster99:
Similar to #794 

I am not sure what is happening with GitHub APIs. It seems they are changing it too frequently and altering such minute details. Our `test-support-script` was failing yet again. See https://github.com/SpoonLabs/sorald/runs/6168004259?check_suite_focus=true.


algomaster99:
@khaes-kth Somehow GitHub does not automatically put the PR title as the message of the squashed commit and hence we are getting commits on `master` which do not follow our conventions so let's be careful from the next time that we update the title.

khaes-kth:
Ture, I forgot to copy the title.

Issue #774, opened 2022-04-01
Workflow is unable to update `combine-prs.yml` file


algomaster99:
Reference #755 

The `combine-prs` workflow fails [here](https://github.com/SpoonLabs/sorald/runs/5785241211?check_suite_focus=true#step:4:53) because it does not have access to updating `combine-prs.yml` which is weird because it was able to edit `tests.yml` and other workflow files.

algomaster99:
Only `tests.yml` can be edited. See https://github.com/SpoonLabs/sorald/runs/5830447298?check_suite_focus=true#step:4:42. This is so peculiar.

algomaster99:
We should use other ways of authentication than `GITHUB_TOKEN`. See https://github.community/t/inconsistent-behaviour-of-updating-workflow-files-with-github-token/243396.

algomaster99:
We have now shifted to using a [GitHub app](https://github.com/apps/combine-pr/) that has permission to update workflows. See https://github.com/SpoonLabs/sorald/pull/829.

PR #803, opened 2022-05-02
chore: create combined PR using GitHub app

algomaster99:
Reference: https://github.com/SpoonLabs/sorald/issues/774

I created a [GitHub app](https://github.com/apps/combine-pr) to deal with the shortcomings listed in the above issue.


algomaster99:
Takeaway: one should explicity pass token while using `actions/checkout` otherwise `GITHUB_TOKEN` is used by default. Reference: https://github.community/t/refusing-to-allow-a-github-app-to-create-or-update-workflow-without-workflows-permission/182573.

PR #804, opened 2022-05-02
tests: be aware of what rules Sorald can report

### LINKED ISSUE #798: Keep track of what rules are added or removed with each upgrade of sonar-java jar
A test case could be added to keep a track of what rules are there for each version of the sonar-java jar. See https://github.com/SpoonLabs/sorald/issues/331#issuecomment-1106357604.

algomaster99:
Fixes #798 

This test is written to be aware of what rules we are scanning for. Such a test was not needed before wrapping our architecture in SonarLintEngine because we used to manually pass "sonar-java checks" to our engine. It is expected to fail with each version upgrade of sonar-java. This is intended, however, since we can report the change of rules supported by Sorald in our changelogs. If this test does not fail, that means sonar-java does not have any change in the set of rules it can scan for.


khaes-kth:
> It is expected to fail with each version upgrade of sonar-java

So, does it mean that we should update `all_rules.txt` to get a green test again?

algomaster99:
@khaes-kth yes.

khaes-kth:
As I understand from you comment (https://github.com/SpoonLabs/sorald/issues/331#issuecomment-1106357604), the goal is to document what rules we mine. I think it is better to mention the sonar version that we use and refer to its documents in our own documents instead of adding this `all-rules.txt`. What do you think?

algomaster99:
The goal is to tell users the _diff_ of rules we mine. That diff should further go into the changelog.

PR #808, opened 2022-05-18
refactor: restructure project to load Sorald as a service using Java SPI

algomaster99:
@MartinWitt, thanks for your work, first of all. It is a great milestone toward integrating the Qodana static analyzer and development in general, as the code quality has definitely increased.

Let's work towards merging this now because there is feature request #807 that require a lot of code movement. I feel it is better to start working on top of the changes in this PR to avoid dealing with merge conflicts later. Some things that need to be done before we merge this PR:
- [x] Update usage in README #810 
- [x] What do we do about the [docs](https://github.com/SpoonLabs/sorald/tree/spi/docs)? Most files are related to `sorald` and not `sorald-parent`. Should we move it? I am in favour of letting them be for two reasons: #809 
  - `docs` should be accessible from the root of the repository
  - There is no immediate need to do so.
- [ ] Your final review of the changes you have made. I hope I have not messed them up during merging from `master`.

@MartinWitt , please give your opinions about subtasks 2 and 3. Once all of them are resolved, we can proceed to merge. Note that this PR will **not** be squashed as each commit is an atomic change in itself.

PR #811, opened 2022-05-18
chore: let Sonar to scan all modules

algomaster99:
Sonar is not being run in the CI for the new changes in SPI. These changes try to fix that.


.github/workflows/tests.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         shell: bash
         run: .github/mem-constrained-exec.sh
       - name: Run Sonar
-        if: ${{ matrix.os == 'ubuntu-latest' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'SpoonLabs/sorald') }}
```
slarse, line 75: That's not a good idea. You can't use secrets in pull requests from forks, so by removing this all PRs from forks will fail.

slarse: The commit before you removed this safety check did run the Sonar scan, as it matches on the head repo name.

algomaster99: I just did this for testing purposes. I am going to revert this eventually.

When I logged into the SonarCloud portal, I couldn't see a report for the scan made for this PR. That is why I tried to force it.

The issue has to do something with the refactoring. Until now, we only needed to register `SpoonLabs_sorald`, but now we have three modules - `sorald_parent`, `sorald`, and `sorald_api` - thanks to the #808 . The scans which I saw throw warnings that they are not able to find the sorald-api JAR.

algomaster99:
We are getting somewhere.  We have the [report on SonarCloud](https://sonarcloud.io/code?id=SpoonLabs_sorald&pullRequest=811) for this PR even though it does not scan the sub modules. Trying to figure out what is wrong.

algomaster99:
Not asking anyone to review this because:
1. I am a bit confident that it should work
2. There is an urgent need to merge #808 because some features rely on it. This PR is a prerequisite for it.
3. Even if the Sonar scan fails on `master`, I can fix it later because it is not a big issue.

PR #812, opened 2022-05-20
refactor: reuse APIs defined in RuleRepository in RuleProvider

algomaster99:
The APIs provided in `RuleProvider` were already implemented in `SonarRuleRepository` so I refactored the former class to use those implementations. This improves the code coverage and removed redundant code.

@MartinWitt could you please review it? I also had one question. `RuleProvider` is now a generic class - it can fetch rules for other static analyzers in future. Does this indicate that we put it inside `sorald-api`?

### DIFF
--- a/sorald/src/main/java/sorald/rule/RuleProvider.java
+++ b/sorald/src/main/java/sorald/rule/RuleProvider.java
@@ -1,12 +1,11 @@
 package sorald.rule;
 
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


algomaster99:
@MartinWitt I cannot wait for your review, actually, so I apologize for merging it quickly. I think my intent is right and that is why I am not hesistating. However, do reply to my question when you get time. :)

PR #815, opened 2022-05-23
refactor: pass CLI context rather than individual parameter

algomaster99:
These changes modify how options from CLI are passed through the program. As of now, each option, like `classpath`, was passed individually as an argument to the concerned methods. However, upon the addition of new options in the CLI, we would have to modify all those methods. This is not good coding practice. Thus, I refactored the methods to use a generic CLI option store to contain all the arguments passed to the program.

### DIFF
--- a/sorald-api/src/main/java/sorald/cli/CLIConfigForStaticAnalyzer.java
+++ b/sorald-api/src/main/java/sorald/cli/CLIConfigForStaticAnalyzer.java
@@ -1,0 +1,9 @@
+package sorald.cli;
+
+import java.util.List;
+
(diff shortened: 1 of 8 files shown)


khaes-kth:
> > Doesn't it affect any tests?
> 
> No. You can still pass the classpath in the command line. It is just passed differently throughout our program. The change just add one intermediate step where the classpath is saved in a config object and then that object is passed.
> 
> > That means we have not tested our CLI?
> 
> We have tested our CLI. There is no change in the behaviour of CLI so there is no expectation of a test failure.

You're right.

PR #837, opened 2022-06-16
chore: fix failing tests

algomaster99:
Apparently, maven launcher was failing to build the classpath due to an unresolvable dependency. For now, I ignored it from the test case's pom so that it is able to build the classpath. However, I am not sure how it works npw because that dependency, even though unresolvable, is a transitive dependency.


khaes-kth:
Thanks @algomaster99

Issue #863, opened 2022-08-17
Scheduled builds are failing on Windows


algomaster99:
The scheduled builds are failing on `windows-latest` since a few days. Example - https://github.com/SpoonLabs/sorald/actions/runs/2872246682. ` sorald.GatherStatsTest.statisticsFile_containsExpectedStats` is failing for `SEGMENT` and `MAVEN`.

PR #864, opened 2022-08-18
DONOTMERGE: chore: debug CI failure on windows-latest

algomaster99:
Debugging #863


algomaster99:
Proposed a fix in #865.

PR #865, opened 2022-08-18
chore: fix windows build failure

### LINKED ISSUE #863: Scheduled builds are failing on Windows

algomaster99:
Fixes #863 

Phew ... this took a while to debug. Playing around in #864 made me realise that the repair time for some test was being reported incorrectly. As you can see [here](https://github.com/SpoonLabs/sorald/runs/7897645976?check_suite_focus=true#step:10:228), the `repairTimeMs` for one of the repairs is `0ms`, and in the tests we want it to be [greater than `0`](https://github.com/SpoonLabs/sorald/blob/master/sorald/src/test/java/sorald/GatherStatsTest.java#L81). I draw upon the following conclusion from my analysis - the repairs were performed in less than a millisecond so the difference was not captured by `repairEnd` and `repairStart` as they both were not precise enough.

I am using `System.nanoTime` for computing time in the changes because 1) it is more precise and 2) it seems to be more reliable according to many online resources including this StackOverflow post [here](https://stackoverflow.com/a/2979239).

### DIFF
--- a/sorald/src/main/java/sorald/event/collectors/RepairStatisticsCollector.java
+++ b/sorald/src/main/java/sorald/event/collectors/RepairStatisticsCollector.java
@@ -118,8 +118,8 @@ public class RepairStatisticsCollector implements SoraldEventHandler {
     }
 
     /** @return The total amount of time spent repairing */
-    public long getRepairTimeMs() {
-        return repairTotal / (long) Math.pow(10, 6);
+    public double getRepairTimeMs() {
+        return repairTotal / Math.pow(10, 6);
     }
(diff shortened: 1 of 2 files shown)


algomaster99:
Discussed it with @khaes-kth in person. Merging.

Issue #570, opened 2021-06-16
MathOnFloatProcessor and CastArithmeticOperandProcessor can produce non-compilable output


slarse:
This is rather tricky. Sorald is technically doing exactly what Sonar says, namely casting to double. It's just that in a few cases, that's not the right thing to do (even though Sonar says so). For example:

```java
// original, sonar says cast operands to double
float c = a + b;
// repaired
float c = (double) a + (double) b;
```

The problem is of course that the right hand side is now a double, but you can't assign a double to a float due to lossy conversion. Sonar only says to cast to double, so from that perspective, Sorald is doing the right thing.

It's the same case for all compilation failures. If the variables are changed to `double` instead of `float`, the compile error disappears. My only suggestion for solving this is that, in the case that whatever reads the result of the expression expects a float, then cast the entire expression to float. E.g. like so:

```java
// original, sonar says cast operands to double
float c = a + b;
// repaired
float c = (float) ((double) a + (double) b);
```

It makes for some really ugly code, though. @fermadeiral thoughts? Any other ideas?

_Originally posted by @slarse in https://github.com/SpoonLabs/sorald/issues/569#issuecomment-861347385_

slarse:
According to discussion during meeting, we don't attempt to repair if we can tell before hand that there will be a compile error.

lyxell:
@lyxell https://github.com/SpoonLabs/sorald/pull/569#issuecomment-861347385
> Does it really fix the problem though? Couldn't it be the case that this approach actually just supresses a potentially important warning for a bug?
> 
> ```java
> public class Test {
>     public static void main(String[] args) {
>          float a = 16777216.0f;
>          float b = 1.0f;
>          float c1 = a + b; // Sonar warning, yields 1.6777216E7 not 1.6777217E7
>          System.out.println(c1 == a); // true
> 
>          float c2 = (float) ((double) a + (double) b); // No Sonar warning, also yields 1.6777216E7 not 1.6777217E7
>          System.out.println(c2 == a); // true
> 
>          double c3 = (double) a + (double) b; // Yields 1.6777217E7
>          System.out.println(c3 == a); // false
>     }
> }
> ```

@slarse https://github.com/SpoonLabs/sorald/pull/569#issuecomment-862100235
> > Does it really fix the problem though? Couldn't it be the case that this approach actually just supresses a potentially important warning for a bug?
> 
[...]

PR #866, opened 2022-08-19
fix: mark `MathOnFloatProcessor` as incomplete

algomaster99:
One of the ignored tests was bothering me for some time so I though I should fix it. It turned out that resolving the bug stated in #570 required some more dependent fixes and refactoring so I proceeded with them as well. Here is the summary of the list of changes in this PR.

1. `MathOnFloatProcessor` is marked as incomplete. It won't attempt to repair when the expected type of the `CtVariable` is `float` as it produces a non-compilable output.
2. Minimise the number of casts we perform. SonarSource suggests that we should cast _both_ of the operands to `double`, but that is not necessary as JVM implicitly does type coercion.
3. Use [type casting API](https://github.com/INRIA/spoon/blob/2fff4c50266f79bc06b961e3b01ca6c6e0fcc1ab/src/main/java/spoon/reflect/code/CtExpression.java#L38) in Spoon to add type cast.


sorald/src/main/java/sorald/processor/MathOnFloatProcessor.java
```diff
@@ -56,12 +56,8 @@ protected void repairInternal(CtBinaryOperator element) {
...
                     binaryOperator.setType(getFactory().Type().doublePrimitiveType());
                 }
-                if (!binaryOperator.getType().equals(getFactory().Type().doublePrimitiveType())) {
-                    binaryOperator
-                            .getRightHandOperand()
-                            .setTypeCasts(List.of(getFactory().Type().doublePrimitiveType()));
-                    binaryOperator.setType(getFactory().Type().doublePrimitiveType());
-                }
+                // We do not need to cast the type of the right hand operand as it is already a
+                // double
```
algomaster99, line 60: @monperrus not related to autocompletion, but when I deleted the above code, the inline comment wa suggested by co-pilot. 🤌🏼

algomaster99:
Please merge this once #868 is merged into this PR.

sorald/src/main/java/sorald/processor/MathOnFloatProcessor.java
khaes-kth: Unbelievable!

khaes-kth (approved):

khaes-kth:
That would be better if we could add a test to cover line 86 of this file: https://app.codecov.io/gh/SpoonLabs/sorald/compare/866/changes#D1L86

sorald/src/main/java/sorald/processor/MathOnFloatProcessor.java
monperrus: crazy

PR #878, opened 2022-09-09
governance: Remove slarse as developer

slarse:
I haven't been actively involved in this project for along time, so I think removing my name from the list(s) of developers. These are to my understanding for those actively maintaining the project and should not act as a historical account of contributors.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,13 +20,6 @@
     </licenses>
     <developers>
         <developer>
-            <id>slarse</id>
-            <name>Simon Larsén</name>
-            <email>slarse@kth.se</email>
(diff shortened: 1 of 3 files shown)


algomaster99:
Hi @slarse ! Sorry for looking at this so late. I lost its email notification under the pile of work I have right now.

Maven says " if the person should not be contacted about the project, they do not need to be listed here.". Reference: https://maven.apache.org/pom.html#Developers. If you wish to opt-out, we can remove you from the developer's list.

algomaster99 (approved):

monperrus:
And let me thank you again for the work you put in Sorald @slarse

PR #900, opened 2022-10-08
deps: update SonarJava dependency

### LINKED ISSUE #659: Upgrade sonar-java to 7.15.0.30507

algomaster99:
Fix #659

### DIFF
--- a/sorald/src/main/resources/config.properties
+++ b/sorald/src/main/resources/config.properties
@@ -1,1 +1,1 @@
-SONAR_JAVA_PLUGIN_URL=https://repo1.maven.org/maven2/org/sonarsource/java/sonar-java-plugin/6.12.0.24852/sonar-java-plugin-6.12.0.24852.jar
+SONAR_JAVA_PLUGIN_URL=https://repo1.maven.org/maven2/org/sonarsource/java/sonar-java-plugin/7.14.0.30229/sonar-java-plugin-7.14.0.30229.jar


algomaster99:
Yes! The comments are no more.

algomaster99:
`S2142` has some two changes:
1. Violation is still detected if catcher of `InterruptedException` is inserted after a catcher whose type is a super type of `InterruptedException`.
2. `ThreadDeath` repairs have also been modified so we need to fix that. 
Need to be discussed with @khaes-kth as this rule was a part of the paper so we need to check what all claims will change before proceeding.

khaes-kth:
I think this doesn't affect our paper, as our paper was published based on an older version. We can focus on the best solutions to adjust our tool/docs.

algomaster99:
I pushed [`75b29b0` (#900)](https://github.com/SpoonLabs/sorald/pull/900/commits/75b29b088fe09483bbc3535a3ceb743bc0d7f92a) to reach our coverage target. Otherwise, `catch(InterruptedException | Exception e) { }` results in a compilation error so it is extremely rare to find this case.

sorald/src/test/resources/processor_test_files/S2142_InterruptedException/InterruptedExceptionForTesting.java.expected
```diff
@@ -93,6 +92,26 @@ class InterruptedExceptionForTesting {
 		}
 	}
 
+	public void run4() {
```
khaes-kth, line 95: There seems to be an indentation issue

sorald/src/test/resources/processor_test_files/S2142_InterruptedException/InterruptedExceptionForTesting.java.expected
```diff
@@ -64,7 +64,6 @@ class InterruptedExceptionForTesting {
 			// clean up state...
 			throw e;
 		} catch (ThreadDeath threadDeath) { // Noncompliant {{Either re-interrupt this method or rethrow the "ThreadDeath".}}
```
khaes-kth, line 66: Remove the noncompliance comment

khaes-kth (changes requested):

PR #903, opened 2022-10-10
[DONOTMERGE]: compare sonar-java quickfix status

algomaster99:
Ran a scan on sonar-java-checks version `7.14.0.30229` to see how we fair with their quick fixes status. Use GitHub rich diff feature to view the comparison.


algomaster99:
Closing it as I prefer to keep this comparison for our research purposes only.

PR #918, opened 2022-10-19
fix(S4973): add null check to the overall expression

### LINKED ISSUE #915: Bug in S4973
Rule [S4973](https://rules.sonarsource.com/java/RSPEC-4973) repair can produce `NullPointerException`. As of now, it does the following repair:

```diff
String a = null;
String b = "xyz";

- return a == b;
+ return a.equals(b);
```

This will result in an `NPE` during runtime. A better fix would be the following and also suggested by SonarSource.
```diff
String a = null;
String b = "xyz";

- return a == b;
+ return a != null a.equals(b);
```
I personally feel that a better fix would be the usage of [`Objects.equals(Object, Object)`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-) however, this is only valid for Java version `>=7`. Anyway, we blindly follow SonarSource.

algomaster99:
Fixes #915 

SonarSource has updated their [repair for S4973](https://rules.sonarsource.com/java/RSPEC-4973) and included a `null` check before comparison. Since we cannot know which operand is `null` while static analysing, we add a `null` check to the left operand.

### DIFF
--- a/sorald/src/main/java/sorald/processor/CompareStringsBoxedTypesWithEqualsProcessor.java
+++ b/sorald/src/main/java/sorald/processor/CompareStringsBoxedTypesWithEqualsProcessor.java
@@ -17,17 +17,39 @@ public class CompareStringsBoxedTypesWithEqualsProcessor
 
     @Override
     protected void repairInternal(CtBinaryOperator<?> element) {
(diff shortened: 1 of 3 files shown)


khaes-kth (approved):
LGTM!

PR #917, opened 2022-10-19
fix(S1068): delete field writes as well

algomaster99:
Fixes #913

PR #921, opened 2022-10-21
fix(S1068): delete field writes of violated field only

algomaster99:
I forgot to add a filter to delete field writes of only those field which are unused in #917 . The changes ensure that all field writes are not deleted.


algomaster99:
I ran the local build of the plugin on `npefix` and now it compiles after the repair.

khaes-kth (approved):
LGTM!

PR #924, opened 2022-10-21
fix: return correct version of sorald

### LINKED ISSUE #895: `-v` always returns `LOCAL`
Running `java -jar sorald-jar-with-dependencies -v` returns `LOCAL`. It needs to fixed to return the correct version.

algomaster99:
Fixes #895 

The CLI always returned `LOCAL` and did not return the actual version. The changes fix that. Now, the SNAPSHOT versions return the `commit sha` and the releases return the version number.


algomaster99:
@khaes-kth have a look at this when you get time.

PR #955, opened 2022-12-20
fix: Do not throw away stacktrace when downloading sonar java fails

I-Al-Istannen:
This is causing some headaches as CI runs are failing without any clue as to *why*. The URL is reachable locally (but maybe not from the runner?) and the saving works (but maybe the runner has a permission or space issue?).

Was there a deliberate reason for swallowing the stacktrace here?


algomaster99:
> Was there a deliberate reason for swallowing the stack trace here?

Not really. I just ignored it because I thought the only error would be that this URL is not reachable. And this sometimes happens. There is some flakiness in the CI because this URL is not reachable at time.

There is no harm in not swallowing the stack trace though.

I-Al-Istannen:
> And this sometimes happens. There is some flakiness in the CI because this URL is not reachable at time.

Ah, I see. That's unfortunate but maybe not really our problem… Thanks for taking care of the PR :)

PR #960, opened 2023-01-17
style: throw exception if cache directory is not created

algomaster99:
Sorald is unable to run on [repairnator CI](https://ci.eclipse.org/repairnator/job/repairnator-ci/job/PR-1256/28/console). I will push commits here and release SNAPSHOT versions and then ask @Sofi1410 to trigger the CI until the issue is fixed.


algomaster99:
The problem seems to be resolved now. See https://ci.eclipse.org/repairnator/job/repairnator-ci/job/PR-1256/31/.

Issue #938, opened 2022-11-10
Unpin version of `sorald-parent` and `sorald-api`


algomaster99:
We pinned these modules' versions in https://github.com/SpoonLabs/sorald/pull/911/ because we needed to make the integration tests pass (somehow they were not able to resolve SNAPSHOT versions). We should revert to the old way where we have SNAPSHOT version because it makes development easier. For example, I needed to update spoon in #934, but now I have to release sorald-parent (https://github.com/SpoonLabs/sorald/pull/936) so that `sorald` can get the version of `sorald-parent` that contains the newer version of spoon. `mvn install` might work, but if it is possible to version these modules as SNAPSHOT only, I would like to do that.

PR #967, opened 2023-01-19
chore: unpin versions

algomaster99:
Fixes https://github.com/SpoonLabs/sorald/issues/938

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
     <groupId>se.kth.castor</groupId>
     <artifactId>sorald-parent</artifactId>
-    <version>0.0.3</version>
+    <version>0.0.4-SNAPSHOT</version>
     <packaging>pom</packaging>
 
     <name>Sorald-parent</name>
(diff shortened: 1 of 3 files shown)


algomaster99:
From e-mail to @khaes-kth 
> The deployment of sorald is a little messed up right now. As of now, I am separately maintaing `sorald-parent` (common dependecies), `sorald-api` (common classes that can also be used if we integrate Qodana), and `sorald` (repair tool using Sonar). Initially, sorald-parent and sorald-api was just SNAPSHOTs, and we never released them, but once sorald was converted to a maven-plugin, its test required that we pin versions. Pinning versions have brought us some disadvantages:

> - Whenever dependencies in sorald-parent and sorald-api is upgraded, I have to make sure that a new version of these modules are released, and they are also used by sorald. So most of the times, sorald just relies on older third-party dependencies.
>  - We need a CD for sorald-parent and sorald-api - more work :/

I think the changes in this PR should resolve the deployment issue.

PR #1021, opened 2023-04-17
ci: use jreleaser for performing releases

algomaster99:
After an excellent experience in using jreleaser, I am integrating this for Sorald as well. :D


.github/workflows/jreleaser.yml
```diff
@@ -0,0 +1,113 @@
...
+          JRELEASER_GPG_PASSPHRASE: ${{ secrets.SORALD_SIGNING_KEY_PASSPHRASE }}
+          JRELEASER_GPG_PUBLIC_KEY: ${{ secrets.SORALD_SIGNING_PUBLIC_KEY }}
+          JRELEASER_GPG_SECRET_KEY: ${{ secrets.SORALD_SIGNING_KEY }}
+          JRELEASER_NEXUS2_MAVEN_CENTRAL_USERNAME: ${{ secrets.OSSH_USERNAME }}
+          JRELEASER_NEXUS2_MAVEN_CENTRAL_PASSWORD: ${{ secrets.OSSH_PASSWORD }}
+
+      # Now we can update the changelog
+      - name: Prepend changelog from out/jreleaser/CHANGELOG.md to CHANGELOG.md with version as header
+        run: |
+          echo "" | cat - CHANGELOG.md > temp && mv temp CHANGELOG.md
```
aalmiray, line 85: FWIW latest JReleaser offers an option to append the generated changelog to a given file. See `append` section in https://jreleaser.org/guide/latest/reference/release/changelog.html

jreleaser.json
```diff
@@ -0,0 +1,76 @@
...
+  "distributions": {
+    "sorald": {
+      "type": "SINGLE_JAR",
+      "stereotype": "CLI",
+      "artifacts": [
+        {
+          "path": "sorald/target/{{distributionName}}-{{projectVersion}}-jar-with-dependencies.jar"
+        },
+        {
+          "path": "sorald/target/{{distributionName}}-{{projectVersion}}-jar-with-dependencies.jar.asc"
```
aalmiray, line 49: signature files should be uploaded by default. This may not be needed.

.github/workflows/jreleaser.yml
algomaster99: Where can I find the default template `src/jreleaser/templates/append-changelog.tpl`? It does not seem to be hosted on https://github.com/jreleaser/jreleaser.

PR #1024, opened 2023-04-17
ci: fix signing of artifact

algomaster99:


algomaster99:
Cannot squash this PR because it has a commit corresponding to a release. So I shall make a merge commit. Sorry git history. :anguished:

PR #1033, opened 2023-05-03
fix(S1481): add repair for enhanced for-loop

algomaster99:
Fixes https://github.com/ASSERT-KTH/sorald/issues/1020

The issue has been there for quite some time. I had prepared a fix long ago, but SonarSource does not suggest it. I repaired the enhanced for-loop like so:
```diff
-        for (String input : inputList) { // noncompliant
+        for (int inputIterator=0; inputIterator < inputList.size(); ++inputIterator) { // compliant
            doSomething();
        }
```
For `Collection`, I use `.size`, and for arrays, I use `.length`.

I think this fix should be okay even though SonarSource does not suggest it. Do you have any opinions @Zuplyx and @khaes-kth against repairing something that SonarSource does not suggest?

### DIFF
--- a/sorald/src/main/java/sorald/processor/UnusedLocalVariableProcessor.java
+++ b/sorald/src/main/java/sorald/processor/UnusedLocalVariableProcessor.java
@@ -56,7 +56,7 @@ public class UnusedLocalVariableProcessor extends SoraldAbstractProcessor<CtLoca
                     factory.createCodeSnippetExpression(
                             newSimpleName
                                     + " < "
-                                    + ((CtVariableRead) forLoopIterable)
+                                    + ((CtVariableRead<?>) forLoopIterable)
                                             .getVariable()
                                             .getSimpleName()
                                     + ".length");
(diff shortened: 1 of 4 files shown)


Zuplyx:
I've got no problems with this approach.

algomaster99:
They claim that [such code](https://github.com/SonarSource/sonar-java/blob/87bafd552ceeb736826c4002eb74be2056b760fd/java-checks-test-sources/src/main/java/checks/unused/UnusedLocalVariableCheck.java#LL41C1-L43C6) is indeed a non-compliant. However, they do not suggest a quick fix for this. Moreover, they explicitly say that they have a [partial quick fix](https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1481.json#L16) for it. I think it is okay if we repair it.

sorald/src/main/java/sorald/processor/UnusedLocalVariableProcessor.java
```diff
@@ -2,13 +2,76 @@
...
+            // inputIterator < input.size()
+            endCondition =
+                    factory.createCodeSnippetExpression(
+                            newSimpleName
+                                    + " < "
+                                    + ((CtVariableRead<?>) forLoopIterable)
+                                            .getVariable()
+                                            .getSimpleName()
+                                    + ".size()");
+        }
```
algomaster99, line 73: @MartinWitt Can you drop your reviews regarding better ways to perform transformations here?

khaes-kth:
@algomaster99 
It seems reasonable to me as we.

